PR #24, opened 2018-07-16
initial pom

### LINKED ISSUE #16: Zeroth Order BOM
Machine consumable file (likely Maven pom.xml) listing:

All immediate libraries we commit to supporting along with versions that work together. E.g. most google-cloud-java libraries plus a few others that aren't yet pulled into google-cloud-java. 

Should be in a new top-level directory.

elharo:
fix #16 @garrettjonesgoogle @netdpb 

Throwing out early for comment. This is my initial thought for what our zeroth order POM could look like.


garrettjonesgoogle:
So far this just looks like a subset of https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-bom/pom.xml , but with Guava added, right?

elharo:
so far.

The initial question is simply the format. If this looks good, then we can discuss what really needs to be here.

@netdpb suggested that what we want initially is simply the Apache Beam dependencies.

boms/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+    </license>
+  </licenses>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <google.cloud.java.ga.version>1.36.0</google.cloud.java.ga.version>
+    <google.cloud.java.ga.version>1.36.0</google.cloud.java.ga.version>
+  </properties>
+
+  <dependencies>
```
garrettjonesgoogle: This needs to be under a `<dependencyManagement>` section to be a BOM

boms/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
```
garrettjonesgoogle: Guava isn't a 0th order library

elharo: It is depended on by Beam. I'm trying to work through the Beam deps, and understand what is and isn't. New PR coming soon.

boms/pom.xml
garrettjonesgoogle: If that's the case, then why doesn't this bom import zero.xml?

boms/pom.xml
elharo: Also see https://issues.apache.org/jira/projects/BEAM/issues/BEAM-812?filter=allopenissues

boms/zero.xml
```diff
@@ -0,0 +1,225 @@
...
+      </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-databind</artifactId>
+        <version>2.9.5</version>
+      </dependency>
+      <dependency>
+        <groupId>io.grpc</groupId>
+        <artifactId>grpc-core</artifactId>
+        <version>1.2.0</version>
```
garrettjonesgoogle: The versions of many things here have been bumped up in the master branch of beam: https://github.com/apache/beam/blob/master/buildSrc/src/main/groovy/org/apache/beam/gradle/BeamModulePlugin.groovy#L298

I am hesitant about using the really old versions here.

boms/zero.xml
```diff
@@ -0,0 +1,225 @@
...
+        <version>0.7.1</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.auth</groupId>
+        <artifactId>google-auth-library-oauth2-http</artifactId>
+        <version>0.7.1</version>
+      </dependency>
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>slf4j-api</artifactId>
```
garrettjonesgoogle: This isn't a Google artifact (and there are others too in this list)

boms/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+      <version>20.0</version>
+    </dependency>
+
+    <!-- Google Cloud Java APIs -->
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-bigquery</artifactId>
+      <version>${google.cloud.java.ga.version}</version>
```
garrettjonesgoogle: This really should be using the google-cloud-java bom I think...

boms/zero.xml
elharo: done; may have missed some

boms/zero.xml
elharo: done

boms/pom.xml
elharo: done

netdpb:
Let's agree that the contents of this file can change after we merge this PR, because we still haven't decided exactly the scope. That's fine; we can start here and iterate.

boms/pom.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+                             http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.google.cloud</groupId>
+  <artifactId>zero-pom-test</artifactId>
```
netdpb: What is the purpose of this artifact?

boms/zero.xml
```diff
@@ -0,0 +1,205 @@
...
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <google.cloud.java.ga.version>1.38.0</google.cloud.java.ga.version>
+  </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.apache.beam</groupId>
```
netdpb: Should this be here? It is not a Google artifact.

Same with the other non-Google artifacts below.

boms/zero.xml
```diff
@@ -0,0 +1,205 @@
...
+        <artifactId>jackson-databind</artifactId>
+        <version>2.9.5</version>
+      </dependency>
+      <dependency>
+        <groupId>io.grpc</groupId>
+        <artifactId>grpc-core</artifactId>
+        <version>1.2.0</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.apis</groupId>
```
netdpb: Can we at least sort these by group ID and then artifact ID? I should hope that the "clusters" won't go across group IDs.

boms/zero.xml
```diff
@@ -0,0 +1,205 @@
...
+        <artifactId>google-api-services-bigquery</artifactId>
+        <version>v2-rev374-1.23.0</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.api</groupId>
```
netdpb: WHY OH WHY is there both `com.google.api` and `com.google.apis`?

(Not your problem, obvs.)

boms/zero.xml
```diff
@@ -0,0 +1,205 @@
...
+      </dependency>
+      <dependency>
+        <groupId>io.grpc</groupId>
+        <artifactId>grpc-all</artifactId>
+        <version>1.2.0</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+</project>
```
netdpb: Should there always be an ending newline?

boms/zero.xml
```diff
@@ -0,0 +1,190 @@
...
+    <dependencies>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>20.0</version>
+      </dependency>
+      <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-databind</artifactId>
+        <version>2.9.5</version>
```
saturnism: should remove.

boms/zero.xml
```diff
@@ -0,0 +1,190 @@
...
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <google.cloud.java.ga.version>1.40.0</google.cloud.java.ga.version>
+  </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>20.0</version>
```
saturnism: this is a problematic one, would prefer not to include.

boms/zero.xml
saturnism: for gRPC, we should include all grpc modules here, and can be replaced by gRPC BOM if when it becomes available.

boms/zero.xml
saturnism: hahaha.. i know :(

Although, I don't feel we should include this since it's managed through google-cloud-java already.

boms/zero.xml
```diff
@@ -0,0 +1,190 @@
...
+      </dependency>
+      <dependency>
+        <groupId>io.grpc</groupId>
+        <artifactId>grpc-netty</artifactId>
+        <version>1.2.0</version>
+      </dependency>
+      <dependency>
+        <groupId>io.netty</groupId>
+        <artifactId>netty-handler</artifactId>
+        <version>4.1.8.Final</version>
```
saturnism: for netty, i'd recommend importing netty BOM.

boms/zero.xml
```diff
@@ -0,0 +1,190 @@
...
+      </dependency>
+      <dependency>
+        <groupId>joda-time</groupId>
+        <artifactId>joda-time</artifactId>
+        <version>2.4</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-core</artifactId>
+        <version>1.36.0</version>
```
saturnism: wouldn't it make sense to just include google-cloud-java's BOM here?

boms/zero.xml
```diff
@@ -0,0 +1,190 @@
...
+        <version>0.10.0</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.auth</groupId>
+        <artifactId>google-auth-library-oauth2-http</artifactId>
+        <version>0.10.0</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.protobuf</groupId>
+        <artifactId>protobuf-java</artifactId>
```
saturnism: for protobuf, include all protobuf submodules if any.

boms/zero.xml
elharo: moot in latest iteration.

boms/zero.xml
elharo: done

garrettjonesgoogle:
As has been stated somewhere else, the more frequent releases are and the more up-to-date you try to keep, the less likely convergence is even possible. I'd like to not spit into the wind here - i.e. I'd like to avoid attempting to achieve convergence.

cpovirk:
RE: convergence: How do you feel about [`requireUpperBoundDeps`](https://maven.apache.org/enforcer/enforcer-rules/requireUpperBoundDeps.html)? ejona86 speaks highly of it, and it seems like a desirable property.

elharo:
Let me experiment with requireUpperBoundDeps. If it makes the tests pass, that might be more practical. I still think we should also know where dependencies are lagging so we can fix them.

boms/upperBoundsCheck/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+        <groupId>com.google.cloud</groupId>
+        <artifactId>zero-bom</artifactId>
+        <version>0.0.1-SNAPSHOT</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>    
+  </dependencyManagement>
+
+  <!-- TODO is there a way to run this test directly on the BOM
```
cpovirk: Internally, I just wrote some details about [what I think we'd want](https://docs.google.com/document/d/1U2Ru2koluWfOXYskqW1NPHBp7uDcugpy23v1nR9M7DU/edit?ts=5b8493cd#heading=h.xhofic44mqoo). It's possible that we could do it in the BOM itself, but we'd have to write our own mojo, and even that might be tricky (though I don't really have any idea). Externally, I put together a hacky shell-based [proof of concept](https://docs.google.com/document/d/1HV-rfos5hMMzqix4orXiObEMTfBQefidN12boQw5U08/edit).

boms/zero-bom/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <google.cloud.java.version>0.59.0-alpha</google.cloud.java.version>
+    <io.grpc.version>1.14.0</io.grpc.version>
+    <protobuf.version>3.6.1</protobuf.version>
+  </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <!-- GRPC; specifically https://github.com/grpc/grpc-java -->
```
saturnism: we should include _all_ grpc sub-modules, e.g. grpc-context, grpc-all, etc.

boms/convergence-check/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+        <configuration>
+          <rules>
+            <dependencyConvergence/>
+          </rules>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+</project>
```
netdpb: Please add newlines.

boms/pom.xml
```diff
@@ -1,73 +1,34 @@
...
-          <rules>
-            <dependencyConvergence/>
-          </rules>
-        </configuration>
-      </plugin>
-    </plugins>
-  </build>
+  <modules>
+    <module>zero-bom</module>
+    <module>upperBoundsCheck</module>
```
netdpb: Should these use a consistent naming scheme? lower-hyphenated vs lowerCamelCase

Specifically, should the directory names match the artifact names?

boms/pom.xml
```diff
@@ -1,73 +1,34 @@
...
-    A compatible set of Google Cloud open source libraries.
-  </description>
-
+  <version>1.0-SNAPSHOT</version>
+  <name>BOM Parent</name>
+  <url>https://github.com/GoogleCloudPlatform/cloud-opensource-java/boms</url>
   <organization>
-    <name>Google LLC</name>
-    <url>https://cloud.google.com</url>
+    <name>Google LLC.</name>
```
netdpb: Sometimes there's a period at the end and sometimes not. I don't know which is correct, but can we be consistent?

Also, are we really an LLC?

boms/pom.xml
```diff
@@ -1,73 +1,34 @@
...
+    <url>http://www.google.com</url>
```
netdpb: You've made this change here, but not in the rest of the POMs. Should we be consistent?

boms/upperBoundsCheck/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+       instead of on a pom.xml that imports the BOM? -->
+  <dependencies>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-core</artifactId>
```
netdpb: Should we depend on *all* of the `com.google.cloud` artifacts here?

boms/pom.xml
elharo: I don't know if directory names should match artifact names, but they certainly shouldn't be the same. Fixed. 

Otherwise, I'm not aware of any conventions on artifact IDs.

boms/pom.xml
elharo: Yes, Google is really an LLC. Alphabet is an Inc. Ping me internally if you want to know about the period.

boms/pom.xml
elharo: I suppose. Fixed.

boms/zero-bom/pom.xml
```diff
@@ -0,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+                             http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.google.cloud</groupId>
+  <artifactId>zero-bom</artifactId>
```
garrettjonesgoogle: I think "zero-bom" doesn't really describe what we're doing here any more.

elharo: Alternatives? core-bom?

boms/zero-bom/pom.xml
garrettjonesgoogle: These are the three that grpc recommends for using grpc in a conventional way. In contrast, we are creating a BOM here. So, anyone who *does* declare a dependency on modules outside of these three needs them to stay in sync with these three, otherwise everything breaks. Thus, all grpc modules should be in the BOM.

boms/zero-bom/pom.xml
garrettjonesgoogle: How about `cloud-oss-bom`?

boms/cloud-oss-bom/pom.xml
```diff
@@ -0,0 +1,144 @@
...
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <google.cloud.java.version>0.59.0-alpha</google.cloud.java.version>
+    <io.grpc.version>1.14.0</io.grpc.version>
+    <protobuf.version>3.6.1</protobuf.version>
+  </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <!-- GRPC; specifically https://github.com/grpc/grpc-java -->
```
garrettjonesgoogle: Add a note that we should import the officially-provided grpc BOM once it is created: https://github.com/grpc/grpc-java/issues/3954

boms/cloud-oss-bom/pom.xml
```diff
@@ -0,0 +1,144 @@
...
+        <artifactId>grpc-protobuf-nano</artifactId>
+        <version>${io.grpc.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>io.grpc</groupId>
+        <artifactId>grpc-testing</artifactId>
+        <version>${io.grpc.version}</version>
+      </dependency>
+
+      <!-- protobufs from https://github.com/protocolbuffers/protobuf/tree/master/java -->
```
garrettjonesgoogle: Add a note that we should import the officially-provided protobuf BOM once it is created - https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/64

boms/cloud-oss-bom/pom.xml
elharo: done

PR #27, opened 2018-07-23
List Dependencies with Maven resolver

elharo:
@garrettjonesgoogle This is my initial work on retrieving the immediate dependencies of an artifact using Apache Maven Resolver (formerly Eclipse Aether.)

The next step will be recurse through the complete transitive tree. However I need to think about the format we want that tree in first.


garrettjonesgoogle:
What is the output when you run it on a package in google-cloud-java? (e.g. google-cloud-pubsub)

elharo:
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
com.google.cloud:google-cloud-core:jar:1.37.1 (compile)
com.google.cloud:google-cloud-core-grpc:jar:1.37.1 (compile)
com.google.api.grpc:proto-google-cloud-pubsub-v1:jar:1.19.1 (compile)
com.google.api.grpc:grpc-google-cloud-pubsub-v1:jar:1.19.1 (compile)
io.grpc:grpc-netty-shaded:jar:1.13.1 (compile)
io.grpc:grpc-stub:jar:1.13.1 (compile)
io.grpc:grpc-auth:jar:1.13.1 (compile)
com.google.auto.value:auto-value:jar:1.4 (compile)
```

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Dependencies.java
```diff
@@ -0,0 +1,134 @@
...
+import org.eclipse.aether.connector.basic.BasicRepositoryConnectorFactory;
+import org.eclipse.aether.spi.connector.RepositoryConnectorFactory;
+import org.eclipse.aether.spi.connector.transport.TransporterFactory;
+import org.eclipse.aether.transport.file.FileTransporterFactory;
+import org.eclipse.aether.transport.http.HttpTransporterFactory;
+
+/**
+ * Based on Apache Maven Artifact Resolver; formerly Eclipse Aether.
+ */
+public class Dependencies {
```
garrettjonesgoogle: Maybe we could call this something more active like `DependencyPrinter`?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Dependencies.java
```diff
@@ -0,0 +1,134 @@
...
+
+import org.apache.maven.repository.internal.MavenRepositorySystemUtils;
+import org.eclipse.aether.connector.basic.BasicRepositoryConnectorFactory;
+import org.eclipse.aether.spi.connector.RepositoryConnectorFactory;
+import org.eclipse.aether.spi.connector.transport.TransporterFactory;
+import org.eclipse.aether.transport.file.FileTransporterFactory;
+import org.eclipse.aether.transport.http.HttpTransporterFactory;
+
+/**
+ * Based on Apache Maven Artifact Resolver; formerly Eclipse Aether.
```
garrettjonesgoogle: I recommend making this parenthetical instead of using the semicolon - the current grammar makes it look like you're saying that the present class (`Dependencies`) was formerly Eclipse Aether.

dependencies/pom.xml
```diff
@@ -0,0 +1,80 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.google.cloud.tools.opensource</groupId>
+  <artifactId>dependencies</artifactId>
```
garrettjonesgoogle: maybe call this `dependency-printer`?

elharo: Printing and the main method are mostly for experimentation. I'm not sure they'll survive in the final product. But the name could certainly be improved. Any other suggestions?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Dependencies.java
elharo: done

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Dependencies.java
elharo: Ditto. Name needs to be better, but DependencyPrinter doesn't feel quite right. DependencyLister? DependencyCollector? DependencyInfo? DependencyUtility?

garrettjonesgoogle: Hmm well what do you think will be the scope of what it does in the end?

elharo: Query Maven Central (or perhaps an arbitrary repo/repos) and return the dependency tree or the immediate dependencies.

garrettjonesgoogle: I think I like `DependencyLister` in that case.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
```diff
@@ -0,0 +1,135 @@
...
+import org.eclipse.aether.repository.RemoteRepository;
+import org.eclipse.aether.resolution.DependencyRequest;
+import org.eclipse.aether.resolution.DependencyResolutionException;
+
+import com.google.common.base.Preconditions;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.maven.repository.internal.MavenRepositorySystemUtils;
```
briandealwis: These imports are oddly sorted.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
```diff
@@ -0,0 +1,135 @@
...
+  private static final RepositorySystem SYSTEM = newRepositorySystem();
+  private static final RemoteRepository CENTRAL =
+      new RemoteRepository.Builder("central", "default", "http://repo1.maven.org/maven2/").build();
+
+  public static void main(String[] args)
+      throws DependencyCollectionException, DependencyResolutionException {
+    
+    if (args.length != 1 || !args[0].contains(":")) {
+      System.err.println("Usage: java "
+          + "com.google.cloud.tools.opensource.dependencies.Dependencies "
```
briandealwis: This is now out-of-date.  Maybe just use `DependencyLister.class.getName()`?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
```diff
@@ -0,0 +1,135 @@
...
+          + "groupdId:artifactId:version");
+      return;
+    }
+    
+    String[] coordinates = args[0].split(":");
```
briandealwis: Worth a `Verify.verify(coordinates.length == 3, "invalid maven coordinate");`?  You'll get an AIOOBE below, and you're also not catching a 4- or 5-item coordinate (e.g., packaging and classifiers).

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
```diff
@@ -0,0 +1,135 @@
...
+    dependencyRequest.setRoot(node);
+
+    SYSTEM.resolveDependencies(session, dependencyRequest);
+    return node;
+  }
+
+  private static RepositorySystemSession newSession() {
+    DefaultRepositorySystemSession session = MavenRepositorySystemUtils.newSession();
+
+    LocalRepository localRepo = new LocalRepository("target/local-repo");
```
briandealwis: This seems an odd choice?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
```diff
@@ -0,0 +1,135 @@
...
+    String version = coordinates[2];
+    
+    DependencyNode node = resolveNode(groupId, artifactId, version);
+    
+    for (DependencyNode dependency : node.getChildren()) {
+      System.out.println(dependency.toString());
+    }
+  }
+
+  private static DependencyNode resolveNode(String groupId, String artifactId, String version)
```
briandealwis: I think it's worth mentioning that this returns the dependencies with `compile` scope.  It's not obvious from the name.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
```diff
@@ -0,0 +1,135 @@
...
+  public static List<Artifact> getImmediateDependencies(String groupId, String artifactId, String version) 
+      throws DependencyCollectionException, DependencyResolutionException {
+    
+    Preconditions.checkNotNull(groupId, "Group ID cannot be null");
+    Preconditions.checkNotNull(artifactId, "Artifact ID cannot be null");
+    Preconditions.checkNotNull(version, "Version cannot be null");
+    
+    List<Artifact> result = new ArrayList<>();
+    
+    DependencyNode node = resolveNode(groupId, artifactId, version);
```
briandealwis: It wasn't clear that this returns only the dependencies with scope `compile`.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
```diff
@@ -0,0 +1,135 @@
...
+    locator.addService(TransporterFactory.class, FileTransporterFactory.class);
+    locator.addService(TransporterFactory.class, HttpTransporterFactory.class);
+
+    return locator.getService(RepositorySystem.class);
+  }
+
+  /**
+   * Returns the direct compile time dependencies of an artifact.
+   */
+  public static List<Artifact> getImmediateDependencies(String groupId, String artifactId, String version)
```
briandealwis: Would it make sense to relocate this to be near `resolveNode()`?

dependencies/pom.xml
garrettjonesgoogle: Remember to update this too

briandealwis: Forager, Navigator, Forester

garrettjonesgoogle: - Forager: this seems a little opaque...
- Navigator: this implies a UI, which doesn't seem to match
- Forester: this seems even more opaque...

Glad to see ideas though!

briandealwis: A forester managers lots of (dependency) trees :-)

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
briandealwis: Seems odd to have "api" methods at the end.

elharo: ack

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
```diff
@@ -0,0 +1,139 @@
...
+
+    CollectRequest collectRequest = new CollectRequest();
+    collectRequest.setRoot(dependency);
+    collectRequest.addRepository(CENTRAL);
+    DependencyNode node = SYSTEM.collectDependencies(session, collectRequest).getRoot();
+
+    DependencyRequest dependencyRequest = new DependencyRequest();
+    dependencyRequest.setRoot(node);
+
+    SYSTEM.resolveDependencies(session, dependencyRequest);
```
garrettjonesgoogle: It feels wrong to me for an identifier with all-caps to be mutated - this doesn't really feel like a constant... (at least, that's how I read this line) - I'd recommend `system` instead for the name.

elharo: done, though loooking at it now I'm not sure it is mutated, and I'm not sure it isn't. The dependencyRequest argument is mutated for sure.

Issue #30, opened 2018-07-24
Tool for generating list of needed dependency upgrades


elharo:
E.g. start with beam; find the full transitive dependency graph; identify all dependencies with multiple versions; and identify the projects and artifacts that need to be updated to later versions.

Maybe even autogenerate PRs fr github.

PR #32, opened 2018-07-24
Methods to get complete dependency graph

### LINKED ISSUE #35: Rename DependencyLister DependencyGraphBuilder
and make DependencyGraph constructor package private.

elharo:
@garrettjonesgoogle @netdpb fix #35 among other work on #30


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Artifacts.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Common utilities for operating on {@code org.eclipse.aether.artifact.Artifact} objects.
+ */
+public class Artifacts {
+
+  /**
+   * Returns the artifact's Maven coordinates in the form groupId:artifactId:version.
+   * Repo and packaging are not included.
+   */
+  public static String toCoordinates(Artifact artifact) {
```
garrettjonesgoogle: Maybe `toCoordinate` singular? If we use the plural for this, it gets awkward when you try to name a list of them.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
```diff
@@ -0,0 +1,80 @@
...
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.SetMultimap;
+import com.google.common.collect.TreeMultimap;
+
+/**
+ * A representation of the complete non-cyclic transitive dependency tree of 
+ * a Maven artifact.
+ * 
+ * The representation is unusual. Imagine we perform a breadth first search
```
garrettjonesgoogle: I'm unclear on why the representation is unusual.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
```diff
@@ -135,5 +157,75 @@ private static RepositorySystem newRepositorySystem() {
     }
     return result;
   }
+  
+  /**
+   * Finds the full compile time, transitive dependency graph including duplicates
+   *  and conflicting versions. This method makes a lot of network connections
```
garrettjonesgoogle: Remove the extraneous space

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
```diff
@@ -135,5 +157,75 @@ private static RepositorySystem newRepositorySystem() {
...
+    DependencyNode node = resolveCompileTimeDependencies(groupId, artifactId, version);  
+    DependencyGraph graph = new DependencyGraph();
+    preorder(new Stack<DependencyNode>(), node, graph);    
+    
+    return graph;
+  }
+  
+  // this finds the actual graph that Maven sees with no duplicates and at most one version per
+  // library.
+  // todo next two methods are duplicate code with only one line difference
```
garrettjonesgoogle: Make `todo` all caps

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -0,0 +1,87 @@
...
+    path.add(artifact);
+  }
+  
+  @Override
+  public String toString() {
+    return Joiner.on(" / ").join(path);
+  }
+  
+  @Override
+  public boolean equals(Object o) {
```
garrettjonesgoogle: Could we use equals/hashCode helper classes to reduce the boilerplate?

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+import com.google.common.truth.Truth;
+
+public class DependencyGraphTest {
+
+  private DependencyGraph graph = new DependencyGraph();
+  private Artifact foo = new DefaultArtifact("com.google:foo:1");
+  private Artifact bar = new DefaultArtifact("com.google:bar:1");
+  private Artifact baz1 = new DefaultArtifact("com.google:baz:1");
+  private Artifact baz2= new DefaultArtifact("com.google:baz:2");
```
garrettjonesgoogle: add a space before the `=`

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Artifacts.java
elharo: Unfortunately "coordinates" (plural) is the common usage for this string in Maven and third party docs. I suppose a singular coordinate is just one of group ID, artifact ID, etc.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
elharo: Perhaps it's not. This represents a tree as a list of every path from the root to each node. Is that a common thing to do?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
elharo: done

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
elharo: done

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
elharo: I tried Objects.hash, but it didn't quite fit because I was hashing values from inside each object in a list and Artifact itself doesn't declare a hashCode method. I could build a list of hashable objects from the artifacts, and then hash that; but it ends up being about as complex as this. 

Objects.equals runs into a similar issue that Artifact does not have a reliable equals method that matches these semantics so we have to drop down further.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphTest.java
elharo: done

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -0,0 +1,93 @@
...
+  public String toString() {
+    
+    StringBuilder builder = new StringBuilder();
+    for (Artifact artifact : path) {
+      builder.append(Artifacts.toCoordinates(artifact));
+      builder.append(" / ");
+    }
+    String s = builder.toString();
+    // chop end
+    return s.substring(0, s.length() - 3);
```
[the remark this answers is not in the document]
elharo: to get rid of the last " / " at the end. The alternative is converting to a list of strings and using a Joiner, or I could use an old style indexed for loop that skips the last element.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
elharo: done

patflynn:
Doesn't Maven have plugins that do this already?

elharo:
The Maven plugins don't usually pull the complete dependency graph including conflicting versions. Furthermore they're not designed for programmatic access. If all we need is a flat list of conflicting versions, Maven enforcer can do that. However for a generating a prioritized list of action items we need something more configurable that can produce different formats and filter the conflicts to identify what needs to be fixed.

patflynn:
@saturnism any thoughts on this? 

I assumed [maven-dependency-plugin](https://maven.apache.org/plugins/maven-dependency-plugin/analyze-mojo.html) et al, would satisfy our usecase.

@elharo what kind of prioritized lists do you have in mind?  

Overall I was expecting us to hook into the maven build/test/reports workflow as a more conventional way to perform and report on our dependency analysis.  I'm still wondering what the benefits are of building our own service w/ UI and custom analysis tools.

elharo:
The  maven-dependency-plugin doesn't consider versions, which is our main concern.
The [maven-enforcer-plugin with the dependency convergence rule](https://maven.apache.org/enforcer/enforcer-rules/dependencyConvergence.html) gets closer but at the end of the day you end up with a lot of redundant data that still needs to be manually curated. @netdpb 's idea was to generate a todo list like this for any given project:

https://docs.google.com/spreadsheets/d/1zGkc8JG8PX7X8UwC5Q-wIccc8PZr5AingK_fnEucz7M/edit#gid=0

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphTest.java
```diff
@@ -0,0 +1,137 @@
...
+  public void testFindConflicts_cloudLanguage() throws DependencyCollectionException, DependencyResolutionException {
+    DependencyGraph graph = DependencyGraphBuilder.getCompleteDependencies("com.google.cloud",
+        "google-cloud-language", "1.37.1");
+    List<DependencyPath> conflicts = graph.findConflicts();
+    List<String> leaves = new ArrayList<>();
+    for (DependencyPath path : conflicts) {
+      leaves.add(Artifacts.toCoordinates(path.getLeaf()));
+    }
+    
+    Truth.assertThat(leaves).contains("com.google.api.grpc:proto-google-common-protos:1.0.0");
```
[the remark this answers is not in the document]
netdpb: `containsAllIn` expects an `Iterable`. `containsAllOf` takes varargs.

dependencies/pom.xml
```diff
@@ -58,6 +58,12 @@
       <version>0.42</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava-testlib</artifactId>
+      <version>25.1-jre</version>
```
garrettjonesgoogle: Use a variable for guava versions (since you use multiple modules from it)

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
```diff
@@ -0,0 +1,137 @@
...
+ * <p>Artifacts are considered to be the same if they have the same group ID, artifact ID, and version.
+ */
+public class DependencyGraph {
+
+  // DependencyGraphBuilder builds this in breadth first order.
+  // That is, this list contains the paths to each node in breadth first order 
+  private final List<DependencyPath> graph = new ArrayList<>();
+  
+  // map of groupId:artifactId to versions
+  // todo if versions' values were the whole coordinate string
```
garrettjonesgoogle: Use caps for "todo" (this applies to other places too)

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
```diff
@@ -0,0 +1,137 @@
...
+  // todo if versions' values were the whole coordinate string 
+  // (or even the Artifact itself), would this be simpler?
+  private final TreeMultimap<String, String> versions =
+      TreeMultimap.create(Comparator.naturalOrder(), new VersionComparator());
+  
+  // map of groupId:artifactId:version to paths
+  private SetMultimap<String, DependencyPath> paths = HashMultimap.create();
+  
+  // hide constructor
+  DependencyGraph() {
```
garrettjonesgoogle: Did you intend to leave accessible to other code in the same package?

elharo: yes. Here I wanted to force use of the builder.

elharo:
@netdpb Any final comments?

PR #57, opened 2018-08-24
Adding JLBP-1 ("Minimize dependencies")

garrettjonesgoogle:


library-best-practices/JLBP-1.md
```diff
@@ -0,0 +1,23 @@
...
+  - All libraries used only for testing should have test scope
+    (e.g. junit, mockito, and truth).
+- Scrutinize all dependency additions. Check the result of
+  `mvn dependency:tree` (after running `mvn install -DskipTests`
+  to build the library) to see which transitive dependencies are
+  added by just adding a single dependency to your own library,
+  and if you require all of the transitive dependencies.
+- Avoid cyclic dependencies
+  - E.g., gRPC depends on OpenCensus which depends on gRPC.
+  - Maven can detect this.
```
elharo: It's a little stronger: Maven does not allow this. Build fails outright if there's a cyclic dependency.

library-best-practices/JLBP-1.md
```diff
@@ -0,0 +1,23 @@
...
+  since it doesn't need to appear on the classpath of consumers.
+  - All libraries used only for testing should have test scope
+    (e.g. junit, mockito, and truth).
+- Scrutinize all dependency additions. Check the result of
+  `mvn dependency:tree` (after running `mvn install -DskipTests`
+  to build the library) to see which transitive dependencies are
+  added by just adding a single dependency to your own library,
+  and if you require all of the transitive dependencies.
+- Avoid cyclic dependencies
+  - E.g., gRPC depends on OpenCensus which depends on gRPC.
```
elharo: Google style eschews Latin. That is, spell out "For example" instead of "E.g."

library-best-practices/JLBP-1.md
garrettjonesgoogle: @saturnism Looking at grpc's dependencies, there is no cycle technically. It goes `grpc-core` -> `opencensus-api` -> `grpc-context`. Maybe you're talking about an inter-project cycle, which is actually not detectable by maven?

garrettjonesgoogle: Thinking about this more, I think it should be a separate item - it's more subtle than just minimizing dependencies. I have pulled it offline (see internal doc).

library-best-practices/JLBP-1.md
garrettjonesgoogle: Done.

garrettjonesgoogle:
Addressed feedback, PTAL

garrettjonesgoogle:
cc @cpovirk

library-best-practices/JLBP-1.md
```diff
@@ -0,0 +1,20 @@
...
+- Use the smallest scope possible. For example, auto-value doesn't
+  need to use compile scope, and can instead use compile-only,
+  since it doesn't need to appear on the classpath of consumers.
+  - All libraries used only for testing should have test scope
+    (for example junit, mockito, and truth).
+- Scrutinize all dependency additions. Check the result of
+  `mvn dependency:tree` (after running `mvn install -DskipTests`
+  to build the library) to see which transitive dependencies are
+  added by just adding a single dependency to your own library,
+  and if you require all of the transitive dependencies.
```
loosebazooka: Is it worth expanding on "if you require all of the transitive dependencies"? Even if you exclude them in your project, any project that depends on yours will not see the "excludes" directives on transitive dependencies.

garrettjonesgoogle: I want to make sure I understand what you're saying.

Let's use the following example: say your library (C) depends on B which transitively depends on A, and in C you exclude A. It seems clear to me that if a user of C (say D) depends on C and also adds their own dependency on B, then they would pull in A (since they didn't add the exclusion of A). However, would D pull in A in the case where D *only* lists C as their dependency?

loosebazooka: Right. I believe that is true, because you don't want someone else to diasable a dependency that might be needed somewhere else. `<exclude>` is not persisted in a published artifacts pom.xml. But now I feel like I need to double check. Let me try it out.

loosebazooka: Sorry, I'm super wrong. I don't know why I thought this.

garrettjonesgoogle: Alrighty. I'm going to go ahead and merge as-is.

library-best-practices/JLBP-1.md
```diff
@@ -0,0 +1,20 @@
...
+Use the minimum number of dependencies that is reasonable.
+Adding a dependency for a large amount of functionality may be ok,
+but avoid pulling in dependencies just to save a few lines of code,
+because every dependency is a liability.
+
+Some specific notes about minimizing dependencies:
+
+- Use the smallest scope possible. For example, auto-value doesn't
+  need to use compile scope, and can instead use compile-only,
+  since it doesn't need to appear on the classpath of consumers.
```
cpovirk: I wouldn't recommend `compile-only` for `auto-value` and other annotations artifacts. It's caused problems for Guava users: See https://github.com/google/guava/issues/2721 and various other bugs files against us.

For AutoValue specifically, I would recommend [the pattern in its user guide](https://github.com/google/auto/blob/master/value/userguide/index.md#in-pomxml), in which `auto-value-annotations` is a `compile` dependency and `auto-value` itself is an annotation processor (which apparently you can accomplish by making it `provided`, but that seems like not the most precise way).

I can send a PR if you'd like. Maybe you _want_ it to include an example of when `compile-only` is the right thing? If so, maybe that's the servlet API case?

elharo: I'm not familiar with auto-value details, but servlets are definitely not compile-only.

cpovirk: Ah, I am going off Ray's comment in an internal thread. It sounds like I can expect to hear a response from you there later :)

garrettjonesgoogle: @cpovirk a PR would be much appreciated! If we don't truly have any examples where we should use compile-only, then we should probably remove the mention of it.

PR #59, opened 2018-08-28
Adding JLBP-3 (Use semantic versioning)

garrettjonesgoogle:
cc @cpovirk

### DIFF
--- a/library-best-practices/JLBP-3.md
+++ b/library-best-practices/JLBP-3.md
@@ -13,8 +13,8 @@ Definition of semver: https://semver.org
     - Google-cloud-java uses `@BetaApi`, `@InternalApi`, and
       `@InternalExtensionOnly`
   - Library documentation should point users to a tool (or tools) that can
-    help them automatically detect when they are inadvertently using features
-    marked with these annotations.
+    help them detect when they are using features marked with these
+    annotations.
   - Even though semver rules don't apply to unstable features, it is
     recommended to bump a library's minor version if unstable features have
     surface breakages.
@@ -24,13 +24,13 @@ Definition of semver: https://semver.org
   - http://www.mojohaus.org/clirr-maven-plugin/
   - Example compatibility report for grpc-core:
     https://abi-laboratory.pro/index.php?view=timeline&lang=java&l=grpc-core 
-- Examples of breaking changes to a stable Public API that require a new
-  major version:
+- Examples of breaking changes to a Public API that require a new major
+  version:
   - Upgrading to a non-compatible dependency that is exposed through a
     library's Public API (for dependencies that follow semver, this happens
     when a dependency is bumped to a higher major version)
   - Changing a method signature
-  - Removing a deprecated method
+  - Removing a method (deprecated or not)
 - Examples of additions that require a new minor version:
   - Upgrading a dependency to a new minor version (compatible, but new
     features) that is exposed through a library's Public API
--- a/library-best-practices/JLBP-3.md
+++ b/library-best-practices/JLBP-3.md
@@ -1,0 +1,45 @@
+[JLBP-3] Use Semantic versioning
+--------------------------------
+
+Definition of semver: https://semver.org
+
+- For the purpose of compliance with semver, the stable surface of a library
+  serves as its "Public API." Annotations can be used to mark features
+  (classes, methods, etc) as unstable, and thus not part of the Public API,
+  such that semver rules don't have to apply. 
+  - Examples of annotations:
+    - Guava uses `@Beta`
+    - grpc-java uses `@ExperimentalApi`
+    - Google-cloud-java uses `@BetaApi`, `@InternalApi`, and
+      `@InternalExtensionOnly`
+  - Library documentation should let users know how to detect usage of
+    features that use these annotations
+  - Even though semver rules don't apply to unstable features, it is
+    recommended to bump a library's minor version if unstable features have
+    surface breakages.
+- Tools are available to help identify accidental incompatibilities within a
+  major version. Examples:
+  - https://lvc.github.io/japi-compliance-checker/ 
+  - http://www.mojohaus.org/clirr-maven-plugin/
+  - Example compatibility report for grpc-core:
+    https://abi-laboratory.pro/index.php?view=timeline&lang=java&l=grpc-core 
+- Examples of breaking changes that require a new major version:
+  - Upgrading to a non-compatible dependency that is exposed through a
+    library's Public API (for dependencies that follow semver, this happens
+    when a dependency is bumped to a higher major version)
+  - Changing a method signature
+  - Removing a deprecated method
+- Examples of additions that require a new minor version:
+  - Upgrading a dependency to a new minor version (compatible, but new
+    features) that is exposed through a library's Public API
+  - Adding a new class
+  - Adding a new method
+- Special case: Bumping up the minimum required Java version doesn't always
+  require a major version bump. Requirements for skipping the major version
+  bump:
+  - Requirement 1: Only a pure change in the minimum required Java version
+    is allowed. Any surface change that breaks users still requires a major
+    version bump.
+  - Requirement 2: The major version can be kept only when the obsoleted
+    version has minimal usage. If an obsoleted version has medium or high
+    usage, a major version bump may be warranted.
--- a/library-best-practices/JLBP-3.md
+++ b/library-best-practices/JLBP-3.md
@@ -12,8 +12,9 @@ Definition of semver: https://semver.org
     - grpc-java uses `@ExperimentalApi`
     - Google-cloud-java uses `@BetaApi`, `@InternalApi`, and
       `@InternalExtensionOnly`
-  - Library documentation should let users know how to detect usage of
-    features that use these annotations
+  - Library documentation should point users to a tool (or tools) that can
+    help them automatically detect when they are inadvertently using features
+    marked with these annotations.
   - Even though semver rules don't apply to unstable features, it is
     recommended to bump a library's minor version if unstable features have
     surface breakages.
@@ -23,7 +24,8 @@ Definition of semver: https://semver.org
   - http://www.mojohaus.org/clirr-maven-plugin/
   - Example compatibility report for grpc-core:
     https://abi-laboratory.pro/index.php?view=timeline&lang=java&l=grpc-core 
-- Examples of breaking changes that require a new major version:
+- Examples of breaking changes to a stable Public API that require a new
+  major version:
   - Upgrading to a non-compatible dependency that is exposed through a
     library's Public API (for dependencies that follow semver, this happens
     when a dependency is bumped to a higher major version)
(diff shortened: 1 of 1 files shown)


library-best-practices/JLBP-3.md
```diff
@@ -0,0 +1,45 @@
...
+--------------------------------
+
+Definition of semver: https://semver.org
+
+- For the purpose of compliance with semver, the stable surface of a library
+  serves as its "Public API." Annotations can be used to mark features
+  (classes, methods, etc) as unstable, and thus not part of the Public API,
+  such that semver rules don't have to apply. 
+  - Examples of annotations:
+    - Guava uses `@Beta`
```
elharo: Could we standardize/recommend only one of these? We shouldn't have three different ways of marking something beta.

library-best-practices/JLBP-3.md
```diff
@@ -0,0 +1,45 @@
...
+    - grpc-java uses `@ExperimentalApi`
+    - Google-cloud-java uses `@BetaApi`, `@InternalApi`, and
+      `@InternalExtensionOnly`
+  - Library documentation should let users know how to detect usage of
```
elharo: "let users know how to detect usage of" --> clearly mark and warn users of

library-best-practices/JLBP-3.md
garrettjonesgoogle: That seems like a good idea, but not within the scope of this best practice. If we do it, we can come back and update this recommendation.

library-best-practices/JLBP-3.md
```diff
@@ -0,0 +1,45 @@
...
+    when a dependency is bumped to a higher major version)
+  - Changing a method signature
+  - Removing a deprecated method
+- Examples of additions that require a new minor version:
+  - Upgrading a dependency to a new minor version (compatible, but new
+    features) that is exposed through a library's Public API
+  - Adding a new class
+  - Adding a new method
+- Special case: Bumping up the minimum required Java version doesn't always
+  require a major version bump. Requirements for skipping the major version
```
elharo: Paragraph break before "Requirements for skipping" since this is not part of the special case.

library-best-practices/JLBP-3.md
```diff
@@ -0,0 +1,45 @@
...
+  - Even though semver rules don't apply to unstable features, it is
+    recommended to bump a library's minor version if unstable features have
+    surface breakages.
+- Tools are available to help identify accidental incompatibilities within a
+  major version. Examples:
+  - https://lvc.github.io/japi-compliance-checker/ 
+  - http://www.mojohaus.org/clirr-maven-plugin/
+  - Example compatibility report for grpc-core:
+    https://abi-laboratory.pro/index.php?view=timeline&lang=java&l=grpc-core 
+- Examples of breaking changes that require a new major version:
```
elharo: We need to call out 0.x versions here to indicate that this only applies when the version is >= 1.0. Per spec "Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable."

This goes hand in hand with a requirement to use and ship 1.0+ versions.

library-best-practices/JLBP-3.md
garrettjonesgoogle: But it is part of the special case... maybe it needs to be reworded to make it clearer? Basically, I'm listing the requirements for the special case of not bumping the major version when it would normally be required (since it's a breaking change for the Java version being dropped).

library-best-practices/JLBP-3.md
garrettjonesgoogle: Why do we need to call it out when it's already in the semver spec? Do you think people will be confused as this is now?

library-best-practices/JLBP-3.md
garrettjonesgoogle: The "detect" part of this sentence references usage of tools like https://github.com/google/guava-beta-checker - does that change your recommendation on how to word this?

elharo: I would expect library documentation to specify the incompatible changes, not suggest tools for users to find out what the developer already knows.

library-best-practices/JLBP-3.md
elharo: Yes, as written I'd read this that 0.1 needs to go 1.0 when a breaking change is pushed.

library-best-practices/JLBP-3.md
garrettjonesgoogle: But... the idea is to do both (documentation + tooling), they are not mutually exclusive. We certainly do need the documentation - that is not in question. The tool part is the same idea as `@Deprecated`, where users can choose to have warnings or build failures if they use deprecated features; we also want to enable users to get warnings or build failures if they use unstable features (it's their choice). It's easy for developers to accidentally use things because of IDE autocomplete, and not realize what they have done. Also, it's something that junior developers on a team can easily do without realizing the consequences, and it's much easier on the senior developers to enforce project rules using tools instead of having to very carefully inspect all changes being made.

library-best-practices/JLBP-3.md
garrettjonesgoogle: Ok, I will try to make that clearer.

library-best-practices/JLBP-3.md
elharo: As written this seems to mix up the recommendation to library vendors and library users. Or perhaps it's a meta-recommendation: we're recommending that library vendors should recommend these tools to their users.

library-best-practices/JLBP-3.md
```diff
@@ -0,0 +1,45 @@
...
+  - Adding a new method
+- Special case: Bumping up the minimum required Java version doesn't always
+  require a major version bump. Requirements for skipping the major version
+  bump:
+  - Requirement 1: Only a pure change in the minimum required Java version
+    is allowed. Any surface change that breaks users still requires a major
+    version bump.
+  - Requirement 2: The major version can be kept only when the obsoleted
+    version has minimal usage. If an obsoleted version has medium or high
+    usage, a major version bump may be warranted.
```
cpovirk: The rules around a new Java version feel odd to me.

- As I just noted in an internal doc, a Java major-version upgrade is, in most ways that are relevant here, more like a semver library's minor-version bump. As such, it doesn't seem like it should _ever_ trigger a major-version bump in its users. What problem does a major-version bump solve? How would it tie into the forthcoming "\[JLBP-5] Avoid creating conflicts between major versions" and its recommendation to change packages for a major-version upgrade? Would using new Java features ever be worth the trouble caused by a new major version of the library? Maybe the only advice we need to give here is to not upgrade the version of Java until you know that most users can upgrade? Maybe we don't need to say anything about that here yet?

- Requirement 1 is the normal requirement for minor-version upgrades, so it feels strange to call it out again but only here.

library-best-practices/JLBP-3.md
garrettjonesgoogle: Yes, it's a meta-recommendation, that's a good way to describe it. Recommendations for library users will be documented separately from the JLBP series. I will try to update the text to make this clear.

library-best-practices/JLBP-3.md
garrettjonesgoogle: * If some day there was a version of Java that broke a bunch of stuff, maybe supporting that version might merit a major version bump. So, maybe we should call out the motivation for this rule.
* Yeah, you're right about requirement 1. I think I'll remove it.

garrettjonesgoogle:
Addressed feedback, PTAL @elharo and @cpovirk

library-best-practices/JLBP-3.md
```diff
@@ -0,0 +1,47 @@
...
+- For the purpose of compliance with semver, the stable surface of a library
+  serves as its "Public API." Annotations can be used to mark features
+  (classes, methods, etc) as unstable, and thus not part of the Public API,
+  such that semver rules don't have to apply. 
+  - Examples of annotations:
+    - Guava uses `@Beta`
+    - grpc-java uses `@ExperimentalApi`
+    - Google-cloud-java uses `@BetaApi`, `@InternalApi`, and
+      `@InternalExtensionOnly`
+  - Library documentation should point users to a tool (or tools) that can
```
elharo: delete "automatically"
perhaps delete "inadvertently" unless a tool can distinguish inadvertent and intentional uses.

library-best-practices/JLBP-3.md
```diff
@@ -0,0 +1,47 @@
...
+    recommended to bump a library's minor version if unstable features have
+    surface breakages.
+- Tools are available to help identify accidental incompatibilities within a
+  major version. Examples:
+  - https://lvc.github.io/japi-compliance-checker/ 
+  - http://www.mojohaus.org/clirr-maven-plugin/
+  - Example compatibility report for grpc-core:
+    https://abi-laboratory.pro/index.php?view=timeline&lang=java&l=grpc-core 
+- Examples of breaking changes to a stable Public API that require a new
+  major version:
```
elharo: Public API. (For dependencies that follow semver, this happens when a dependency is bumped to a higher major version.)

library-best-practices/JLBP-3.md
```diff
@@ -0,0 +1,47 @@
...
+  - Upgrading to a non-compatible dependency that is exposed through a
+    library's Public API (for dependencies that follow semver, this happens
+    when a dependency is bumped to a higher major version)
+  - Changing a method signature
+  - Removing a deprecated method
```
elharo: Removing a method (that is, deprecated or not)

garrettjonesgoogle: Done

library-best-practices/JLBP-3.md
garrettjonesgoogle: Done

library-best-practices/JLBP-3.md
garrettjonesgoogle: Done

library-best-practices/JLBP-3.md
cpovirk: The new text LGTM, thanks.

More thoughts, which don't need to affect anything you have here: (Sorry that I'm just working my way through this now....)

If someone asked about supporting Java versions, I would probably say:

- In the 99% case, keep supporting the old one until usage is minimal.
- If your library is used from Android, distinguish between the language-feature level that you're targeting and the library level that you're targeting (since tools can make new language features work with old versions of Android but mostly can't help with libraries). (Also, consider [Android market share](https://developer.android.com/about/dashboards/).)
- If you do want to add features for (say) Java 12, then try consider putting them in a separate artifact and package that contains only those features.
- You could also consider doing something like Guava has, with parallel versions. But this has its own downsides.
- Otherwise, you might actually be best served by doing only _minor_ version bumps for the immediate future. This ensures that your consumers, like other libraries, can stick with the current version of your library and still work compatibly with consumers who have moved on to the newer versions that require Java 12.
- Our advice on specific Java versions may change as we see how the new Java release schedule plays out.

elharo: LGTM on the text of the PR. On @cpovirk's points, I would be very cautious about separate artifacts for Java 12, e.g. o parallel versions. It's hard enough to maintain a single artifact, and split packages are becoming even more troublesome than they are now.

If necessary, one could maintain a single artifact in multiple branches; e.g. JUnit 3 and JUnit 4, or servlet 2.5 and servlet 3. Move the minor versions of each as needed for maintenance but keep the packages and groupId:artifactId the same. Thus we force developers to pick one explicitly and avoid getting both into the classpath. I've seen the opposite and it always seems to be a ^*&&*%^#$# of pain and confusion.

garrettjonesgoogle: Sounds like we should consider a JLBP to cover this issue - I'll add it in the candidate doc.

PR #60, opened 2018-08-28
Adding JLBP-4 (Avoid dependencies known to cause runtime conflicts)

garrettjonesgoogle:
cc @cpovirk

### DIFF
--- a/library-best-practices/JLBP-4.md
+++ b/library-best-practices/JLBP-4.md
@@ -3,12 +3,12 @@
 
 - Unstable libraries are defined as libraries that allow for breaking changes to
   their Public API within the same major version. For libraries following
-  semver, this means libraries with a 0.x.y version. (See [JLBP-3.md] for more
-  details on the recommendations for following semver.)
+  semver, this means libraries with a 0.x.y version. (See [JLBP-3](JLBP-3.md)
+  for more details on the recommendations for following semver.)
 - Unstable features are defined as features that are not part of the stable
   Public API of a stable library, often marked with annotations like `@Beta`.
-  (See [JLBP-3.md] for more details on the recommendations for annotating
-  unstable features.)
+  (See [JLBP-3](JLBP-3.md) for more details on the recommendations for
+  annotating unstable features.)
 - If your library depends on an unstable library or feature, and that feature
   experiences a breaking change between versions, your library will be locked to
   a specific version of your dependency.
--- a/library-best-practices/JLBP-4.md
+++ b/library-best-practices/JLBP-4.md
@@ -27,8 +27,8 @@
     your dependency, and it will be unsafe for your users to upgrade your
     library on its own, creating opportunities for hard-to-diagnose runtime
     conflicts for users.
-- Given the consequences of depending unstable features in dependencies, avoid
-  doing so.
+- Given the consequences of depending on unstable features in dependencies,
+  avoid doing so.
   - Depending on unstable features between submodules of a single library is
     acceptable, provided that users can easily force their build system to use
     compatible versions of the submodules. Some strategies for library owners
--- a/library-best-practices/JLBP-4.md
+++ b/library-best-practices/JLBP-4.md
@@ -1,0 +1,33 @@
+[JLBP-4] Avoid dependencies known to cause runtime conflicts
+------------------------------
+
+- Avoid using dependencies that are not stable.
+  - Example violation: The App Engine SDK introduced a non-final version of the
+    JCache API into its classpath, which conflicts with the final version of
+    JCache.
+  - The current exception is JSR 305 (`@NonNull`). It is widely referenced in
+    existing libraries and there is no compelling alternative, so it is allowed
+    as a dependency. However, it should also be noted that jsr305 has classes in
+    the javax.annotation package, which means that under the Java module system
+    introduced in Java 9, it conflicts with other artifacts that use that
+    package. The Error Prone team and others are in conversation with Jetbrains
+    and others to produce a good replacement.
+- Avoid using dependencies that duplicate classes with other dependencies.
+  - Most common case: a library that has published the same classes under
+    multiple artifact names. Always use the primary variant.
+  - Example: Guava's main artifact is `guava`, but from versions 13.0 to 17.0,
+    another artifact `guava-jdk5` was also published with classes that overlap
+    with `guava`. Maven doesn't dedupe `guava-jdk5` with `guava` because the
+    artifact names are different, which means that the same class names are
+    provided by multiple jars, and runtime errors result from classes and
+    methods not being found. The only known Google library propagating this
+    dependency is google-api-client <= 1.23.0, and it has removed the
+    `guava-jdk5` dependency (and switched to the `guava` artifact) since 1.24.1.
+- Avoid depending on `servlet-api` if not necessary.
+  - The only type of library that should depend on `servlet-api` is one
+    functioning as a utility library for use in web apps. Even in this case, the
+    `servlet-api` dependency should use `provided` scope. Other libraries should
+    not use `servlet-api` as a dependency at all, for example, to get http
+    status code definitions. Usage of this library needs to be restricted
+    because there are multiple published versions that conflict with each other.
+- New problematic dependencies may be added here later.
--- a/library-best-practices/JLBP-4.md
+++ b/library-best-practices/JLBP-4.md
@@ -1,16 +1,10 @@
 [JLBP-4] Avoid dependencies on unstable libraries and features
 --------------------------------------------------------------
 
-- At a high level, don't depend on things that can have frequent breaking
-  changes (where frequent means more than once every couple years).
-- The primary type of unstable library is one that allows for breaking changes
-  to their Public API within the same major version. For libraries following
-  semver, this means libraries with a 0.x.y version. (See [JLBP-3](JLBP-3.md)
-  for more details on the recommendations for following semver.)
-- Another type of unstable library is one that provides stable versions
-  according to semver, but which bumps major versions frequently.
-  - The primary example of this is Guava before version 22.0, which published a
-    new major version once or more per year.
+- Unstable libraries are libraries that allow for breaking changes to their
+  Public API within the same major version. For libraries following semver, this
+  means libraries with a 0.x.y version. (See [JLBP-3](JLBP-3.md) for more
+  details on the recommendations for following semver.)
 - Unstable features are features that are not part of the stable Public API of a
   stable library, often marked with annotations like `@Beta`.  (See
   [JLBP-3](JLBP-3.md) for more details on the recommendations for annotating
@@ -35,23 +29,3 @@
     include:
     - Using the same version number for all submodules in a library
     - Providing a [BOM](https://www.baeldung.com/spring-maven-bom)
-- Additionally, avoid depending on libraries that implement non-finalized JSRs,
-  even if the library otherwise meets the criteria for stability (for example,
-  using a stable version like 1.0.0).
-  - A non-finalized JSR can change before reaching the Final status, meaning
-    that an implementation of the JSR would need to make a breaking change.
-  - Example: App Engine depended on JSR-107 (JCache) before it was finalized,
-    and now the version that App Engine requires conflicts with the final
-    version.  This shouldn't be read as a warning to avoid App Engine, but
-    instead an example to warn other library implementers when making new
-    dependency choices.
-  - The current exception is JSR 305 (`@NonNull`), whose JSR status is currently
-    "Dormant." This JSR is specifically implemented in
-    `com.google.code.findbugs:jsr305`. This library is widely referenced in
-    existing libraries, it is not experiencing breaking changes, and there is no
-    compelling alternative, so it is allowed as a dependency. However, it should
-    also be noted that `jsr305` has classes in the `javax.annotation` package,
-    which means that under the Java module system introduced in Java 9, it
-    conflicts with other artifacts that use that package. The Error Prone team
-    and others are in conversation with Jetbrains and others to produce a good
-    replacement.
--- a/library-best-practices/JLBP-4.md
+++ b/library-best-practices/JLBP-4.md
@@ -1,33 +1,44 @@
-[JLBP-4] Avoid dependencies known to cause runtime conflicts
-------------------------------
+[JLBP-4] Avoid dependencies on unstable libraries and features
+--------------------------------------------------------------
 
-- Avoid using dependencies that are not stable.
-  - Example violation: The App Engine SDK introduced a non-final version of the
-    JCache API into its classpath, which conflicts with the final version of
-    JCache.
-  - The current exception is JSR 305 (`@NonNull`). It is widely referenced in
+- Unstable libraries are defined as libraries that allow for breaking changes to
+  their Public API within the same major version. For libraries following
+  semver, this means libraries with a 0.x.y version. (See [JLBP-3.md] for more
+  details on the recommendations for following semver.)
+- Unstable features are defined as features that are not part of the stable
+  Public API of a stable library, often marked with annotations like `@Beta`.
+  (See [JLBP-3.md] for more details on the recommendations for annotating
+  unstable features.)
+- If your library depends on an unstable library or feature, and that feature
+  experiences a breaking change between versions, your library will be locked to
+  a specific version of your dependency.
+  - If you expose the unstable feature on your library's surface, then your
+    library's current major version will be permanently locked to the version
+    you initially exposed, and you won't be able to upgrade without making a
+    breaking change to your users.
+  - If you only use the unstable feature in your implementation, then each minor
+    or patch version of your library will require a very specific version of
+    your dependency, and it will be unsafe for your users to upgrade your
+    library on its own, creating opportunities for hard-to-diagnose runtime
+    conflicts for users.
+- Given the consequences of depending unstable features in dependencies, avoid
+  doing so.
+  - Depending on unstable features between submodules of a single library is
+    considered acceptable, given there is a way for users to ensure they are
+    using compatible versions of the submodules together easily, e.g. by having
+    the library use the same version for all submodules, or by providing a BOM.
+- Additionally, avoid depending on libraries that implement non-finalized JSRs,
+  even if the library otherwise meets the criteria for stability.
+  - Example: App Engine depended on JSR-107 (JCache) before it was finalized,
+    and now the version that App Engine requires conflicts with the final
+    version.  This shouldn't be read as a warning to avoid App Engine, but
+    instead an example to warn other library implementers when making new
+    dependency choices.
+  - The current exception is JSR 305 (`@NonNull`), specifically as implemented
+    in `com.google.code.findbugs:jsr305`. This library is widely referenced in
     existing libraries and there is no compelling alternative, so it is allowed
-    as a dependency. However, it should also be noted that jsr305 has classes in
-    the javax.annotation package, which means that under the Java module system
-    introduced in Java 9, it conflicts with other artifacts that use that
+    as a dependency. However, it should also be noted that `jsr305` has classes
+    in the `javax.annotation` package, which means that under the Java module
+    system introduced in Java 9, it conflicts with other artifacts that use that
     package. The Error Prone team and others are in conversation with Jetbrains
     and others to produce a good replacement.
-- Avoid using dependencies that duplicate classes with other dependencies.
-  - Most common case: a library that has published the same classes under
-    multiple artifact names. Always use the primary variant.
-  - Example: Guava's main artifact is `guava`, but from versions 13.0 to 17.0,
-    another artifact `guava-jdk5` was also published with classes that overlap
-    with `guava`. Maven doesn't dedupe `guava-jdk5` with `guava` because the
-    artifact names are different, which means that the same class names are
-    provided by multiple jars, and runtime errors result from classes and
-    methods not being found. The only known Google library propagating this
-    dependency is google-api-client <= 1.23.0, and it has removed the
-    `guava-jdk5` dependency (and switched to the `guava` artifact) since 1.24.1.
-- Avoid depending on `servlet-api` if not necessary.
-  - The only type of library that should depend on `servlet-api` is one
-    functioning as a utility library for use in web apps. Even in this case, the
-    `servlet-api` dependency should use `provided` scope. Other libraries should
-    not use `servlet-api` as a dependency at all, for example, to get http
-    status code definitions. Usage of this library needs to be restricted
-    because there are multiple published versions that conflict with each other.
-- New problematic dependencies may be added here later.
--- a/library-best-practices/JLBP-4.md
+++ b/library-best-practices/JLBP-4.md
@@ -1,14 +1,20 @@
 [JLBP-4] Avoid dependencies on unstable libraries and features
 --------------------------------------------------------------
 
-- Unstable libraries are defined as libraries that allow for breaking changes to
-  their Public API within the same major version. For libraries following
+- At a high level, don't depend on things that can have frequent breaking
+  changes (where frequent means more than once every couple years).
+- The primary type of unstable library is one that allows for breaking changes
+  to their Public API within the same major version. For libraries following
   semver, this means libraries with a 0.x.y version. (See [JLBP-3](JLBP-3.md)
   for more details on the recommendations for following semver.)
-- Unstable features are defined as features that are not part of the stable
-  Public API of a stable library, often marked with annotations like `@Beta`.
-  (See [JLBP-3](JLBP-3.md) for more details on the recommendations for
-  annotating unstable features.)
+- Another type of unstable library is one that provides stable versions
+  according to semver, but which bumps major versions frequently.
+  - The primary example of this is Guava before version 22.0, which published a
+    new major version once or more per year.
+- Unstable features are features that are not part of the stable Public API of a
+  stable library, often marked with annotations like `@Beta`.  (See
+  [JLBP-3](JLBP-3.md) for more details on the recommendations for annotating
+  unstable features.)
 - If your library depends on an unstable library or feature, and that feature
   experiences a breaking change between versions, your library will be locked to
   a specific version of your dependency.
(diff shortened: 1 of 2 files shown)


library-best-practices/JLBP-4.md
```diff
@@ -0,0 +1,33 @@
...
+    existing libraries and there is no compelling alternative, so it is allowed
+    as a dependency. However, it should also be noted that jsr305 has classes in
+    the javax.annotation package, which means that under the Java module system
+    introduced in Java 9, it conflicts with other artifacts that use that
+    package. The Error Prone team and others are in conversation with Jetbrains
+    and others to produce a good replacement.
+- Avoid using dependencies that duplicate classes with other dependencies.
+  - Most common case: a library that has published the same classes under
+    multiple artifact names. Always use the primary variant.
+  - Example: Guava's main artifact is `guava`, but from versions 13.0 to 17.0,
```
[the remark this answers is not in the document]
garrettjonesgoogle: I can't tell which of those ones conflict. Which should be the primary one to use?

library-best-practices/JLBP-4.md
```diff
@@ -0,0 +1,33 @@
+[JLBP-4] Avoid dependencies known to cause runtime conflicts
+------------------------------
+
+- Avoid using dependencies that are not stable.
```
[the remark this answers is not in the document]
garrettjonesgoogle: I think JLBP-11 & JLBP-12 in the candidate best practices address these points

library-best-practices/JLBP-4.md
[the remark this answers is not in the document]
elharo: It depends on whether you want servlet 2.5, 3.0. 3.1, or 4.0. :-( 

javax.servlet:javax.servlet-api:3.1.0 for Java 8 on App Engine Standard
javax.servlet:servlet-api:2.5 for the Java & runtime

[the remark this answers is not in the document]
garrettjonesgoogle: This doesn't seem to be a concern for libraries though, right? (since our recommendation is to use the scope `provided`). Maybe the best place for this information could be the section on why `servlet-api` should be `provided`?

library-best-practices/JLBP-4.md
cpovirk: "Stable" feels like not quite the right word for these 2 cases. Presumably App Engine isn't going to change its version of JCache, nor is JSR 305 going to remove classes.

Maybe these are just 2 additional top-level bullets. Actually, the App Engine one sounds like another example of the existing "Avoid using dependencies that duplicate classes with other dependencies" bullet (if I'm understanding right). So that would just leave a bullet about dependencies that share packages with other dependencies. (That's almost a generalization of "duplicate classes with other dependencies" to cover packages, too.)

garrettjonesgoogle: The mention of JCache is more about App Engine's mistake, it isn't about App Engine stability. Because of this mistake, and App Engine's promise not to break surface, it is stuck on the unstable JCache version. I'm guessing that both App Engine's version of JCache and the final version use the same artifact name; if so, this is not an example of multiple dependencies providing duplicate classes, it would be a single artifact that Maven can choose one of (but which results in conflicts for someone). Should more details be given on this point?

As for JSR 305 - you're probably right that jsr 305 probably won't be changed to break people, although the proposal is technically dormant and not final. But then again, the findbugs implementation is a stable version (3.x), so it might not make sense to worry about this aspect of it.

The other point about JSR 305 is the issue with the module system. @elharo or @saturnism , will this result in runtime conflicts?

cpovirk: Oh, sorry, I'm misreading the text: "Non-final" doesn't mean "They made their own copy that removed the `final` keyword"; it just means "They added a dependency on an unstable version of JCache." Right?

This seems obvious in hindsight, but I think I was thrown by the difference in phrasing: The top-level bullet talks about "unstable" and "dependencies"; the sub-bullet talks about "non-final" and "classpath."

And as you note, I think I was also confused because I thought this was a recommendation to avoid App Engine, not a recommendation to avoid unstable JCache. (I mean, I guess it's sort of an indirect recommendation to avoid App Engine, to avoid these problems, but that's not the point :))

This also helps me understand why you refer to JSR305 here: As a non-accepted JSR, it could in theory change like the JCache API did.

I will think about this some more and try to figure out if there's something that could make it clearer or if that's not necessary and I'm just biased to read "non-final" as "not having the Java `final` keywords" because of my particular background :)

cpovirk: Another possible part of the confusion: The early JCache API probably wouldn't have been "known to cause classpath conflicts" (per the title of this page) when App Engine added a dependency on it. The problem was likely that they didn't put enough weight on its _potential_, as an unstable API, to cause classpath conflicts later. Can we change the title to something that emphasizes that we should be anticipating particular problems (as the doc in fact says, contrary to my initial reading of this part :)), not just reacting after the fact?

cpovirk: (It probably wasn't also clear to me that JCache was a JSR.)

cpovirk: Anyway:

Possible page title: "Avoid dependencies that may cause runtime conflicts" (or, if you prefer, "Avoid dependencies that may conflict with other dependencies")

Possible first bullet and sub-bullets:

- Avoid using dependencies that are not stable.
  - For semantically versioned dependencies, this means avoiding both 0.x libraries and libraries that release many major versions. \[maybe a note about Guava's version numbering?\]
  - Avoid using unstable APIs \[link to JLBP-3?\] (except from other modules in your own library).
  - For JSRs, avoid using them until they are final.
    - The current exception is JSR 305 ...
    - But dependencies on other JSRs have caused real problems. For example, App Engine began to depend on JSR-107 (JCache) before it was finalized, and now the version it requires conflicts with the final version.

library-best-practices/JLBP-4.md
```diff
@@ -0,0 +1,33 @@
...
+    multiple artifact names. Always use the primary variant.
+  - Example: Guava's main artifact is `guava`, but from versions 13.0 to 17.0,
+    another artifact `guava-jdk5` was also published with classes that overlap
+    with `guava`. Maven doesn't dedupe `guava-jdk5` with `guava` because the
+    artifact names are different, which means that the same class names are
+    provided by multiple jars, and runtime errors result from classes and
+    methods not being found. The only known Google library propagating this
+    dependency is google-api-client <= 1.23.0, and it has removed the
+    `guava-jdk5` dependency (and switched to the `guava` artifact) since 1.24.1.
+- Avoid depending on `servlet-api` if not necessary.
```
cpovirk: This seems like a sub-bullet of "dependencies that duplicate classes with other dependencies." OK, it's maybe slightly different because the conflict is often with a "dependency" provided by your runtime. But I think it would be nice to have 2 bullets here to clearly call out the 2 cases: unstable dependencies and dependencies that overlap with others.

library-best-practices/JLBP-4.md
cpovirk: (We might also just omit the App Engine example. I'd like to think that, of all the things we're saying here, our target audience understands why depending on unstable libraries is bad :) Or if they don't, I'm not sure that a one-sentence example is going to change that.)

garrettjonesgoogle:
I decided to split this into two separate best practices so that they are more concrete and focused, and are easier to point to. The content is mostly rewritten. PTAL @elharo and @cpovirk

library-best-practices/JLBP-4.md
```diff
@@ -0,0 +1,44 @@
...
+    breaking change to your users.
+  - If you only use the unstable feature in your implementation, then each minor
+    or patch version of your library will require a very specific version of
+    your dependency, and it will be unsafe for your users to upgrade your
+    library on its own, creating opportunities for hard-to-diagnose runtime
+    conflicts for users.
+- Given the consequences of depending unstable features in dependencies, avoid
+  doing so.
+  - Depending on unstable features between submodules of a single library is
+    considered acceptable, given there is a way for users to ensure they are
```
cpovirk: This might be nitpicking of phrasing, but: Using the same version number for all modules (or, likewise, providing a BOM) doesn't "ensure" much. It just means that, once users recognize that there's a problem, they have a relatively easy way to fix it.

I do think we'll want to do some work to help users detect these problems in the first place -- possibly by following [gRPC's lead](https://repo1.maven.org/maven2/io/grpc/grpc-auth/1.14.0/grpc-auth-1.14.0.pom) in using single-element version ranges like `[1.14.0]`. (We've at least casually referred to this option internally, but I have yet to flesh it out. Hopefully later today...)

library-best-practices/JLBP-4.md
```diff
@@ -0,0 +1,44 @@
...
+    using compatible versions of the submodules together easily, e.g. by having
+    the library use the same version for all submodules, or by providing a BOM.
+- Additionally, avoid depending on libraries that implement non-finalized JSRs,
+  even if the library otherwise meets the criteria for stability.
```
cpovirk: I'm not sure what you mean that a JSR "otherwise meets the criteria for stability." It seems like we haven't given any criteria for stability of a JSR, since it doesn't follow semantic versioning. Hmm, or maybe some JSRs do nowadays? But my impression is that JCache wasn't 1.0 until it was finalized -- possibly it didn't even have a version number until then? -- so it would have failed the semver criteria.

I guess I don't know much about how JSRs are usually versioned. I know there's a jsr305 "version 3," but I don't know that it removed anything, so possibly it's not following semver at all? Maybe the problem is that someone slapped 1.0 on jsr305 even though it wasn't finalized?

But this sounds kind of like we're just saying: "If it doesn't follow semver, then you still need to consider stability. Some JSRs use semver; JSR305 is weird because someone called in 1.0 even though it wasn't."

Sorry, I'm not sure what actionable comment I'm trying to make here.

- Maybe, if the weird thing is that someone called JSR305 "1.0," we should say that (except that it doesn't really cause stability problems per se, just the module weirdness, so maybe we don't even need to).
- If people have been doing this kind of thing with other JSRs, then maybe there's a more general point. Do we know whether they have been?
- Maybe we should call out that unreleased JSRs are an example of something that might not even have a version number, and that's another smell that something might be unstable?

library-best-practices/JLBP-4.md
cpovirk: Agreed, we don't need to wait to finalize this item (or at least to create it -- we can always edit later :))

For our current purposes, how would you feel about phrasing like...

  - Depending on unstable features between submodules of a single library is acceptable, provided that users can force their build system to use compatible versions of the submodules easily, e.g. by having the library use the same version number for all submodules, or by providing a BOM.

library-best-practices/JLBP-4.md
```diff
@@ -0,0 +1,44 @@
+[JLBP-4] Avoid dependencies on unstable libraries and features
+--------------------------------------------------------------
+
+- Unstable libraries are defined as libraries that allow for breaking changes to
```
elharo: "Unstable libraries are defined as libraries that allow for" --> Unstable libraries allow

library-best-practices/JLBP-4.md
```diff
@@ -0,0 +1,44 @@
...
+  their Public API within the same major version. For libraries following
+  semver, this means libraries with a 0.x.y version. (See [JLBP-3](JLBP-3.md)
+  for more details on the recommendations for following semver.)
+- Unstable features are defined as features that are not part of the stable
```
elharo: delete "defined as"

library-best-practices/JLBP-5.md
```diff
@@ -0,0 +1,25 @@
+[JLBP-5] Avoid dependencies that overlap classes with other dependencies
+------------------------------------------------------------------------
+
+- Definition: When two artifacts (with either a distinct group or distinct
+  artifact id) "overlap classes," that means that there is
+  at least one class identity (same package and same name) provided by both
```
elharo: class identity  --> fully qualified class name

library-best-practices/JLBP-5.md
```diff
@@ -0,0 +1,25 @@
...
+  of the artifacts.
+- This most commonly happens when a library has published the same classes
+  under multiple artifact names. Always use the primary variant.
```
elharo: "under multiple artifact names" --> in multiple artifacts.

library-best-practices/JLBP-4.md
garrettjonesgoogle: JSRs just use status labels AFAICT: https://jcp.org/en/introduction/glossary#status ; the status labels are what I'm trying to refer to. Essentially: even if the *implementation* of a JSR is 1.x, if the JSR itself is in status Dormant or Active (instead of Final), then avoid depending on it because the specification that it's based on can change (meaning the implementation would need to bump to 2.x or whatever).

library-best-practices/JLBP-4.md
garrettjonesgoogle: Why?

library-best-practices/JLBP-4.md
garrettjonesgoogle: I will use a variant of what you proposed.

library-best-practices/JLBP-4.md
garrettjonesgoogle: I will try to add more clarification to distinguish JSR status (which doesn't use semver) from JSR implementations (which would use semver).

garrettjonesgoogle:
More feedback addressed, PTAL

library-best-practices/JLBP-4.md
```diff
@@ -0,0 +1,57 @@
...
+  - If you expose the unstable feature on your library's surface, then your
+    library's current major version will be permanently locked to the version
+    you initially exposed, and you won't be able to upgrade without making a
+    breaking change to your users.
+  - If you only use the unstable feature in your implementation, then each minor
+    or patch version of your library will require a very specific version of
+    your dependency, and it will be unsafe for your users to upgrade your
+    library on its own, creating opportunities for hard-to-diagnose runtime
+    conflicts for users.
+- Given the consequences of depending unstable features in dependencies, avoid
```
cpovirk: "...depending _on_ unstable..."

library-best-practices/JLBP-5.md
```diff
@@ -0,0 +1,25 @@
+[JLBP-5] Avoid dependencies that overlap classes with other dependencies
+------------------------------------------------------------------------
+
+- Definition: When two artifacts (with either a distinct group or distinct
+  artifact id) "overlap classes," that means that there is at least one
+  full-qualified class name provided by both of the artifacts.
```
cpovirk: full-qualified -> fully qualified?

library-best-practices/JLBP-4.md
garrettjonesgoogle: oops! Will fix.

library-best-practices/JLBP-5.md
garrettjonesgoogle: oops, will fix.

library-best-practices/JLBP-4.md
```diff
@@ -0,0 +1,57 @@
+[JLBP-4] Avoid dependencies on unstable libraries and features
+--------------------------------------------------------------
+
+- At a high level, don't depend on things that can have frequent breaking
+  changes (where frequent means more than once every couple years).
```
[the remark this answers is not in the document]
garrettjonesgoogle: Do you think there is a concrete change I need to make here?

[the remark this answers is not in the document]
cpovirk: I guess the main thing is that this seems too binary. "Oh, it makes breaking changes every 3 years, so it's fine to depend on" sounds like a recipe for pain, especially when multiplied by _n_ dependencies.

The main message I'd send is probably:

- Don't depend on anything that makes breaking changes unless you have a very good reason.
- If you think you need to depend on something that makes breaking changes, try advocating for them to stop. I have some experience being the offender here, so I can offer some concrete tips:
  - If necessary and practical, provide resources to help them reach the point at which they're comfortable stopping.
  - Maybe show them data on the changes they've made historically to help them judge how important breaking changes are to them nowadays.
  - Maybe ask us for horror stories about how people are never going to upgrade because it's hard enough to coordinate major-version bumps within a closed ecosystem, let alone across ecosystems.
- If you do need to depend on something that makes breaking changes, then at least try to get its owners to deprecate things well in advance of removal (and maybe batch together the removals into fewer, larger breaking changes, too).

library-best-practices/JLBP-4.md
```diff
@@ -0,0 +1,57 @@
+[JLBP-4] Avoid dependencies on unstable libraries and features
+--------------------------------------------------------------
+
+- At a high level, don't depend on things that can have frequent breaking
```
cpovirk: Oh, and separately: It's probably worth distinguishing the projects that make "breaking changes" only when they change Java packages and artifacts from the projects that make _true_ breaking changes, since the latter are more dangerous. I know that there's going to be a whole item about this from the perspective of the library owner, but it could be useful to discuss here from the perspective of the library user.

library-best-practices/JLBP-4.md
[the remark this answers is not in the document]
garrettjonesgoogle: Hmm you seem to be talking about advocacy for robust practices, which seems to be outside the original scope of this JLBP, which is just to avoid dependencies on libraries that don't follow robust practices. There is a candidate best practice (JLBP-10 - Infrequent major version bumps) that I think covers your idea here (and should probably incorporate your specific points). Could we get by with a much more limited adjustment for the purposes of the present JLBP? Maybe:

"At a high level, don't depend on things that can have frequent or unnecessary breaking changes. Here, 'frequent' means more than once every couple years, and 'unnecessary' means breaking features that aren't causing problems."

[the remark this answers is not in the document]
cpovirk: Sorry that I continue to be so scattered about this -- and that I'm thinking through much of this only now :(

> Hmm you seem to be talking about advocacy for robust practices, which seems to be outside the original scope of this JLBP, which is just to avoid dependencies on libraries that don't follow robust practices.

Yes and no. I do agree that the details don't belong in this page. What I think does belong here is some encouragement to users to advocate when possible. It was apparently very easy for people to go "Well, Guava's unstable, but I like Guava, so I'll just use it," as if "talk to the Guava team" was some foreign, non-engineering task. OK, that's not fair: We did of course have a number of people talk to us, though often not in the most helpful ways -- hence all the details I mentioned above, which yeah, should wait for JLBP-10. I'm not trying to make this out to be easy, and ultimately the main problem was us on the Guava team, but someone could probably have helped us along earlier, and that may well have had greater impact than whatever else that person was doing at the time. (To reiterate, the main problem was still our fault! I just additionally place a small amount of fault on people who behaved as if they had no choice but to depend on an unstable API.)

But anyway, that's only one option. The other two things I would suggest for here are:

1. It's possible to shade or fork. We probably don't want to get into that deeply, but it's an option if you really do need to use something unstable. (Maybe gRPC did this with Netty or something??? I may be thinking of the wrong example.)

2. I think "more than once every couple years" is giving people permission to depend on libraries that break every few years, and I think that's going to cause problems. I would probably concede that less frequent breakages are better -- but I could Devil's Advocate for the opposite, arguing that it's better to depend on something that breaks frequently so that people see the risks in action and can't minimize them :) -- but I don't think we should give a number that suggests "no problem if you clear this bar; problem if you don't."

(It's possible that I'm too focused on the Guava case here. Are there other experiences with major-version updates that you've had that I should be thinking about in addition to Guava?)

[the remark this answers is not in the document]
cpovirk: (Maybe the summary here is: "The Guava team made bad decisions. Here's how you can cope with them so that you don't end up in the blast radius.")

library-best-practices/JLBP-4.md
```diff
@@ -0,0 +1,57 @@
...
+  a specific version of your dependency.
+  - If you expose the unstable feature on your library's surface, then your
+    library's current major version will be permanently locked to the version
+    you initially exposed, and you won't be able to upgrade without making a
+    breaking change to your users.
+  - If you only use the unstable feature in your implementation, then each minor
+    or patch version of your library will require a very specific version of
+    your dependency, and it will be unsafe for your users to upgrade your
+    library on its own, creating opportunities for hard-to-diagnose runtime
+    conflicts for users.
```
briandealwis:
```
  - If your library's release cadence does not match that of your dependencies,
    then your library may be the cause of dependency conflicts for consumers of your library.
```

This is the problem we've seen with Eclipse projects, where we were shipping 3 different versions of Guava at one point.

library-best-practices/JLBP-4.md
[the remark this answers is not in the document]
garrettjonesgoogle: I understand why you want the advocacy - I want it too. Maybe we could even create a JLBP for it - "advocate for best practices with your dependencies." However, it's a meta-action. Anyway, these JLBP's may become a launch requirement down the road, such that for any new libraries, no advocacy is required. 

1. Yes, grpc did create a variant of grpc-netty called grpc-netty-shaded. (As an aside, having both doesn't create conflicts, because they have a specific design that still works when both are present.) For shading in general, I'm uneasy about recommending it because of all the drawbacks it has (such as the fact that size can really explode). I agree that we shouldn't go deeply into that here; I think I want to create a JLBP later on the topic of shading (which will generally recommend against it). I'd honestly want someone to read the present JLBP, think "hey, now I can't do X," then talk to us, as opposed to the scenario where we recommend something with difficult tradeoffs, and have it over-applied. You could argue that having a user shade a dependency is still complying with the rules here anyway (even if we don't mention it), because it's no longer an artifact dependency (since it's bundled). 

2. If we don't give a time frame, then "frequent" will be left up to people's imagination. Some people might interpret that as "once every month," and conclude that it's fine to depend on something that only breaks surface once per year. I think saying once every five years is much too conservative for some cases (e.g. API clients, where the underlying APIs themselves can actually deprecate things in a year's time frame). I was trying to qualify the "frequent" thing by saying that it also needed to be the case that a library wouldn't break surface "unnecessarily," such that a library that technically didn't break "frequently" (say they only break once very two years) would still not be suitable if those breakages were just surface refinements and not really necessary. Let me know if that qualification doesn't come across strong enough, and let me know if that requirement still isn't sufficiently satisfying.

library-best-practices/JLBP-4.md
garrettjonesgoogle: True, but arguably, it's not great to depend on something that creates a new package with every breaking change, and they do so every six months. That is littering the namespace with packages and bloat. Additionally, such a library will be less likely to continue supporting the old major versions than a library that doesn't do that. Do you think there is a different choice library users should make depending on this distinction? We'd probably have to consider all combinations:

* frequent changes, no renaming (old Guava)
* frequent changes, with renaming (I don't personally have examples)
* infrequent changes, no renaming (new Guava)
* infrequent changes, with renaming (ok http)

My guess is that it's just the frequency dimension that matters (but I'm interested to hear if you disagree).

library-best-practices/JLBP-4.md
garrettjonesgoogle: What exactly do you mean by release cadence here? I don't think having a dependency with a faster release cadence of minor version bumps would be a problem - it seems that this only matters for major version bumps.

library-best-practices/JLBP-4.md
garrettjonesgoogle: Seeing the comment by @briandealwis , and thinking about frequency: Maybe the right thing to say is "Don't depend on something that can make breaking changes more frequently than the cadence you want for your own library." That covers the core reason for the rule, and leaves open the exact frequency to be elaborated in another best practice (candidate: JLBP-10).

library-best-practices/JLBP-4.md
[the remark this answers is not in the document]
garrettjonesgoogle: [comment moved] Seeing the comment by @briandealwis , and thinking about frequency: Maybe the right thing to say is "Don't depend on something that can make breaking changes more frequently than the cadence you want for your own library." That covers the core reason for the rule, and leaves open the exact frequency to be elaborated in another best practice (candidate: JLBP-10).

library-best-practices/JLBP-4.md
briandealwis: Release cadence = frequency of making releases.  Each release introduces risk of breaking change — sometimes inadvertent, but sometimes deliberate — to un- or under-documented behaviour.

library-best-practices/JLBP-4.md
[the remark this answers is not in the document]
cpovirk: > I'd honestly want someone to read the present JLBP, think "hey, now I can't do X," then talk to us, as opposed to the scenario where we recommend something with difficult tradeoffs, and have it over-applied.

I think that's a good principle. I think, though, that "it's OK if the API breaks only every couple years" is itself an example of a recommendation with difficult tradeoffs. More on that in a second.

> I think saying once every five years is much too conservative for some cases (e.g. API clients, where the underlying APIs themselves can actually deprecate things in a year's time frame)

Thanks for bringing this back up. I know you've mentioned it before, but I default to thinking in terms of Guava, not the API libraries.

> "Don't depend on something that can make breaking changes more frequently than the cadence you want for your own library."

I like this way of approaching it, although I think this exact phrasing is still a little optimistic.

If I want to be stable for 2 years, and I depend on a library that's stable for 2 years, I'm probably not giving 2 years of stability to my users because the two schedules likely aren't synchronized. I'm giving more like 1 year on average. (As usual, users can probably use my library for longer than that, since the specific breakages might not affect them.) And that's if I have one such dependency: If I've got 3 dependencies, we're talking more like 6 months. And of course users might take a few months to adopt my new versions, so maybe they get only ~3 months, rather than the ~21 months I'd expect if I subtract 3 months from my "2-year" guarantee.

So the full story here involves some math that isn't really _that_ hard (something like "add up the number of breakage events per year") but is more complex than `min(myStability, dep1Stability, dep2Stability, ...)`. I do very much like the idea that, say, gRPC and google-cloud-vision may have different tolerances for breakages, so I think we should include that. But if we want to give a rough mathematical rule, I'd say that you're _adding_ all your dependencies' breakage rates to your own, not taking the min. If you break things frequently, anyway, then that's dominated by your own rate, and you don't need to worry much about your dependencies. But if you're gRPC, even if you're OK with breaking every 5 years, you don't want Guava, GSON, and OpenCensus to all be breaking every 5 years, too, unless you can synchronize those schedules.

Discussion-starter (discussion-continuer? :)) proposal: "Roughly speaking, your library can breaks its users every time it makes a breaking change plus every time one of its dependencies makes a breaking change. If you think you need to depend on a library but its breakage rate makes you nervous, contact Foo-Team to discuss options."

library-best-practices/JLBP-4.md
cpovirk: Good point about the matrix of possibilities.

I do think that renaming matters in at least some cases. As in our other thread, part of the difference may be the Guava perspective vs. the Google Cloud Library perspective. If Guava had to depend on an unstable HTTP library, I'd much prefer that it be the with-renaming OkHttp model. We might not want to _switch_ to the new OkHttp every time it comes out (especially if releases were frequent, though as you point out, that's not the case for OkHttp), but we'd like the idea that we can keep using OkHttp 1.0 forever, even as the rest of the world moves onward to OkHttp 42.0. Or we can move to OkHttp 42.0 if that's better for the ecosystem. But at least we have a choice. The package bloat is sad but I still think less sad than package conflicts.

Also: You noted that "frequent changes, with renaming" seems rare. If "renaming" typically implies "infrequent changes," then, practically speaking, I think we can say good things about renaming, and users will take the action that we want. If someday OkHttp does decide to release version 42.0, we can revisit :)

Hmm, you know, new line of thinking: I'm probably partially just uncomfortable with referring to what OkHttp does as "breaking changes" at all. I mean, it _is_ in the sense that, if you remove the OkHttp 1 jar from your classpath and add the OkHttp 2 jar, your code breaks :) But it would be great to have some terminology to distinguish this from what I've called "true" breaking changes (perhaps ironically, since "true" breaking changes might or might not break users, but package renames definitely do!). I'm not sure what to call them... "in-place breaking changes" vs. something? For purposes of this doc, maybe we can just say "dependencies that release incompatible versions?" But arguably OkHttp1 and OkHttp2 are "incompatible" in that the types don't interoperate....

library-best-practices/JLBP-4.md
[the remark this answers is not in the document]
garrettjonesgoogle: Yeah, that's a good point about the non-alignment of release cycles. I'll work on some wording and push an edit.

garrettjonesgoogle:
@elharo and @cpovirk take a look at the rescoped version. (Sorry I had to rebase - there is only 1 new commit though)

cpovirk:
I think this is a great place to draw the line. Thanks!

It doesn't cover Guava, but surely no one will be dumb enough to do what Guava did again ;)

(LGTM; maybe I should see about getting myself added to this project as dpb did so that I can be a "real" reviewer.)

library-best-practices/JLBP-4.md
```diff
@@ -0,0 +1,31 @@
...
+  Public API within the same major version. For libraries following semver, this
+  means libraries with a 0.x.y version. (See [JLBP-3](JLBP-3.md) for more
+  details on the recommendations for following semver.)
+- Unstable features are features that are not part of the stable Public API of a
+  stable library, often marked with annotations like `@Beta`.  (See
+  [JLBP-3](JLBP-3.md) for more details on the recommendations for annotating
+  unstable features.)
+- If your library depends on an unstable library or feature, and that feature
+  experiences a breaking change between versions, your library will be locked to
+  a specific version of your dependency.
```
elharo: your dependency --> the dependency

garrettjonesgoogle: I don't feel like that reads quite right...

briandealwis:
[This tweet](https://twitter.com/garybernhardt/status/1035694098739486721?s=21) and its reply seemed somewhat relevant:

screen shot 2018-09-01 at 12 02 27 am

PR #62, opened 2018-08-31
Use homegrown dependency lister instead of Maven enforcer plugin in dashboard

elharo:
This will enable us to provide much more customized output.


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
```diff
@@ -104,9 +104,9 @@ void addPath(DependencyPath path) {
     return versions.get(coordinates);
   }
 
-  // TODO consider whether we need an update class with parent, from, and to, and
+  // TODO consider whether we need an update class with parent, from, and to; and
   // a toString method rather than returning strings
-  List<String> findUpdates() {
+  public List<String> findUpdates() {
```
briandealwis: This would benefit from javadoc as to what it returns.  It seems to return a list of human-meaningful strings of updates for the set of configured artifacts in a form that can be dropped into an HTML document (without further encoding? or does freemarker handle that?).

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -69,7 +66,8 @@ public void testGenerateDashboard()
         Assert.assertTrue(fileName + " is missing", Files.isRegularFile(componentReport));
         try {
           Document report = builder.build(componentReport.toFile());
-          // TODO test the content of the report
+          Nodes updates = report.query("//li");
+          Assert.assertTrue("didn't find updates", updates.size() > 2);
```
briandealwis: It's intended that this test is non-hermetic, right?

elharo: You mean, it's intended that it talks to Maven central? Yes, that is intended.

dependencies/pom.xml
```diff
@@ -1,16 +1,17 @@
...
     <groupId>com.google.cloud.tools.opensource</groupId>
     <artifactId>cloud-tools-opensource-parent</artifactId>
     <version>1.0-SNAPSHOT</version>
   </parent>
-  <artifactId>dependency-lister</artifactId>
-  <packaging>jar</packaging>
 
   <name>Dependency Lister</name>
   <url>https://github.com/GoogleCloudPlatform/cloud-opensource-java/dependencies</url>
+  <artifactId>dependency-lister</artifactId>
```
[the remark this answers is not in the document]
elharo: I did, don't remember why though. Anyway, adjusted.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
elharo: done. I don't think this escapes right now. I'm pretty sure the return type is going to change t something more structured as per the TODO comment so I don't want to steel-plate this right now.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
briandealwis: What I really meant was that this test will break when there's an update published.

PR #63, opened 2018-08-31
Adding JLBP-6 (Avoid creating conflicts between major versions)

garrettjonesgoogle:


library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,141 @@
...
+- *If* the Java package is changed, use different maven coordinates (for example
+  different artifact id).
+- *If* the Java package is kept the same, use the same maven artifact ID.
+- In other words, rename the Java package if and only if the maven coordinates
+  change.
+
+Background
+----------
+
+When a library B in Java depends on another library A using Apache Maven, there
```
elharo: seems to be about using the maven repo system, not necessarily maven

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,141 @@
...
+    `com.google.a1`
+  - With a rename, the new major version's Java package is `com.google.a2`
+
+Given this scenario, here are the possible combinations of renamings:
+
+- **Don't rename Java package**:
+  - **Don't rename Maven coordinates (case 1)**: This approach can result in
+    diamond dependency conflicts because different branches of a dependency tree
+    can depend on different major versions, and only one of them can be pulled
+    in by maven in a particular dependency tree. Users are forced to change
```
elharo: We should try to be consistent about capitalizing Maven or not.

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,141 @@
...
+    their code between versions.
+  - **Rename maven coordinates (case 2)**: Users can easily pull in both the old
+    jar (`g1:a1:1.0.0`) and the new jar (`g2:a2:2.0.0`) accidentally through
+    transitive dependencies, because maven treats the two artifacts as distinct,
+    but since they have classes with the same Java packages, they are loaded in
+    unpredictable ways, leading to runtime exceptions (for example
```
elharo: I'm not sure it's unpredictable but it can be surprising.

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,141 @@
...
+    `ClassNotFoundException`). It is difficult for users to ensure their build
+    tree only includes one of the artifacts. **NEVER DO THIS**. Example: guava
+    vs guava-jk5.
```
elharo: jdk5

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,141 @@
...
+- **Rename Java package**:
+  - **Don't rename Maven coordinates (case 3)**: The classes from `g1:a1:1.0.0`
```
elharo: does junit do this? I think that latest junit jar includes the old 3.x packages too. Worth checking.

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,141 @@
...
+(renaming the Java package while keeping the maven coordinates the same), and
+both should be avoided.
+
+Of the remaining two, case 4 (renaming both maven coordinates and the Java
+package) is the least disruptive and the only option that won't lead to diamond
+dependency conflicts, so it should be the generally preferred approach. However,
+given that a full transition from one major version to the next requires
+updating all import statements regardless of whether the corresponding classes
+have changed, there is potentially a very high cost to this option, which should
+be weighed carefully. Case 1 (neither renaming the maven coordinates nor the
```
elharo: I feel like case 1 is the default today. Package renames are a mess. I'm not sure we should recommend against as strongly as we do here.

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,141 @@
...
+*When making breaking changes between major versions, prefer to rename both the
+maven coordinates and the Java package. However, if the ratio of surface
+breakage to library size is sufficiently low and the number of usage points of
+breaking code is sufficiently low, keeping the same maven coordinates and Java
+package can be preferred instead. Under no circumstance should only one of the
+two be renamed without the other being renamed also.*
+
+Example renames in open source 
+[Square has established it as a policy](http://jakewharton.com/java-interoperability-policy-for-major-version-updates/)
+for its Java libraries (examples include OkHttp and Retrofit). Numerous open
```
elharo: I feel like far more have not renamed the packages.

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,141 @@
...
+  than dealing with the fallout of dealing with diamond dependency conflicts for
+  that particular method. Only perform an in-place breakage if users can easily
+  replace their usage of the old library with the new library. The larger the
+  portion of a library that breaks between major versions, the higher the value
+  is from renaming the java package.
+- *If* the Java package is changed, use different maven coordinates (for example
+  different artifact id).
+- *If* the Java package is kept the same, use the same maven artifact ID.
+- In other words, rename the Java package if and only if the maven coordinates
+  change.
```
briandealwis: Need to specify what is a _coordinates change_ as even bumping the micro version is a coordinate change.  Which contradicts the first point (line 4).

library-best-practices/JLBP-6.md
garrettjonesgoogle: Good point, will fix.

library-best-practices/JLBP-6.md
garrettjonesgoogle: This is probably true, but I didn't collect the evidence. Do you have some examples that I can put in? (Guava is obviously one)

library-best-practices/JLBP-6.md
garrettjonesgoogle: Ahh, good point about the meaning of "coordinates" for Maven - I wasn't being careful there. I think many of my uses of the word "coordinates" in this doc are misleading, so I'll fix that.

library-best-practices/JLBP-6.md
garrettjonesgoogle: If the old version is bundled with the new version, this has the same impact as described in case 4 (they can be used side by side). I will figure out how to integrate this idea.

garrettjonesgoogle:
@elharo PTAL again. @cpovirk care to take a look?

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,152 @@
...
+   a particular class within the jar pulled from Maven Central.
+
+Now when breaking changes are introduced to library A between major version 1
+and major version 2, a choice needs to be made: to rename or not rename? This
+question applies to both items listed above, the Maven ID (1) and the Java
+package (2). Consider the following renaming scenario:
+
+- The Maven coordinates start as `g1:a1:1.0.0`
+  - With no rename of the Maven ID, the new major version's Maven coordinates
+    are `g1:a1:2.0.0`. (Only one of the versions can be present in the
```
elharo: depends on what you mean by "present". Better yet, only one of these versions appears in the classpath, even if both are in the dependency tree.

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,152 @@
...
+    `g2:a2:2.0.0`. (Both versions can be present in the dependency tree.)
+- The Java package starts as `com.google.a1`
+  - With no rename, the new major version's Java package is still
+    `com.google.a1`
+  - With a rename, the new major version's Java package is `com.google.a2`
+
+Given this scenario, here are the possible combinations of renamings:
+
+- **Don't rename Java package**:
+  - **Don't rename Maven ID (case 1)**: This approach can result in diamond
```
elharo: Honestly, this can happen even in minor version updates too; for instance if your library depends on a new, non-breaking method addition introduced in a new minor version.

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,152 @@
...
+    This has the same impact as case 4 below, except with the slight drawback
+    that the user's class space is polluted with both versions, whether both are
+    used or not.
+  - **Rename Maven ID (case 4)**: The two major versions can be used side by
+    side, allowing users to incrementally transition from the old to the new
+    version, or even use them side by side indefinitely if
+    necessary. Transitioning fully to the new version does require code changes
+    between versions though, even for classes whose surface remains the
+    same. There will be no conflicts using this approach. This is the approach
+    taken by Go, described in https://research.swtch.com/vgo-import .
```
elharo: hot link?

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,152 @@
...
+Example renames in open source 
+[Square has established it as a policy](http://jakewharton.com/java-interoperability-policy-for-major-version-updates/)
+for its Java libraries (examples include OkHttp and Retrofit). Numerous open
+source packages have performed such a rename between large major version
+updates:
+- OkHttp (com.squareup.okhttp -> com.squareup.okhttp3)
+- Apache Commons Lang (org.apache.commons.lang -> org.apache.commons.lang3)
+- RxJava (rx (version 1.x) -> io.reactivex (version 2.x))
+- JDOM (org.jdom -> org.jdom2)
+- JUnit (junit.framework (versions 1.x-3.x) -> org.junit (version 4))
```
elharo: This may be wrong. JUnit 4 is case 1. They did not change the package name and did not change the artifact and group IDs. They *added* a new org.junit package but kept the old junit.framework package too.

JUnit 5 is more complicated, and arguably something we don't consider here. They've now broken the product into several jars and multiple artifacts with new IDs and packages. However they do have a junit-vintage artifact that includes backwards compatibility.

garrettjonesgoogle: For JUnit 4, if the "new" JUnit provided was under a new package name (`org.junit`), I would argue that they did a rename of the package. The fact that the old `junit.framework` was also included seems like a distribution strategy to me (it's at the packaging level, not the code level). Thus, I think case 1 is ruled out for JUnit. 

I think the text of case 3 includes the JUnit scenario (at least I tried to make it do so), but I think now that it's potentially confusing. I think I'll create a case 5 for it (rename Java package, bundle old & new in the same maven ID) so that it's clearer.

garrettjonesgoogle:
@elharo I spent some time thinking about your recommendation. Accepting it has some pretty fundamental consequences for solving diamond dependency conflicts. Essentially it would mean we are accepting the occurrence of conflicts as a matter of course, and we will need to lean more heavily on making it easy and robust to pull in versions of artifacts that don't conflict (e.g. using BOMs). 

Given that @saturnism has previously made the claim that diamond dependency conflicts are a fact of life for open source software (and we will need to figure out how to live with it), and @ejona86 has also expressed concerns about package renames, I think it's the approach we will need to take. I was hoping we could tilt more in the direction of reducing conflicts globally, but it appears such a desire was quixotic. 

I will rename the JLBP as "Package and artifact renaming rules" to indicate that a specific rule is not specifically recommended, but that certain renaming combinations should still be avoided (specifically cases 2 and 3).

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,152 @@
...
+----------
+
+When a library B in Java depends on another library A through the Maven
+repository system, there are two handles that library B needs to use to access
+library A:
+
+1. The Maven coordinates, following the form
+   `${groupId}:${artifactId}:${version}`, for example
+   `com.google.guava:guava:26.0-jre`. The Maven coordinates are used to pull the
+   library's jar from Maven Central. Given a group ID and artifact ID, Maven
```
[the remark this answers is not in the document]
garrettjonesgoogle: That was my intended meaning - I will make it clearer.

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,152 @@
...
+   will choose only one version to pull in.
+2. The Java package (the package used in `import` statements), for example
+   `com.google.common.collect.ImmutableList`. The Java package is used to select
```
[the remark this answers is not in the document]
garrettjonesgoogle: Yeah this wasn't clearly stated. I will make it clearer.

library-best-practices/JLBP-6.md
garrettjonesgoogle: I'll change this wording, thanks.

library-best-practices/JLBP-6.md
garrettjonesgoogle: It is automatically hotlinked by Github's md rendering

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,152 @@
...
+   a particular class within the jar pulled from Maven Central.
```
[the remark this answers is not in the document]
garrettjonesgoogle: I'll make this more precise.

library-best-practices/JLBP-6.md
ejona86: I will note that Go is a different language and with different practices, so the choice does play out differently. Namely, unchanged interfaces between the two versions will still apply to the same structs. In Java you end up getting only the old or only the new interface. In Go, additionally, it is trivial to create an interface for existing structs which means it is much easier to avoid exposing your dependencies to your users, even those needing to be injected. Also, changing the package name in Go is pretty easy and you can do it files at a time and there's even some tooling to help. In Java, it'd be much more painful given the relative sizes of Java modules and the tightly-coupled interaction between packages (in our user's code, not in the libraries).

Basically, Java depends _very strongly_ on the inheritance hierarchy, even for interfaces. And the inheritance hierarchy is broken when doing a rename.

(This should not be construed to say that we shouldn't do case 4. I think case 4 is appropriate at times. But I think the pain in the approach is higher than in Go and so that may cause us to use the approach less frequently in Java.)

garrettjonesgoogle:
@elharo & @cpovirk I have done a rewrite, PTAL

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,165 @@
...
+package (2).
+
+Recommendations:
+
+- *If* the new library surface is delivered under a new Java package, either use
+  a different Maven ID (different group ID or artifact ID) or bundle the old and
+  new packages together in the under the original Maven ID.
+- *If* the breaking change is made in-place and the Java package is kept the
+  same, use the same Maven ID (same group ID and same artifact ID).
+- There is a tradeoff: if an extremely large library (for example thousands of
```
cpovirk: It might not be clear that this is a tradeoff between two options presented above. Maybe explicitly say "a tradeoff between the two options presented above," or modify the previous two bullets to be nested under a top-level bullet that says something like "When making a breaking change, take one of two approaches." (The latter would also let the final bullet become one of two top-level bullets, which may make sense, since (as it says) it's not solely about breaking changes.)

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,165 @@
...
+  classes) breaks a single method, the cost of a package rename will be higher
+  than dealing with the fallout of dealing with diamond dependency conflicts for
+  that particular method. The larger the portion of a library that breaks
+  between major versions, the higher the value is from renaming the Java
+  package.
+- Whether making breaking changes or making no breaking changes, don't publish
+  the same classes under multiple Maven IDs. Or put another way: Once you have
```
cpovirk: Any idea how often this comes up? If it's a nod at guava_jdk5, it may be a bit oblique. But you may have other examples in mind.

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,165 @@
...
+    is that Go can accept types with the same structure even though they have
+    different identities, so this approach is more disruptive in Java. In Go, a
+    single method can accept types from multiple interface declarations, but
+    Java can accept only one. Transitioning fully to the new version in Java
+    requires code changes between versions, even for classes whose surface
+    remains the same, because all import statements need to be updated. There
+    will be no diamond dependency conflicts using this approach, but adoption
+    can be blocked if there are consuming libraries that have not also created
+    new major versions that can accept new types from this library.
+  - **Bundle old and new in the existing Maven ID (case 5)**: Like case 4, the
```
ejona86: This may be a bit too specific. I am assuming this is when you have a shim from the old API to the new API. In that case I'd tend to suggest the old artifact (with old package) to depend on a new artifact (with new package). It does mean there are now two artifacts to choose from, but if you choose the old name things "just work."

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,165 @@
...
+    depend on different major versions, and the build system (Maven or Gradle)
+    will only choose one of them to use. Users are forced to change their code
+    between versions, but only for library surface that was changed.
+  - **Rename Maven ID (case 2)**: Users can easily pull in both the old jar
+    (`g1:a1:1.0.0`) and the new jar (`g2:a2:2.0.0`) accidentally through
+    transitive dependencies, because Maven treats the two artifacts as distinct,
+    but since they have classes with the same Java packages, they are loaded in
+    unpredictable ways, leading to runtime exceptions (for example
+    `ClassNotFoundException`). It is difficult for users to ensure that their
+    build tree only includes one of the artifacts. **NEVER DO THIS**.
```
elharo: I think the servlet API did exactly this. Was there a good reason we're missing here (e.g. provided jars) or did they screw up?

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,165 @@
...
+    `g1:a1:2.0.0` could technically be used together, but since they share the
+    Maven group ID and artifact ID, only one jar can be pulled in. Users are
+    forced to change their code if they want to add any references to the new
+    version. It is strictly better to either also rename the Maven ID (case 4)
+    or to keep the Maven ID and also bundle the old package (case 5), so that
+    the major versions can be used side by side.
+  - **Rename Maven ID (case 4)**: The two major versions can be used side by
+    side, allowing users to incrementally transition from the old to the new
+    version, or even use them side by side indefinitely if necessary. This is
+    similar to the approach taken by Go, described in
```
elharo: This document's audience is Java developers. Are we just muddying the waters by talking about Go?

ejona86:
I agree with @elharo's view that package renaming is sort of like making a completely new library. But I think there's a spectrum involved and in some cases it will not be obvious to go with case 1 or case 4. I think we're mostly all on the same page, but may disagree at which point on the spectrum which case wins.

I would put one _very_ strong requirement on case 1 migrations. Most libraries/applications using your library must be able to have code that supports the old and new version simultaneously. That generally would mean there are "stepping stone" releases that introduce any new required APIs into the old version. The major release could then _just_ drop deprecated APIs.

For example, this was done with OkHttp 1→2 in the form of [OkHttp 1.6](https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-160). When I did the upgrade I was able to pre-migrate most of Google and then have a single CL that did the swap-over. Only a _single_ source file had to be updated atomically because it couldn't be pre-migrated (cl/74800711).

Let's say we're introducing a 2.x version. How I'd expect this to play out in OSS is that libraries stay on the last 1.x version for a year or two waiting for other libraries and applications to become compatible with the 2.x version. Then libraries would begin bumping to the 2.x version. The more widely used the library and the greater the breakages introduced in 2.x increases the amount of time this step takes.

At some point it becomes easier to just consider it a new library and swap to "case 1". Having a shim implementation of the 1.x API that uses the 2.x API can greatly reduce migration burden in this case.

I do view case 4 as creating a new library, because I've seen many times code in maintenance mode (which is most code, even within healthy codebases) doesn't upgrade such that the old version could still around a decade later. I've seen this for JUnit 3→4, Apache HTTP Client → HTTP Components, (and to a lesser extent; time will tell) OkHttp 2→3, Netty 3→4. I've seen very few users migrate from OkHttp 2 to 3. The majority of migrations away from OkHttp 3 I've seen were migrations to a different HTTP library.

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,165 @@
...
+
+**Case 1**
+- Guava
+- Hibernate
+- Joda Time
+
+**Case 2**
+- `guava` vs `guava-jdk5`
+  - This technically wasn't a new major version, but it is an example of case 2
+    that has caused a lot of problems.
```
elharo: servlets!

library-best-practices/JLBP-6.md
```diff
@@ -0,0 +1,165 @@
...
+
+**Case 4**
+- [Square has established it as a policy](http://jakewharton.com/java-interoperability-policy-for-major-version-updates/)
```
elharo: delete "it as";

also the link should be over "policy for its Java libraries", not Square has established as a policy

library-best-practices/JLBP-6.md
garrettjonesgoogle: Great idea, will do.

library-best-practices/JLBP-6.md
garrettjonesgoogle: Yes, guava-jdk5 is one example, but I want to highlight it because people may not be careful with this if they move code & artifacts across source repos. Basically, you need to keep the same Maven ID even if you're bundling it in a new place. Should I mention that here?

cpovirk: Thinking about this more, I realize that I have seen other cases in which people fork or repackage existing projects without changing the package. (People have released per-package Guava artifacts, for example, and various people have made jsr305 available.)

Maybe there's a way to re-frame this to cover that, too? The current phrasing talks in terms of "you" as if you're the only person who did any publishing, but I'm thinking of scenarios in which multiple people publish the same classes.

All that said, we have much bigger fish to fry than the phrasing here, so feel free to just keep what you have. I may have just interpreted it through the lens of guava_jdk5. Other readers are less likely to have that particular lens.

elharo: That's a very good point: third party republishing with a different group ID and often the same artifact ID, same package, and often ACTUAL INCOMPATIBLE BREAKING CHANGES. I need to think about that one. Don't block on me though. This may well require a separate item.

library-best-practices/JLBP-6.md
garrettjonesgoogle: Actually I had in mind the junit case, where I'm guessing they don't use a shim. I had overlooked the shim idea. Do you feel it needs to be included?

library-best-practices/JLBP-6.md
garrettjonesgoogle: Based on my (limited) understanding, they screwed up - now we have to go around advising everyone (e.g. in JLBP-5) to use `provided` as the scope for servlet-api, otherwise people will see conflicts.

ejona86: Servlet API _needs_ to be provided by the servlet container itself. That's the only sane way for the container to communicate with the servlets themselves. It should always be `provided`, in the same way any AppEngine APIs and Android APIs would be `provided`; they are part of the runtime. But all of those are different from normal libraries.

library-best-practices/JLBP-6.md
garrettjonesgoogle: Yeah, probably. I will remove it.

library-best-practices/JLBP-6.md
garrettjonesgoogle: I will add this.

ejona86:
>  I think your point about making transitions between breaking changes is important, but not really part of the renaming rules.

Ah, true. Yeah, this PR is about a finer detail about the larger process. Having it elsewhere sounds good.

garrettjonesgoogle:
Ok, I have attempted to address the latest round of feedback - PTAL @elharo & @ejona86

garrettjonesgoogle:
Since I have two approvals I am going to merge now - we can do more PRs later if we want to make additional edits.

PR #66, opened 2018-09-09
Include date time of last run at bottom of main dashboard page

elharo:
@netdpb @garrettjonesgoogle

### DIFF
--- a/dashboard/src/main/resources/templates/dashboard.ftl
+++ b/dashboard/src/main/resources/templates/dashboard.ftl
@@ -10,5 +10,7 @@
         <li><a href='${artifact?replace(":", "_")}.html'>${artifact}</a></li>
       </#list>
       </ul>
+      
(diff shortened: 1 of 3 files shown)


dashboard/src/main/resources/templates/dashboard.ftl
```diff
@@ -10,5 +10,7 @@
         <li><a href='${artifact?replace(":", "_")}.html'>${artifact}</a></li>
       </#list>
       </ul>
+      foo
+      <p id='updated'>Last generated at ${lastUpdated}</p>
```
garrettjonesgoogle: So this is essentially when the page was generated?

elharo: Correct. I mostly threw it in because otherwise a failed run generates no obvious output here. This will tell us if it's stale.

PR #67, opened 2018-09-09
add bom to top reactor

elharo:
@garrettjonesgoogle

### DIFF
--- a/boms/cloud-oss-bom/pom.xml
+++ b/boms/cloud-oss-bom/pom.xml
@@ -41,26 +41,6 @@
            grpc/grpc-java#3954 -->
       <dependency>
         <groupId>io.grpc</groupId>
-        <artifactId>grpc-netty-shaded</artifactId>
-        <version>${io.grpc.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>io.grpc</groupId>
-        <artifactId>grpc-protobuf</artifactId>
-        <version>${io.grpc.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>io.grpc</groupId>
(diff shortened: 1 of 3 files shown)


boms/cloud-oss-bom/pom.xml
```diff
@@ -79,36 +79,16 @@
...
         <version>${io.grpc.version}</version>
       </dependency>
-      <dependency>
-        <groupId>io.grpc</groupId>
-        <artifactId>grpc-stub</artifactId>
-        <version>${io.grpc.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>io.grpc</groupId>
-        <artifactId>grpc-core</artifactId>
```
garrettjonesgoogle: Why are these modules being removed?

elharo: They aren't. They were dupes and Maven complained.

garrettjonesgoogle: Could you alphabetize the list so that it's easy to see dupes?

elharo: That raises the question of whether the order of dependencies in a dependencyManagement section matters in dependency resolution. I ask because the order of dependencies in a dependency section does matter. So far, it's just the order I pasted them in, but I wonder if we need to be more careful about what comes first?

garrettjonesgoogle: Excellent question... are you interested in doing the research?

elharo: Looking over the maven docs, it doesn't seem to matter in any way I can see. I've uploaded a new version that alphabetizes the gRPC dependencies.

PR #73, opened 2018-09-14
Read BOM to find out what to put on dashboard

elharo:
@garrettjonesgoogle @netdpb If anyone can figure out a way we can run this without pre-installing the BOM locally first, that would be great.

Also, you'll notice I've explicitly filtered a couple of dependencies from the google-cloud-java BOM because they seem to contain a cycle. See https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3680 and https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3683

### DIFF
--- a/boms/pom.xml
+++ b/boms/pom.xml
@@ -3,15 +3,19 @@
      xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
                http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
+  
+  <parent>
+    <groupId>com.google.cloud.tools.opensource</groupId>
+    <artifactId>cloud-tools-opensource-parent</artifactId>
+    <version>1.0-SNAPSHOT</version>
+  </parent>
 
-  <groupId>com.google.cloud.tools.opensource</groupId>
   <artifactId>cloud-tools-opensource-boms</artifactId>
   <packaging>pom</packaging>
-  <version>1.0-SNAPSHOT</version>
   <name>BOM Parent</name>
   <url>https://github.com/GoogleCloudPlatform/cloud-opensource-java/boms</url>
   <organization>
-    <name>Google LLC.</name>
+    <name>Google LLC</name>
     <url>http://www.google.com</url>
   </organization>
 
@@ -30,25 +34,5 @@
     <module>cloud-oss-bom</module>
     <module>upper-bounds-check</module>
   </modules>
-
-  <build>
-    <pluginManagement>
-      <plugins>
-        <!-- TODO figure out why this isn't inheriting.
-             Maybe the parent isn't what I think it is. -->
-        <plugin>
-          <groupId>org.codehaus.mojo</groupId>
-          <artifactId>exec-maven-plugin</artifactId>
-          <version>1.6.0</version>
-          <configuration>
-            <skip>true</skip>
-            <exec>java</exec>
-            <executable>maven</executable>
-            <mainClass>none</mainClass>
-          </configuration>
-        </plugin>
-      </plugins>
-    </pluginManagement>
-  </build>
 
 </project>
--- a/boms/pom.xml
+++ b/boms/pom.xml
@@ -30,5 +30,14 @@
     <module>cloud-oss-bom</module>
     <module>upper-bounds-check</module>
   </modules>
+  
+  <!-- Include cloud-oss-bom/pom.xml in classpath as a resource -->
+  <build>
+    <resources>
+      <resource>
+        <directory>cloud-oss-bom</directory>
+      </resource>
+    </resources>
+  </build>
 
 </project>
--- a/boms/pom.xml
+++ b/boms/pom.xml
@@ -31,4 +31,24 @@
     <module>upper-bounds-check</module>
   </modules>
 
+  <build>
+    <pluginManagement>
+      <plugins>
+        <!-- TODO figure out why this isn't inheriting.
+             Maybe the parent isn't what I think it is. -->
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>exec-maven-plugin</artifactId>
+          <version>1.6.0</version>
+          <configuration>
+            <skip>true</skip>
+            <exec>java</exec>
+            <executable>maven</executable>
+            <mainClass>none</mainClass>
+          </configuration>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+  </build>
+
 </project>
--- a/boms/pom.xml
+++ b/boms/pom.xml
@@ -30,14 +30,5 @@
     <module>cloud-oss-bom</module>
     <module>upper-bounds-check</module>
   </modules>
-  
-  <!-- Include cloud-oss-bom/pom.xml in classpath as a resource -->
-  <build>
-    <resources>
-      <resource>
-        <directory>cloud-oss-bom</directory>
-      </resource>
-    </resources>
-  </build>
 
 </project>
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -59,7 +59,8 @@ public class DashboardMain {
   public static void main(String[] args) 
       throws IOException, TemplateException, ArtifactDescriptorException {
 
-    generate();
+    Path output = generate();
+    System.out.println("Wrote dashboard into " + output.toAbsolutePath());
   }
 
   public static Path generate() throws IOException, TemplateException, ArtifactDescriptorException {
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -48,6 +48,7 @@ import org.eclipse.aether.resolution.ArtifactDescriptorRequest;
 import org.eclipse.aether.resolution.ArtifactDescriptorResult;
 import org.eclipse.aether.resolution.DependencyResolutionException;
 
+import com.google.cloud.tools.opensource.dependencies.Artifacts;
 import com.google.cloud.tools.opensource.dependencies.DependencyGraph;
 import com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder;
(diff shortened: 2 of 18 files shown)


garrettjonesgoogle:
Is there a way to get a preview of the resulting dashboard?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -31,39 +31,47 @@
...
   
-  // todo move this to config file
-  static String[] ARTIFACTS = {
-      "com.google.cloud:google-cloud-core:1.41.0",
-      "com.google.cloud:google-cloud-datastore:1.41.0"
-  };
-  
   public static void main(String[] args) 
-      throws IOException, TemplateException {
+      throws IOException, TemplateException, ArtifactDescriptorException {
```
garrettjonesgoogle: It doesn't seem useful to declare explicit exceptions that `main` throws - why not use use `throws Exception`?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -88,6 +97,42 @@ private static void generateReports(Configuration configuration, Path output) {
     }
   }
 
+  @VisibleForTesting
+  static List<String> readBom() throws ArtifactDescriptorException {
+    
+    DefaultArtifact artifact =
+        new DefaultArtifact("com.google.cloud:cloud-oss-bom:pom:0.0.1-SNAPSHOT");
```
garrettjonesgoogle: It seems like it would be a pain to update this every time the version bumps up. Have you thought about how we could manage to control the version in a single place? In google-cloud-java, we do this using a file with versions ( https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/versions.txt ), tags throughout the repo, and a version updater script ( https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/utilities/replace_versions.py ).

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -88,6 +97,42 @@ private static void generateReports(Configuration configuration, Path output) {
...
+    List<Exception> exceptions = resolved.getExceptions();
+    if (!exceptions.isEmpty()) {
+      throw new ArtifactDescriptorException(resolved, exceptions.get(0).getMessage());
+    }
+    
+    List<String> result = new ArrayList<>();
+    for (Dependency dependency : resolved.getManagedDependencies()) {
+      Artifact managed = dependency.getArtifact();
+      if (managed.getArtifactId().contains("logback")
+          || managed.getArtifactId().contains("contrib")) {
```
garrettjonesgoogle: Add a `TODO` to remove this hack

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -89,11 +86,12 @@ private static DependencyNode resolveCompileTimeDependencies(Artifact artifact)
 
     CollectRequest collectRequest = new CollectRequest();
     collectRequest.setRoot(dependency);
-    collectRequest.addRepository(CENTRAL);
+    collectRequest.addRepository(RepositoryUtility.CENTRAL);
     DependencyNode node = system.collectDependencies(session, collectRequest).getRoot();
 
     DependencyRequest dependencyRequest = new DependencyRequest();
     dependencyRequest.setRoot(node);
+    // todo might be able to speed up by using collectDependencies here instead
```
garrettjonesgoogle: capitalize `todo`?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
elharo: I'm on the far end of the checked exceptions distribution. Declare all the things! If I accidentally introduce code that throws a new exception type, I want a red squiggly line so I can think about what I've done and whether I really want that.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
elharo: That makes sense. It's likely part of figuring out how to read from the local code instead of the repo. Haven't figure that out yet.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
elharo: done

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
elharo: done

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -88,6 +97,44 @@ private static void generateReports(Configuration configuration, Path output) {
...
+    if (!exceptions.isEmpty()) {
+      throw new ArtifactDescriptorException(resolved, exceptions.get(0).getMessage());
+    }
+    
+    List<String> result = new ArrayList<>();
+    for (Dependency dependency : resolved.getManagedDependencies()) {
+      Artifact managed = dependency.getArtifact();
+      // TODO remove this hack once we get these out of 
+      // google-cloud-java's BOM
+      if (managed.getArtifactId().contains("logback")
```
netdpb: Is there something more specific we can use here? "logback" could be a substring of many artifacts; same with "contrib".

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -88,6 +97,44 @@ private static void generateReports(Configuration configuration, Path output) {
...
+          || managed.getArtifactId().contains("contrib")) {
+        continue;
+      }
+      String coords =
+          managed.getGroupId() + ":" + managed.getArtifactId() + ":" + managed.getVersion();
```
netdpb: Don't we have `Artifacts.toCoordinates(Artifact)` for this?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -88,6 +97,44 @@ private static void generateReports(Configuration configuration, Path output) {
...
+      if (!result.contains(coords)) {
```
netdpb: Should result be a set?

netdpb: and how should the list be ordered?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -197,12 +195,20 @@ private static void fullPreorder(Stack<DependencyNode> path, DependencyNode curr
       if (!"system".equals(child.getDependency().getScope())) {
         try {
           child = resolveCompileTimeDependencies(child.getArtifact());
+          // somehow we've got an infinite recursion here
```
netdpb: Should we throw?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -28,28 +28,34 @@
...
-final class RepositoryUtility {
+public final class RepositoryUtility {
   
+  public static final RemoteRepository CENTRAL =
+      new RemoteRepository.Builder("central", "default", "http://repo1.maven.org/maven2/").build();
+
   private RepositoryUtility() {}
 
-  static RepositorySystem newRepositorySystem() {
+  public static RepositorySystem newRepositorySystem() {
```
netdpb: These probably should have Javadoc.

boms/pom.xml
```diff
@@ -3,15 +3,19 @@
...
 
-  <groupId>com.google.cloud.tools.opensource</groupId>
   <artifactId>cloud-tools-opensource-boms</artifactId>
   <packaging>pom</packaging>
-  <version>1.0-SNAPSHOT</version>
   <name>BOM Parent</name>
   <url>https://github.com/GoogleCloudPlatform/cloud-opensource-java/boms</url>
   <organization>
-    <name>Google LLC.</name>
+    <name>Google LLC</name>
```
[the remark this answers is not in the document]
elharo: Copyright notice in source code needs period. Everywhere else seems not to use period on LLC.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
elharo: done

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
elharo: done

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
elharo: Order matters in a POM

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
elharo: I thought about this, but my gut is no. This actually happens in our graph when we include contrib and/or logback which is why I dropped them out. But I don't want to fail completely and not process the remaining items in this case. Hence the short circuit here.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
elharo: done

PR #74, opened 2018-09-14
Adding index for best practices

garrettjonesgoogle:


library-best-practices/README.md
```diff
@@ -0,0 +1,19 @@
+Java Library Best Practices
+---------------------------
+
+The Java Library Best Practices are a collection of practices that we believe
```
elharo: "The Java Library Best Practices" --> These

library-best-practices/README.md
```diff
@@ -0,0 +1,19 @@
...
+will minimize problems for consumers of interconnected Java libraries.  These
+practices come from decades of aggregated experience in maintaining open source
+Java libraries and are informed by many hard-earned lessons from mistakes that
```
elharo: hard-learned

library-best-practices/README.md
```diff
@@ -0,0 +1,19 @@
...
+have been made. We have found that following these best practices will result in
```
elharo: best practices --> rules

library-best-practices/README.md
```diff
@@ -0,0 +1,19 @@
...
+higher quality Java libraries with fewer diamond dependency conflicts and other
```
elharo: delete diamond

library-best-practices/README.md
```diff
@@ -0,0 +1,19 @@
...
+kinds of problems. The list is open-ended, so new ones may be added from time to
+time.
+
+- [JLBP-1](JLBP-1.md): Minimize dependencies
+- [JLBP-2](JLBP-2.md): Minimize API surface
+- [JLBP-3](JLBP-3.md): Use Semantic versioning
+- [JLBP-4](JLBP-4.md): Avoid dependencies on unstable libraries and features
+- [JLBP-5](JLBP-5.md): Avoid dependencies that overlap classes with other
```
elharo: classes or packages?

garrettjonesgoogle: "overlapping classes" is the way we describe the problem in JLBP-5 and JBLP-6; it's when multiple packages have the same fully-qualified class names.

elharo: so it's not split packages, then? In that case, we probably do need a new JLBP to "not split packages" and we may want to reread 1-6 in this light. It's particularly important in Java 9 and later where split packages are now banned.

garrettjonesgoogle:
@elharo feedback addressed, PTAL

library-best-practices/README.md
garrettjonesgoogle: Ok makes sense, I'll queue it up.

PR #76, opened 2018-09-17
Adding JLBP-7: Make breaking transitions easy

garrettjonesgoogle:
cc @ejona86

### DIFF
--- a/library-best-practices/JLBP-7.md
+++ b/library-best-practices/JLBP-7.md
@@ -7,8 +7,8 @@ this can create a lot of work for consumers to adapt to the change. If a change
 is "atomic" (meaning that the old version of something like a method is removed
 at the same time the new version is added), all usages of the old method must be
 converted to the new method in a single change to each consumer's codebase. If
-the usage is wide enough, and especially if usage spans packages by multiple
-owners and spans multiple levels of the dependency tree, the cost of
+the usage is widespread enough, and especially if usage spans packages by
+multiple owners and spans multiple levels of the dependency tree, the cost of
 transitioning may be too high to do at all.
 
 Consequently, libraries should instead make two-phase breaking changes to stable
--- a/library-best-practices/JLBP-7.md
+++ b/library-best-practices/JLBP-7.md
@@ -1,0 +1,37 @@
+[JLBP-7] Make breaking transitions easy
+---------------------------------------
+
+When a library makes an in-place breaking change (meaning that the
+Java package is not renamed - see [JLBP-6](JLBP-6.md) for more
+discussion on renaming rules), this can create a lot of work for
+consumers to adapt to the change. If a change is "atomic" (meaning
+that the old version of something like a method is removed at the same
+time the new version is added), all usages of the old method must be
+converted to the new method in a single change to each consumer's
+codebase. If the usage is wide enough and especially if it spans
+multiple packages, the cost might be too high to adopt at all.
+
+Consequently, libraries should instead make two-phase breaking
+changes:
+
+1. Mark old methods/classes as `@Deprecated` at the same time as
+adding new methods/classes. This is a "stepping stone" release.
+2.  Delete the deprecated methods/classes.
+
+When a breaking change is introduced as two phases, consumers can
+adapt to the change in #1 incrementally. Then when they have completed
+the transition, adapting to #2 is trivial - upgrading to the version
+with the removed methods and classes shouldn't cause any build
+failures. The ideal breaking-change release would have *only* removals
+of deprecated features.
+
+The larger the number of places that use the old version's code, and
+the deeper in the tree that the library is used, the longer the time
+needs to be between #1 and #2. Such a phase can even last years,
+depending on the release cadence of all the libraries using the old
+code.
+
+As an example, a stepping stone release was released for OkHttp 1 to
+help users upgrade to OkHttp 2, in the form of OkHttp 1.6. This
+allowed users to transition large code bases to the OkHttp 2 API (in
+1.6) before flipping from OkHttp 1 to OkHttp 2.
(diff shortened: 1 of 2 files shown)


library-best-practices/JLBP-7.md
```diff
@@ -0,0 +1,37 @@
+[JLBP-7] Make breaking transitions easy
+---------------------------------------
+
+When a library makes an in-place breaking change (meaning that the
```
elharo: Begin by defining a breaking change. E.g. "a breaking change is a chaneg to the public or protected API of an existing class."
Would an out-of-place change ever be breaking?

library-best-practices/JLBP-7.md
```diff
@@ -0,0 +1,37 @@
...
+Java package is not renamed - see [JLBP-6](JLBP-6.md) for more
+discussion on renaming rules), this can create a lot of work for
+consumers to adapt to the change. If a change is "atomic" (meaning
```
elharo: I'm not sure the "atomic" section is important since if it's not atomic, it isn't breaking.

library-best-practices/JLBP-7.md
garrettjonesgoogle: I think in a virtual sense, you can consider a transition from one major version to the next of an artifact that renames the Java package to be a breaking change to users who consider the new major version to be the true successor of the prior version, since they are forced to make code changes to adopt it. If you don't consider a successor with a renamed package to be the "same" library, then you have to consider the prior major version to be either an abandoned library or a deprecated library on long-term support.

library-best-practices/JLBP-7.md
garrettjonesgoogle: In the non-atomic process, the second step (removal of deprecated methods) is breaking though...

garrettjonesgoogle:
@cpovirk could you take a look?

library-best-practices/JLBP-7.md
```diff
@@ -0,0 +1,37 @@
...
+that the old version of something like a method is removed at the same
+time the new version is added), all usages of the old method must be
+converted to the new method in a single change to each consumer's
+codebase. If the usage is wide enough and especially if it spans
+multiple packages, the cost might be too high to adopt at all.
+
+Consequently, libraries should instead make two-phase breaking
```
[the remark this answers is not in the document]
garrettjonesgoogle: Right, I will clarify that.

garrettjonesgoogle:
@elharo I have addressed your feedback

library-best-practices/JLBP-7.md
```diff
@@ -0,0 +1,35 @@
+[JLBP-7] Make breaking transitions easy
+---------------------------------------
+
+When a library makes an in-place breaking change (meaning that the Java package
+is not renamed - see [JLBP-6](JLBP-6.md) for more discussion on renaming rules),
+this can create a lot of work for consumers to adapt to the change. If a change
+is "atomic" (meaning that the old version of something like a method is removed
+at the same time the new version is added), all usages of the old method must be
+converted to the new method in a single change to each consumer's codebase. If
+the usage is wide enough and especially if it spans multiple packages, the cost
```
cpovirk: Is "multiple packages" the most relevant metric? I'd expect the main complications to be things like multiple owners or multiple release cycles.

library-best-practices/JLBP-7.md
```diff
@@ -0,0 +1,35 @@
...
+might be too high to adopt at all.
+
+Consequently, libraries should instead make two-phase breaking changes to stable
+features:
+
+1. Mark old methods/classes as `@Deprecated` at the same time as adding new
+   methods/classes. This is a "stepping stone" release.
+2. Delete the deprecated methods/classes. This wouldn't necessarily be in the
+   next release, particularly if major consumers haven't updated yet.
```
cpovirk: Given a 6-week release cycle, I'd strongly state that it _shouldn't_ be the next release. Fortunately, once you're already doing the change non-atomically, it's usually not too long to postpone the removal longer.

You do say this later on, but it could be nice to make clear from the beginning (assuming that doing so isn't too awkward).

library-best-practices/JLBP-7.md
garrettjonesgoogle: Yeah it's not necessarily simply multiple artifacts, but that it is artifacts across multiple owners and multiple levels of the tree. I will rework this.

library-best-practices/JLBP-7.md
garrettjonesgoogle: I don't fully grok why the 6-week release cycle would influence the decision. It seems like the adoption rate by consumers completely determines the course of action, regardless of the release cycle of the library making the breakage.

cpovirk: A 6-week cycle gives consumer 6 weeks to adopt the new thing. If everyone does release every 6 weeks, that's OK, but that seems optimistic.

Again, though, it seems like you're already saying what I'm advocating for later on, so maybe I'm misunderstanding what you're saying here?

garrettjonesgoogle: I have clarified the conditions for deleting the deprecated stuff. (PTAL)

The reason I want to leave the release cycle out of that is that the release cycle will be suggested by a separate best practice, and I don't want the bast practices unnecessarily coupled. Anyway, the condition of adoption is the key condition that needs to be met, and talking about release cycles is just a proxy anyway.

library-best-practices/JLBP-7.md
```diff
@@ -0,0 +1,37 @@
+[JLBP-7] Make breaking transitions easy
+---------------------------------------
+
+When a library makes an in-place breaking change (meaning that the Java package
+is not renamed - see [JLBP-6](JLBP-6.md) for more discussion on renaming rules),
+this can create a lot of work for consumers to adapt to the change. If a change
+is "atomic" (meaning that the old version of something like a method is removed
+at the same time the new version is added), all usages of the old method must be
+converted to the new method in a single change to each consumer's codebase. If
+the usage is wide enough, and especially if usage spans packages by multiple
```
cpovirk: nit: s/wide/widespread/ ?

PR #77, opened 2018-09-17
Adding JLBP-8: Advance widely used functionality to a stable version

garrettjonesgoogle:


library-best-practices/JLBP-8.md
```diff
@@ -0,0 +1,11 @@
+[JLBP-8] Advance widely used functionality to a stable version
+--------------------------------------------------------------
+
+Libraries at 0.x versions and unstable features marked with
+annotations make no promises about stability, and if they are widely
```
elharo: stability, and if they are --> stability. If they become

library-best-practices/JLBP-8.md
```diff
@@ -0,0 +1,11 @@
...
+used, they can cause a lot of problems when there are surface
+breakages (which is allowed per the rules of unstable functionality,
+explained in [JLBP-3](JLBP-3.md)). If such breakages can cause
+widespread pain, it can be seen as "common-law GA"; the preferable
+path is to release the library as-is, with all its warts, maybe
+marking less-desirable elements as deprecated.
```
elharo: Releasing deprecated as 1.0 smells really bad. I don't think we need to mention the deprecation here.

elharo: What I would like to recommend here is that the question in advance of 1.0 is, are we happy with what we have? Not "What do we still want to add?"

When a library has useful functionality and a stable API, it should be released. Do not wait for every feature to be finished.

garrettjonesgoogle: Sure, releasing deprecated at 1.0 smells bad, but it could be extremely disruptive to rip out something that's deprecated if it is widely used at multiple levels of a dependency tree; the value of having a stable artifact might be higher than the negative of deprecated features in the stable artifact.

I completely agree with the guidance that functionality doesn't have to be finished, it just has to be a frozen surface.

garrettjonesgoogle:
1. For clients not depending on pre-1.0: This is covered by JLBP-4
2. Devrel/docs: this is outside of the scope of the libraries themselves

Finishing things that devs start: Developers could theoretically "finish" functionality, but still leave the version at 0.x (see: google-auth-library-java). The key thrust of this best practice is not about finishing things, but about preferring an earlier transition to 1.0 than a later one, because of the impact to the ecosystem.

library-best-practices/JLBP-8.md
```diff
@@ -0,0 +1,12 @@
+[JLBP-8] Advance widely used functionality to a stable version
+--------------------------------------------------------------
+
+Libraries at 0.x versions and unstable features marked with annotations make no
+promises about stability. If they become widely used, they can cause a lot of
```
elharo: delete "a lot of"

library-best-practices/JLBP-8.md
```diff
@@ -0,0 +1,12 @@
...
+problems when there are surface breakages (which is allowed per the rules of
+unstable functionality, explained in [JLBP-3](JLBP-3.md)). If such breakages can
+cause widespread pain, it can be seen as "common-law GA"; the preferable path is
```
elharo: it can be seen as --> the unstable features can become "common-law GA". The preferable...

library-best-practices/JLBP-8.md
```diff
@@ -0,0 +1,12 @@
...
+to release the library as-is, with all its warts, maybe marking less-desirable
+elements as deprecated. Marking a library or feature as stable is not about
```
elharo: Paragraph break before "Marking"

library-best-practices/JLBP-8.md
```diff
@@ -0,0 +1,12 @@
...
+completeness. Accordingly, it's preferable to mark something complete once the
```
elharo: perhaps delete "it's preferable to"

library-best-practices/JLBP-8.md
```diff
@@ -0,0 +1,14 @@
+[JLBP-8] Advance widely used functionality to a stable version
+--------------------------------------------------------------
+
+Libraries at 0.x versions and unstable features marked with annotations make no
```
cpovirk: "unstable features marked with annotations" might not be clear out of context. Readers might ask "Isn't is true that _all_ unstable features make no promises about stability, not just those with annotations?"

Maybe something like: "Some libraries expose APIs but make no promises of stability. This includes all APIs of libraries at 0.x versions, and it may include specially annotated APIs in an otherwise stable library."

library-best-practices/JLBP-8.md
```diff
@@ -0,0 +1,14 @@
...
+promises about stability. If they become widely used, they can cause problems
+when there are surface breakages (which is allowed per the rules of unstable
+functionality, explained in [JLBP-3](JLBP-3.md)). If such breakages can cause
+widespread pain, the unstable features can become "common-law GA"; the
+preferable path is to release the library as-is, with all its warts, maybe
+marking less-desirable elements as deprecated.
+
+Marking a library or feature as stable is not about completeness. Accordingly,
+mark something complete once the surface is satisfactory, even if the
```
cpovirk: "mark something _stable_," rather than "complete?"

garrettjonesgoogle: oops! will fix.

library-best-practices/JLBP-8.md
garrettjonesgoogle: I will make this statement clearer.

garrettjonesgoogle:
PTAL @cpovirk - I have added recommendations about breaking surface before stability and usage of `@Deprecated`.

You bring up interesting points about how to treat `@Deprecated`. I think it might make sense to make a new JLBP which recommends moving off of deprecated features as quickly as possible, in order to widen the set of versions that can work together in the ecosystem. Such a practice depends on interpreting `@Deprecated` as a message that the user really needs to stop using the functionality.

cpovirk:
> You bring up interesting points about how to treat `@Deprecated`. I think it might make sense to make a new JLBP which recommends moving off of deprecated features as quickly as possible, in order to widen the set of versions that can work together in the ecosystem. Such a practice depends on interpreting `@Deprecated` as a message that the user really needs to stop using the functionality.

+1

PR #84, opened 2018-09-21
Adding JLBP-9: Support the minimum Java version of your consumers

garrettjonesgoogle:
…Java version

### DIFF
--- a/library-best-practices/JLBP-9.md
+++ b/library-best-practices/JLBP-9.md
@@ -5,6 +5,10 @@ If a library B requiring Java 1.7 depends on a library A requiring Java 1.7, and
 library A bumps its minimum version to Java 1.8, then library B cannot pick up
 the new version of library A until it also bumps its own minimum Java
 version. This can result in a situation where the ecosystem is stuck on older
-library versions, preventing the pickup of important things like security
-fixes. Thus, libraries should make sure their consumers advance their minimum
-version of Java before they advance it themselves.
+library versions, preventing the pickup of important things like security fixes,
+and it can also lead to diamond dependency conflicts where no version works for
+all consumers. Thus, libraries should make sure their actively-maintained
+consumers advance their minimum version of Java before they advance it
+themselves. (Waiting for every last consumer could block an upgrade permanently,
+due to cases like abandoned projects.)
+
--- a/library-best-practices/JLBP-9.md
+++ b/library-best-practices/JLBP-9.md
@@ -1,5 +1,5 @@
-[JLBP-9] Ensure consumers always use a higher or equal minimum Java version
----------------------------------------------------------------------------
+[JLBP-9] Support the minimum Java version of your consumers
+-----------------------------------------------------------
 
 If a library B requiring Java 1.7 depends on a library A requiring Java 1.7, and
 library A bumps its minimum version to Java 1.8, then library B cannot pick up
@@ -12,3 +12,9 @@ consumers advance their minimum version of Java before they advance it
 themselves. (Waiting for every last consumer could block an upgrade permanently,
 due to cases like abandoned projects.)
 
+As an alternative, if there is sufficient need for supporting both consumers on
+the old version of Java and consumers wanting features provided by a newer
+version of Java, a long-term support (LTS) release can be created to continue
+supporting the old Java version, and a new major version of the library can be
+released with a higher minimum Java version. The Java package and maven ID
+should be kept the same in the new major version.
--- a/library-best-practices/JLBP-9.md
+++ b/library-best-practices/JLBP-9.md
@@ -1,0 +1,10 @@
+[JLBP-9] Ensure consumers always use a higher or equal minimum Java version
+---------------------------------------------------------------------------
+
+If a library B requiring Java 1.7 depends on a library A requiring Java 1.7, and
+library A bumps its minimum version to Java 1.8, then library B cannot pick up
+the new version of library A until it also bumps its own minimum Java
+version. This can result in a situation where the ecosystem is stuck on older
+library versions, preventing the pickup of important things like security
+fixes. Thus, libraries should make sure their consumers advance their minimum
+version of Java before they advance it themselves.
--- a/library-best-practices/JLBP-9.md
+++ b/library-best-practices/JLBP-9.md
@@ -12,9 +12,9 @@ consumers advance their minimum version of Java before they advance it
 themselves. (Waiting for every last consumer could block an upgrade permanently,
 due to cases like abandoned projects.)
 
-As an alternative, if there is sufficient need for supporting both consumers on
-the old version of Java and consumers wanting features provided by a newer
-version of Java, a long-term support (LTS) release can be created to continue
-supporting the old Java version, and a new major version of the library can be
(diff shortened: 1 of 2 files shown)


library-best-practices/JLBP-9.md
```diff
@@ -0,0 +1,10 @@
+[JLBP-9] Ensure consumers always use a higher or equal minimum Java version
+---------------------------------------------------------------------------
+
+If a library B requiring Java 1.7 depends on a library A requiring Java 1.7, and
+library A bumps its minimum version to Java 1.8, then library B cannot pick up
+the new version of library A until it also bumps its own minimum Java
+version. This can result in a situation where the ecosystem is stuck on older
+library versions, preventing the pickup of important things like security
```
cpovirk: Worth calling out that it also leads to diamond-dependency problems -- ones where there's _no_ version that works for all consumers?

library-best-practices/JLBP-9.md
```diff
@@ -0,0 +1,10 @@
...
+fixes. Thus, libraries should make sure their consumers advance their minimum
+version of Java before they advance it themselves.
```
elharo: This gives dependents veto power over advancement. In a world where many dependents are abandoned, I'm not sure I'm comfortable with that. 

I also think different teams move at different speeds, and I do like the idea of slowing down the advancement of minimum Java versions, so I really don't know what I think about this one.

library-best-practices/JLBP-9.md
garrettjonesgoogle: Good point, I'll add that in.

library-best-practices/JLBP-9.md
garrettjonesgoogle: I would say that you can ignore abandoned dependents for the purposes of this rule; you have to wait for your consumers that "matter." I'll add a clarification.

garrettjonesgoogle:
@elharo I addressed feedback, PTAL

garrettjonesgoogle:
If we want to enable libraries to easily pick up the latest version of all their dependencies, then those dependencies have to account for their consumers' minimum Java versions. Otherwise, we cannot maintain that property. This what got us into a bind for a while for Guava & grpc.

library-best-practices/JLBP-9.md
```diff
@@ -0,0 +1,14 @@
+[JLBP-9] Ensure consumers always use a higher or equal minimum Java version
+---------------------------------------------------------------------------
+
+If a library B requiring Java 1.7 depends on a library A requiring Java 1.7, and
+library A bumps its minimum version to Java 1.8, then library B cannot pick up
+the new version of library A until it also bumps its own minimum Java
+version. This can result in a situation where the ecosystem is stuck on older
+library versions, preventing the pickup of important things like security fixes,
+and it can also lead to diamond dependency conflicts where no version works for
+all consumers. Thus, libraries should make sure their actively-maintained
```
elharo: I don't think I agree with this statement any more. The preferable approach is to branch and maintain at least bug fixes and other mandatory changes for the Java 1.7 version.

garrettjonesgoogle: So branching is a code management strategy that isn't relevant to consumers. We need to talk in terms of artifacts, how those look to consumers, and how the artifacts relate to each other (e.g. naming and potential conflicts). So then, what should library owners do?

1. Create a new major version with a higher minimum Java version, retaining Java package and Maven artifact names. The old major version becomes an LTS release, and gets patches as needed, but no new major features. The dependency tree needs to settle on one of the major versions (since only one can be used at a time), which might just continue to be the old major version.
2. Create a new major version with a higher minimum Java version, renaming the Java package and Maven artifact names. The old major version becomes an LTS release, and gets patches as needed, but no new major features. The dependency tree will probably just keep using the old major version.
3. Deliver two variants of the library indefinitely having different qualifiers, with features stripped out of the one supporting a lower Java version (the Guava approach, with -jre and -android).

elharo: Not quite. What I'm proposing is that the qualifier doesn't change, only the version numbers. For example:

Jaxen 1.x, currently 1.1.6, that is jaxen:jaxen:1.1.6 supports Java 1.4. It maintains compatibility with existing code. If we need a bug fix release, then 1.1.7, 1.1.8, maybe even 1.2.0, and so on. Maybe we never need Jaxen 1.1.7, but if we do the flexibility is there.

At the same time, Jaxen 2.0 moves forward to require Java 5 and later and make other breaking changes. This can be jaxen:jaxen:2.0.0, jaxen:jaxen:2.0.1, jaxen:jaxen:2.1.0, etc. If it makes sense, we change the group ID, artifact ID and/or package. 

Clients can upgrade on their own time, and at their own speed, without blocking the library from advancing to newer versions of Java. 

A well tested stable library should not need a lot of bug fix releases. Jaxen 1.1.6 has lasted for five years now without updates, and I'm only now beginning to think about it again because of issues with Java 9 and later.

garrettjonesgoogle: So it sounds like you're suggesting that library owners choose between (1) or (2) in my list.

garrettjonesgoogle: ping @elharo - I wanted to come to some kind of conclusion on this topic

elharo: I like #1 and I think we should recommend that unconditionally. If the only breaking change is the minimum Java version, then there's no reason to change Maven IDs or package names.

That should happen only for major API changes.

library-best-practices/JLBP-9.md
```diff
@@ -0,0 +1,14 @@
+[JLBP-9] Ensure consumers always use a higher or equal minimum Java version
```
elharo: I suggest rewording as, "Support your consumers' minimum Java version."

I still don't like the term consumer here. Maybe "clients' minimum Java version" if you don't like dependents? To me the word "consumer" implies that the library is being used up somehow, and that's not the case.

garrettjonesgoogle:
I have attempted to work in the feedback, PTAL

library-best-practices/JLBP-9.md
```diff
@@ -12,3 +12,9 @@ consumers advance their minimum version of Java before they advance it
 themselves. (Waiting for every last consumer could block an upgrade permanently,
 due to cases like abandoned projects.)
 
+As an alternative, if there is sufficient need for supporting both consumers on
```
elharo: "there is sufficient need for supporting" --> "you need to support"

library-best-practices/JLBP-9.md
```diff
@@ -12,3 +12,9 @@ consumers advance their minimum version of Java before they advance it
...
+the old version of Java and consumers wanting features provided by a newer
+version of Java, a long-term support (LTS) release can be created to continue
```
elharo: not sure if we want to use the phrase "long-term support" here. The main point is that you can release new minor and micro versions while also releasing new major versions, but we can futz with the wording later. The intent is fine.

garrettjonesgoogle: Ok, I'll keep it for now because I don't have an alternative yet.

PR #86, opened 2018-09-21
Minor cleanup

garrettjonesgoogle:
Just some cleanup I thought I would do while trolling about the code base.

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -16,10 +16,6 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
-import com.google.cloud.tools.opensource.dependencies.Artifacts;
-import com.google.cloud.tools.opensource.dependencies.DependencyGraph;
-import com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder;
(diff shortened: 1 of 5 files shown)


dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -16,7 +16,10 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
-import freemarker.core.ParseException;
+import com.google.cloud.tools.opensource.dependencies.Artifacts;
```
elharo: This looks like internal import order. I don't think our externally published Eclipse stylesheet matches this yet. If so, I'd prefer not to have to manually reorder our imports.

garrettjonesgoogle: I used the google-java-format plugin in IntelliJ to produce this order (so it's easily reproducible exclusively in OSS). I can revert this reordering if it disrupts your workflow though.

elharo: We may want to take an action item or even an OKR to update the corresponding Eclipse plugin, but until that's done please revert.

garrettjonesgoogle:
Feedback addressed, PTAL @elharo

PR #88, opened 2018-09-21
Adding JLBP-10: Maintain API stability as long as needed for consumers

garrettjonesgoogle:
cc @ejona86


library-best-practices/JLBP-10.md
```diff
@@ -0,0 +1,33 @@
+[JLBP-10] Infrequent breaking changes
+-------------------------------------
+
+Every breaking change incurs a cost. There is the immediate cost of the
```
elharo: to the consumers

library-best-practices/JLBP-10.md
```diff
@@ -0,0 +1,33 @@
...
+consumers who are forced to change their own code to adapt to the newer version,
+but the far higher cost is the inconsistency introduced into the ecosystem and
+the resulting potential for diamond dependency conflicts. The deeper a library
+is in the dependency tree, the higher the inconsistency cost, regardless of the
+magnitude of the direct change cost.
+
+Consequently, a library should try to make breaking changes very infrequently if
+ever. It's hard to recommend a frequency for all cases because of the widely
+varying cost. On one end of the spectrum is pre-1.0 libraries which don't
```
elharo: are pre-1.0

library-best-practices/JLBP-10.md
```diff
@@ -0,0 +1,33 @@
...
+promise stability, and which can have breaking changes between every release. On
+the other end is Java standard libraries which all libraries depend on, most of
```
elharo: are Java standard

library-best-practices/JLBP-10.md
```diff
@@ -0,0 +1,33 @@
...
+whose classes have not had any breaking changes since they were introduced. As
+an anchoring point, no more than once every five years might be a good default
+for any library used by other open source libraries.
+
+One thing to keep in mind is that many consumers might be highly resistant to
```
elharo: might be --> are

library-best-practices/JLBP-10.md
```diff
@@ -0,0 +1,33 @@
...
+accepting breaking changes for various reasons:
+
+- They might have low test coverage, making it costly to verify correctnes
```
elharo: correctness.

library-best-practices/JLBP-10.md
```diff
@@ -0,0 +1,33 @@
...
+- They might not have enough people to do dependency upgrade work
```
elharo: period
people --> developers and time

library-best-practices/JLBP-10.md
```diff
@@ -0,0 +1,33 @@
...
+- They might have a high risk aversion
```
elharo: period

patflynn:
I remember a previous doc proposing to always change the package name when changing the semantic version (i.e. breaking).  Are we still proposing that?

garrettjonesgoogle:
@patflynn Here is the current best practice on the topic: https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/library-best-practices/JLBP-6.md

The conclusion was that we couldn't make a global recommendation for renaming vs not, because there are difficult tradeoffs.

patflynn:
I worry that JLBP-6 in its current form doesn't provide much value since
it's hard to parse any strong recommendations from it.  My guess is that it
will be ignored.

My personal preference is for requiring the major version to be included in
package namespace, similarly to Go modules, and one platform APIs.

I don't believe that package renaming on the client side is a significant
enough cost to users that it outweighs the benefits of avoiding dependency
conflicts.

garrettjonesgoogle:
JLBP-6 does provide value in that it should stop people from releasing packages like guava-jdk5.

I don't think our discussion of JLBP-6 affects the present best practice - maybe we should discuss it separately? I'd recommend reading over the PR conversations at https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/63 - we had quite a bit of back and forth on the topic.

garrettjonesgoogle:
@elharo I have addressed your feedback, PTAL

library-best-practices/JLBP-10.md
```diff
@@ -0,0 +1,33 @@
+[JLBP-10] Infrequent breaking changes
+-------------------------------------
+
+Every breaking change incurs a cost to the consumers. There is the immediate
```
elharo: I'm not sure the cost is so high if older versions move forward in parallel branches. I think we need to rethink this JLBP while being explicit that there can be more than one branch.

library-best-practices/JLBP-10.md
```diff
@@ -0,0 +1,33 @@
...
+deeper a library is in the dependency tree, the higher the inconsistency cost,
+regardless of the magnitude of the direct change cost.
+
+Consequently, a library should try to make breaking changes very infrequently if
+ever. It's hard to recommend a frequency for all cases because of the widely
+varying cost. On one end of the spectrum are pre-1.0 libraries which don't
+promise stability, and which can have breaking changes between every release. On
+the other end are Java standard libraries which all libraries depend on, most of
+whose classes have not had any breaking changes since they were introduced. As
+an anchoring point, no more than once every five years might be a good default
```
elharo: I don't think a specific duration matters a great deal. What's more important is that the current version without breaking changes is supported for 5+ years or until no one's using it. A new branch can move forward at any time.

library-best-practices/JLBP-10.md
garrettjonesgoogle: In that case, the cost is high for the maintainer, because now there are multiple branches (and releases) to maintain and keep track of. The cost is high for someone.

library-best-practices/JLBP-10.md
garrettjonesgoogle: It doesn't feel like you're necessarily disagreeing with the text; the text suggests five years "as an anchoring point," but still leaves the exact time frame up to the library maintainer.

library-best-practices/JLBP-10.md
elharo: Yes, the cost is higher for the maintainer; but is that a deal breaker? This documents best practices and I'm not surprised that the best practice here costs more. Quality and reliability aren't cheap. A maintainer may reasonably choose to prioritize cost for any given library, but I don't think we need to allow for that as a best practice.

garrettjonesgoogle: It's not merely cost; the cost is more complexity. I don't think we want to be in the business of pushing people to adopt more complex solutions if don't really, really have to.

library-best-practices/JLBP-10.md
```diff
@@ -0,0 +1,33 @@
+[JLBP-10] Infrequent breaking changes
```
elharo: Do we want to phrase these as full, imperative, positive statements? If so, something like, "Maintain API stability for multiple years."

If the older version is maintained and available, is there even such a thing as a breaking change? If clients can migrate at their own pace and speed, this is not a big issue. It's more like, "here's this library that still works and here's a different library with extra features  and a new package and artifact ID you can use if you prefer."

library-best-practices/JLBP-10.md
elharo: I think the benefits justify the cost, at least across the sum of consumers and suppliers. It is more complex than single repo, single branch, but I think it's a necessary one for libraries to not get locked into old APIs indefinitely.

library-best-practices/JLBP-10.md
garrettjonesgoogle: I like the wording of "maintain stability", but the "years" part doesn't seem to apply to all cases.

If the older version stays maintained, then you are correct, there hasn't really been a breaking change. *However*, if the older version becomes an LTS release that has a deprecation timeline, consumers will be forced to accept a breakage eventually.

elharo: "Maintain API stability as long as needed for clients"?

It occurs to me that we don't actually care about clients that are not themselves in active development, since they'll just use the old artifacts no matter what we do. The real issue is indeed Java versions since libraries do choose to continue with older versions.

garrettjonesgoogle: That sounds like good wording, I'll use that.

garrettjonesgoogle:
Addressed feedback, PTAL

library-best-practices/JLBP-10.md
```diff
@@ -0,0 +1,45 @@
...
+One thing to keep in mind is that many consumers are highly resistant to
+accepting breaking changes for various reasons:
+
+- They might have low test coverage, making it costly to verify correctness.
+- They might not have enough developers and time to do dependency upgrade work.
+- They might have a high risk aversion.
+- They might have put your major version on their own surface and they don't
+  want to force a breaking change on their own users for a long period.
+
+As soon as you introduce breaking changes, they might decide to just stick with
```
elharo: delete "just"

library-best-practices/JLBP-10.md
```diff
@@ -0,0 +1,45 @@
...
+the last working version indefinitely. As an example of a popular library that
+took a long time to upgrade (no judgment on them!), Hadoop introduced Guava
```
elharo: "(no judgment on them!)" sounds more judgey than if you didn't say the parenthetical at all.

library-best-practices/JLBP-10.md
garrettjonesgoogle: will do.

library-best-practices/JLBP-10.md
garrettjonesgoogle: Heh ok, I'll remove it.

PR #90, opened 2018-09-22
Introduce Update class

elharo:
@garrettjonesgoogle This refactoring is a step on the way to more configurable reports including upper bounds.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Update.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Update.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2018 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.opensource.dependencies;
+
+import java.util.Objects;
+
+import org.eclipse.aether.artifact.Artifact;
+
+/**
+ * Holder for a suggested dependency update. 
+ */
+public class Update {
+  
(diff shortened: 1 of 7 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Update.java
```diff
@@ -0,0 +1,91 @@
...
+package com.google.cloud.tools.opensource.dependencies;
+
+import java.util.Objects;
+
+import org.eclipse.aether.artifact.Artifact;
+
+/**
+ * Holder for a suggested dependency update. 
+ */
+public class Update {
```
garrettjonesgoogle: Have you considered using [AutoValue](https://github.com/google/auto/tree/master/value)? It reduces a lot of boilerplate.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/ArtifactsTest.java
```diff
@@ -34,5 +34,5 @@ public void testToCoordinates_withPackaging() {
     Artifact artifact = new DefaultArtifact("com.google.example", "example", "jar", "1.3.2");
     Assert.assertEquals("com.google.example:example:1.3.2", Artifacts.toCoordinates(artifact));
   }
-
+
```
garrettjonesgoogle: Undo whitespace regression?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Update.java
elharo: Yes, and rejected it for multiple reasons including excess dependencies and idiosyncratic code.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/ArtifactsTest.java
elharo: https://groups.google.com/a/google.com/forum/#!msg/java-users/_ABiES5o020/J3TVUclzN2gJ

elharo: Official Google Java code review policy is to never comment on trailing white space. Critique doesn't even show it. Unfortunately Github does, but we should still not worry about it.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Update.java
elharo: Also we need a specific toString method.

PR #96, opened 2018-09-25
assume there can be more than one version

elharo:
@garrettjonesgoogle also bring into sync with SemVer spec terminology. I removed a warning about Java version bumps that:

A) is not about semantic versioning
B) Implicitly assumes only one version can be maintained.

### DIFF
--- a/library-best-practices/JLBP-4.md
+++ b/library-best-practices/JLBP-4.md
@@ -5,10 +5,10 @@
   public API within the same major version. For libraries following semantic
   versioning, this means libraries with a 0.x.y version. (See [JLBP-3](JLBP-3.md) 
   for more details on the recommendations for semantic versioning.)
-- Unstable features are not part of the public API of a
-  stable library. They are often marked with annotations like `@Beta`. (See
-  [JLBP-3](JLBP-3.md) for more details on the recommendations for annotating
-  unstable features.)
+- Features that are not part of the public API of a
+  stable library are called *unstable*. Unstable features in a stable library 
+  should be marked with annotations such as `@Beta`. (See
+  [JLBP-3](JLBP-3.md) for more details on annotating unstable features.)
 - If your library depends on an unstable library or feature, and that feature
   experiences a breaking change between versions, your library will be locked to
   a specific version of that dependency.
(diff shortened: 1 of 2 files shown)


library-best-practices/JLBP-4.md
```diff
@@ -1,17 +1,17 @@
...
-  Public API within the same major version. For libraries following semver, this
-  means libraries with a 0.x.y version. (See [JLBP-3](JLBP-3.md) for more
-  details on the recommendations for following semver.)
-- Unstable features are features that are not part of the stable Public API of a
-  stable library, often marked with annotations like `@Beta`.  (See
+- Unstable libraries are libraries that allow breaking changes to their
+  public API within the same major version. For libraries following semantic
+  versioning, this means libraries with a 0.x.y version. (See [JLBP-3](JLBP-3.md) 
+  for more details on the recommendations for semantic versioning.)
+- Unstable features are not part of the public API of a
```
garrettjonesgoogle: Removing "are features that" changes the meaning here. 

- Prior wording: the "given" is features that aren't part of the stable public API; this definition gives those a label.
- New wording: the "given" is unstable features, and allows for other features that are not part of the public API but aren't called "unstable."

I think this rewording makes it less clear.

elharo: Reworded. let me know what you think now.

library-best-practices/JLBP-4.md
```diff
@@ -5,10 +5,10 @@
   public API within the same major version. For libraries following semantic
   versioning, this means libraries with a 0.x.y version. (See [JLBP-3](JLBP-3.md) 
   for more details on the recommendations for semantic versioning.)
-- Unstable features are not part of the public API of a
-  stable library. They are often marked with annotations like `@Beta`. (See
-  [JLBP-3](JLBP-3.md) for more details on the recommendations for annotating
-  unstable features.)
+- Features that are not part of the public API of a
+  stable library are called *unstable*. Unstable features in a stable library 
+  should be marked with annotations such as `@Beta`. (See
```
garrettjonesgoogle: Maybe we should make this more direct: "To mark a feature as unstable and not part of the public API, use annotations such as `@Beta`."

elharo: Good idea. Done.

PR #100, opened 2018-09-28
Put test results (PASS/FAIL) on front page

elharo:
@garrettjonesgoogle We can prettify this later, but this brings us into sync with the dashboard plans.

### DIFF
--- a/dashboard/src/main/resources/templates/dashboard.ftl
+++ b/dashboard/src/main/resources/templates/dashboard.ftl
@@ -1,6 +1,16 @@
 <html lang="en-US">
   <head>
     <title>Google Cloud Platform Code Health Open Source Dashboard</title>
+    <style>
+    .PASS {
+      background-color: green;
+      font-weight: bold;
+    }
+    .FAIL {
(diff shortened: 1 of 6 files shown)


dashboard/src/main/resources/templates/dashboard.ftl
```diff
@@ -1,15 +1,32 @@
...
-        <li><a href='${artifact?replace(":", "_")}.html'>${artifact}</a></li>
+      <table>
+      <tr>
+        <th>Artifact</th><th>Upper Bounds</th><th>Dependency Convergence</th>
+      </tr>
+      <#list table as row>
+        <tr>
+          <td><a href='${row[0]?replace(":", "_")}.html'>${row[0]}</a></td>
+          <td class='${row[1]}'>${row[1]}</td>
+          <td class='${row[2]}'>${row[2]}</td>
```
garrettjonesgoogle: Using an array for the row makes it harder to reason about the code being plumbed around. Would you consider a POJO instead?

PR #101, opened 2018-09-28
Jlbp1

elharo:
@garrettjonesgoogle I forgot to switch to master before branching, but that will take care of itself when #100 is in. Meanwhile look here at library-best-practices/JLBP-1.md

### DIFF
--- a/library-best-practices/JLBP-1.md
+++ b/library-best-practices/JLBP-1.md
@@ -5,7 +5,7 @@ Use the minimum number of dependencies that is reasonable.
 Adding a dependency for a large amount of functionality may be ok,
 but avoid pulling in dependencies just to save a few lines of code,
 because every dependency of a library is a liability of both
-that library and that library's dependents.
+that library and that library's consumers.
 
 Some specific notes about minimizing dependencies:
 
--- a/library-best-practices/JLBP-1.md
+++ b/library-best-practices/JLBP-1.md
@@ -29,11 +29,12 @@ Some specific notes about minimizing dependencies:
 
 - For any given functionality, pick exactly one library. For example,
   GSON, Jackson, and javax.json all parse JSON files. If one is already
-  pulled in by another dependency, use that. Otherwise choose the one
-  you prefer. Do not include more than one in your dependency tree.
+  pulled in by another dependency, use that. Otherwise choose one
+  and standardize on it. Do not include more than one in your dependency tree.
+  Do not allow different team members to choose different libraries.
 
 - If you can reasonably reimplement functionality instead of adding
   another dependency, do so. For example, if the only classes you're 
   using from Guava are `Preconditions` and `Strings`, it's not 
-  worth it. You can easily reimplement any methods in those classes
-  you're using.  
+  worth adding a dependency on Guava. You can easily reimplement 
+  any methods in those classes you're using.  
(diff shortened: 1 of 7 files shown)


library-best-practices/JLBP-1.md
```diff
@@ -13,8 +14,26 @@ Some specific notes about minimizing dependencies:
...
+
 - Scrutinize all dependency additions. Check the result of
   `mvn dependency:tree` (after running `mvn install -DskipTests`
   to build the library) to see which transitive dependencies are
   added by just adding a single dependency to your own library,
   and if you require all of the transitive dependencies.
+
+- Prefer JDK classes where available. For example, XOM and JDOM
+  are very convenient and far easier to use than DOM. However, most 
+  uses of these library can be satisfied with the `org.w3c.dom`
```
garrettjonesgoogle: grammar (uses of these library)

library-best-practices/JLBP-1.md
```diff
@@ -13,8 +14,26 @@ Some specific notes about minimizing dependencies:
...
+
+- For any given functionality, pick exactly one library. For example,
+  GSON, Jackson, and javax.json all parse JSON files. If one is already
+  pulled in by another dependency, use that. Otherwise choose the one
+  you prefer. Do not include more than one in your dependency tree.
+
+- If you can reasonably reimplement functionality instead of adding
+  another dependency, do so. For example, if the only classes you're 
+  using from Guava are `Preconditions` and `Strings`, it's not 
+  worth it. You can easily reimplement any methods in those classes
```
garrettjonesgoogle: -> it's not worth adding Guava as a dependency

elharo: done

library-best-practices/JLBP-1.md
elharo: done

library-best-practices/JLBP-1.md
elharo: done

library-best-practices/JLBP-1.md
```diff
@@ -4,7 +4,8 @@
 Use the minimum number of dependencies that is reasonable.
 Adding a dependency for a large amount of functionality may be ok,
 but avoid pulling in dependencies just to save a few lines of code,
-because every dependency is a liability.
+because every dependency is a liability inherited by your own
+dependents.
```
garrettjonesgoogle: (sorry I didn't notice on the first read-through)

It seemed like you were nudging me away from using any second-person pronouns :-P

Anyway it feels like a switch of target in the middle of this sentence. How about:

"because every dependency of a library is a liability of both that library and that library's consumers."

library-best-practices/JLBP-1.md
garrettjonesgoogle: still "these library"

library-best-practices/JLBP-1.md
elharo: How about consumers --> dependents?

In general second person is fine if the reader/you is really doing something. E.g. "When you type ls -a in the console..." However, in tech writing "you" sometimes becomes a general pronoun that can mean the reader/the compiler/the VM/the class/etc. depending on context. E.g. "If you depend on Guava then watch out for this problem." The reader doesn't depend on Guava. The reader's project does. 

Third person is often more specific helps the reader see what's really happening.

library-best-practices/JLBP-1.md
elharo: Huh, thought I fixed that.

PR #107, opened 2018-10-02
Resolution exception to be shown in dashboard rather than stderr

suztomo:
Now ResolutionException is shown in dashboard rather than stderr

Dummy dashboard output generated by a test case:

invalid_artifact_in_dashboard

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/ArtifactResults.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/ArtifactResults.java
@@ -61,6 +61,7 @@ public final class ArtifactResults {
   /**
    *  @return message of exception occurred when running test, null for no exception
    */
+  @Nullable
   public String getExceptionMessage() {
     return exceptionMessage;
   }
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/ArtifactResults.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/ArtifactResults.java
@@ -32,9 +32,14 @@ public final class ArtifactResults {
 
   private final Map<String, Boolean> results = new HashMap<>();
   private final Artifact artifact;
-  
+  private String exceptionMessage;
+
   public ArtifactResults(Artifact artifact) {
     this.artifact = artifact;
+  }
+
+  public void setExceptionMessage(String exceptionMessage) {
+    this.exceptionMessage = exceptionMessage;
   }
 
   void addResult(String testName, boolean result) {
@@ -53,4 +58,10 @@ public final class ArtifactResults {
     return Artifacts.toCoordinates(artifact);
   }
 
+  /**
+   *  @return message of exception occurred when running test, null for no exception
+   */
+  public String getExceptionMessage() {
+    return exceptionMessage;
+  }
 }
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -16,17 +16,6 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
-import com.google.cloud.tools.opensource.dependencies.Artifacts;
-import com.google.cloud.tools.opensource.dependencies.DependencyGraph;
-import com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder;
-import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
-import com.google.cloud.tools.opensource.dependencies.Update;
-import com.google.cloud.tools.opensource.dependencies.VersionComparator;
-import com.google.common.annotations.VisibleForTesting;
-import freemarker.template.Configuration;
-import freemarker.template.Template;
-import freemarker.template.TemplateException;
-import freemarker.template.Version;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -42,12 +31,25 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+
+import freemarker.template.Configuration;
+import freemarker.template.Template;
+import freemarker.template.TemplateException;
+import freemarker.template.Version;
 import org.eclipse.aether.RepositoryException;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.artifact.DefaultArtifact;
 import org.eclipse.aether.collection.DependencyCollectionException;
 import org.eclipse.aether.resolution.ArtifactDescriptorException;
 import org.eclipse.aether.resolution.DependencyResolutionException;
+
+import com.google.cloud.tools.opensource.dependencies.Artifacts;
+import com.google.cloud.tools.opensource.dependencies.DependencyGraph;
+import com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder;
+import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
+import com.google.cloud.tools.opensource.dependencies.Update;
+import com.google.cloud.tools.opensource.dependencies.VersionComparator;
+import com.google.common.annotations.VisibleForTesting;
 
 public class DashboardMain {
   public static final String TEST_NAME_UPPER_BOUND = "Upper Bounds";
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -16,6 +16,13 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
+import com.google.cloud.tools.opensource.dependencies.Artifacts;
+import com.google.cloud.tools.opensource.dependencies.DependencyGraph;
+import com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder;
+import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
+import com.google.cloud.tools.opensource.dependencies.Update;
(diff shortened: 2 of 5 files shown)


dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -47,54 +51,57 @@
...
+
   private static Path outputDirectory;
   private Builder builder = new Builder();
-  
+
   @BeforeClass
-  public static void setUp() throws ArtifactDescriptorException, IOException, TemplateException {
-    outputDirectory = DashboardMain.generate();
+  public static void setUp() throws IOException {
+    outputDirectory = Files.createDirectories(Paths.get("target", "dashboard"));
```
suztomo: `dashboardMain.generate()` at `setUp` was assuming `testDashboard()` is called immediately after it. However it didn't work when I added another test `testDashboardForRepositoryException()`. So I moved `dashboardMain.generate()` into `testDashboard()`.

suztomo:
@elharo As per discussion, I implemented enhancement to dashboard to show "UNAVAILABLE" when there's problem in generating artifact result. Would you review the change?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/ArtifactResults.java
```diff
@@ -53,4 +58,10 @@ public String getCoordinates() {
     return Artifacts.toCoordinates(artifact);
   }
 
+  /**
+   *  @return message of exception occurred when running test, null for no exception
+   */
+  public String getExceptionMessage() {
```
elharo: @Nullable

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -35,23 +42,18 @@
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-
+import org.eclipse.aether.RepositoryException;
```
elharo: we need to standardize the import order; there's a bit of a discrepancy between the external and internal stylesheets

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -60,71 +62,71 @@ public static void main(String[] args)
...
       try {
         ArtifactResults results = generateReport(configuration, output, artifact);
         table.add(results);
-      } catch (DependencyCollectionException | DependencyResolutionException | IOException
-          | TemplateException ex) {
-        // TODO the dashboard should somehow show that it failed to generate this report;
-        // not just silently drop it from the index
-        System.err.println("Error generating report for " + artifact);
-        System.err.println(ex.getMessage());
+      } catch (RepositoryException | IOException | TemplateException ex) {
```
elharo: TemplaetException needs to be separated into its own catch block with different behavior

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -151,15 +152,15 @@ private static ArtifactResults generateReport(Configuration configuration, Path
       DependencyGraph transitiveDependencies) {
     Map<String, String> expectedVersionMap = graph.getHighestVersionMap();
     Map<String, String> actualVersionMap = transitiveDependencies.getHighestVersionMap();
-    
+
```
elharo: Please set your IDE not to muck with existing trailing white space. Internal tools work a bit better than github here, but official policy is not to change this and not to comment on it.

dashboard/src/main/resources/templates/dashboard.ftl
```diff
@@ -21,9 +25,20 @@
       </tr>
       <#list table as row>
         <tr>
-          <td><a href='${row.getCoordinates()?replace(":", "_")}.html'>${row.getCoordinates()}</a></td>   
-          <td class='${row.getResult("Upper Bounds")?string('PASS', 'FAIL')}'>${row.getResult("Upper Bounds")?string('PASS', 'FAIL')}</td>
-          <td class='${row.getResult("Dependency Convergence")?string('PASS', 'FAIL')}'>${row.getResult("Dependency Convergence")?string('PASS', 'FAIL')}</td>
+          <#if row.getResult("Upper Bounds")?? >
```
elharo: what does ?? do?

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
elharo: You might need to move the test to another class because the setup here is `@BeforeClass` and not `@Before` (It's expensive and read-only.)

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -47,54 +51,57 @@
...
     DashboardMain.main(null);
   }
-  
+
   @Test
   public void testDashboard()
       throws IOException, TemplateException, ParsingException, ArtifactDescriptorException {
-    
+    // Creates "dashboard.html" in outputDirectory
+    outputDirectory = DashboardMain.generate();
```
elharo: I think you've introduced order dependence into the tests. This isn't guaranteed to work. I suggest leaving this test alone and moving your test to a new class.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: As talked:

- IOException -> show in dashboard
- TemplateException -> RuntimeException

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
suztomo: Moved the new tests to DashboardUnavailableArtifactTest

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Got it.

dashboard/src/main/resources/templates/dashboard.ftl
suztomo: Added Freemarker comment on the logic.

dashboard/src/main/resources/templates/dashboard.ftl
```diff
@@ -21,9 +25,22 @@
...
-          <td class='${row.getResult("Dependency Convergence")?string('PASS', 'FAIL')}'>${row.getResult("Dependency Convergence")?string('PASS', 'FAIL')}</td>
+          <#if row.getResult("Upper Bounds")?? ><#-- checking isNotNull() -->
+            <#-- When it's not null, it means the test ran. It's either PASS or FAIL -->
+            <#assign upper_bound_test_label = row.getResult("Upper Bounds")?then('PASS', 'FAIL') >
+          <#else>
+            <#-- Null means there's an exception and test couldn't run -->
+            <#assign upper_bound_test_label = "UNAVAILABLE">
+          </#if>
+          <td><a href='${row.getCoordinates()?replace(":", "_")}.html'>${row.getCoordinates()}</a></td>
+          <td class='${upper_bound_test_label}' title="${row.getExceptionMessage()!""}">${upper_bound_test_label}</td>
```
elharo: double double quote at end of ${row.getExceptionMessage()!""}" looks suspicious

dashboard/src/main/resources/templates/dashboard.ftl
```diff
@@ -21,9 +25,22 @@
...
+
+          <#if row.getResult("Dependency Convergence")?? >
+            <#assign dependency_convergence_test_label = row.getResult("Dependency Convergence")?then('PASS', 'FAIL') >
+          <#else>
+            <#assign dependency_convergence_test_label = "UNAVAILABLE">
+          </#if>
+          <td class='${dependency_convergence_test_label}' title="${row.getExceptionMessage()!""}">${dependency_convergence_test_label}</td>
```
elharo: double double quote again

dashboard/src/main/resources/templates/dashboard.ftl
suztomo: Actually this doesn’t cause problem. Freemarker processes ${}, which contain only one pair of double quotes.

If you feel problem in readability, I’m happy to update this piece.

suztomo: Well, as you had to add a comment, readability is not good. I’ll use a pair of single quotes to easily distinguish them.

elharo: Is this some weirdness with the exclamation point? E.g. if it's null then return the empty string instead or something like that?

suztomo: Exactly. If it's null, then empty string.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardUnavailableArtifactTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+    Assert.assertEquals(
+        "The size of rows in table should match the number of artifacts + 1 (header)",
+        tr.size(),table.size() + 1);
+
+    Nodes tdForValidArtifact = tr.get(1).query("td");
+    Assert.assertEquals(
+        Artifacts.toCoordinates(validArtifact), tdForValidArtifact.get(0).getValue());
+    Element firstResult = (Element) (tdForValidArtifact.get(1));
+    Truth.assertThat(firstResult.getValue()).isEqualTo("PASS");
```
elharo: Personal preference but I find assertEquals a little more obvious.

PR #112, opened 2018-10-03
Dependency section in component.ftl

### LINKED ISSUE #111: Proposal: Dependency section in component.ftl
This proposal is to have dependency tree shown in component page. This will help to understand problem of version mismatch.

As the `DependencyTreeFormatter.formatDependencyPaths(dependencyPathList)` is already there. Integrating it to component.ftl is simple and won't cause additional network communication.

Example implementation:

screen shot 2018-10-03 at 09 58 34

@elharo What do you think?

suztomo:
Fixes #111

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -46,6 +46,7 @@ import org.eclipse.aether.resolution.DependencyResolutionException;
 import com.google.cloud.tools.opensource.dependencies.Artifacts;
 import com.google.cloud.tools.opensource.dependencies.DependencyGraph;
 import com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder;
+import com.google.cloud.tools.opensource.dependencies.DependencyTreeFormatter;
 import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
 import com.google.cloud.tools.opensource.dependencies.Update;
 import com.google.cloud.tools.opensource.dependencies.VersionComparator;
@@ -132,6 +133,7 @@ public class DashboardMain {
       Map<Artifact, Artifact> upperBoundFailures =
           findUpperBoundsFailures(completeDependencies, transitiveDependencies);
 
+      String dependencyTree = DependencyTreeFormatter.formatDependencyPaths(completeDependencies.list());
       Template report = configuration.getTemplate("/templates/component.ftl");
 
       Map<String, Object> templateData = new HashMap<>();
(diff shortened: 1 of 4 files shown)


dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -16,22 +16,25 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
+import com.google.cloud.tools.opensource.dependencies.Artifacts;
```
suztomo: reverting imports

suztomo:
@elharo Would you review the change for #111 "dependency tree shown in component" ? (the issue has the screenshot)

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -132,6 +133,7 @@ private static ArtifactResults generateReport(Configuration configuration, Path
       Map<Artifact, Artifact> upperBoundFailures =
           findUpperBoundsFailures(completeDependencies, transitiveDependencies);
 
+      String dependencyTree = DependencyTreeFormatter.formatDependencyPaths(completeDependencies.list());
```
elharo: Not in this PR, but I suspect we're going to want this to return a more complex data structure in the future that can be more neatly formatted in HTML

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -136,7 +136,8 @@ public void testComponent_success() throws IOException, ValidityException, Parsi
       Nodes greens = document.query("//h3[@style='color: green']");
       Assert.assertTrue(greens.size() >= 2);
       Nodes pres = document.query("//pre");
-      Assert.assertEquals(0, pres.size());
+      // There's a pre tag for dependency
+      Assert.assertEquals(1, pres.size());
```
elharo: This could fail if there's another pre element for the wrong reason. I wasn't planning on one, but now there is. :-) Maybe add an id attribute and check for that to distinguish? Or check the contents of the element?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: I agree. I think ul/li tags with collapse/expand would be nice. e.g., `formatDependencyPathsHtml`

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
suztomo: Enhanced test cases to distinguish pre tags.

Issue #110, opened 2018-10-03
Static Linkage Checker


elharo:
We need a tool that can verify that a single artifact is linkage compatible. This should verify that all static references in a given classpath are valid. This is a necessary tool to verify that we're actually meeting the objective of not surfacing conflicts to users. (Verifying version agreement is not sufficient.)

The tool takes the Maven coordinates?/pom.xml file? as input. It resolves the dependency tree according to the usual Maven dependency mediation algorithm. Thus all artifacts are assigned specific versions. This is the runtime classpath. Then it checks this for linkage conflicts. 

Linkage conflict: The signature of a non-private method or class in a dependency has changed in an incompatible way between the version supplied at compile time and the version invoked at runtime. For example, a public method may be removed from a class or a class may be made final. Linkage conflicts detected at runtime manifest as ReflectiveOperationException, NoClassDefFoundError, NoSuchFieldException, MethodNotFoundException, LinkageError, or other related exceptions.
Or, another perspective: In cases where binary compatibility and source compatibility are the same, a linkage conflict is when compilation would fail if the libraries in the classpath were all built together from their originating source code, or when reflection would fail.
Opposite: Linkage-compatible.
Sub-type: Static linkage conflict: A linkage conflict caused by a direct code reference (non-reflective).
Sub-type: Reflective linkage conflict: A linkage conflict caused by a reflective reference.

It runs through the classpath, finds every statically referenced class, interface, method, and field and verifies the signatures match up. E.g. no signature has changed or been removed in an incompatible way in the specific version found in the classpath.

We are not considering access via reflection.

Implementation: maven plugin? main() method? dashboard report? all of this?

Start with a one page design doc?

How does this differ from simple compilation checking?
Are these errors common in practice?

@garrettjonesgoogle Can you flesh this issue out a bit?

patflynn:
question: I feel like I've often seen these errors occur when multiple versions of the same library are packaged into the runtime classpath rather than one selected using some build system algorithm.  Is that impression wrong or have build systems changed to not do that?

elharo:
If Maven builds the classpath that won't happen if the group ID and artifact ID are the same. It can happen if the group IDs and artifact IDs are different though. E.g. Servlet 2.5 and servlet 3.0. It perhaps most commonly happens when someone forks and rereleases a library without shading or repackaging it. 

If the classpath is not set by maven but by something else, then it can certainly pick up multiple jars though only one applies.

[...]

PR #114, opened 2018-10-05
Adding test cases to explore applicability of BCEL to #110 Static Lin…

### LINKED ISSUE #113: Test cases with BCEL for Static Linkage Checker
As preparation for #110 . As discussed with Elliottte, first let's write test cases where we can get enough information from class files, using [BCEL JavaClass](https://commons.apache.org/proper/commons-bcel/manual/bcel-api.html).

- Get list of method signatures from a class file (target)
- Get constant pool information (source)

suztomo:
Adding test cases to explore applicability of BCEL to Static Linkage Checker

Fixes #113 Test cases with BCEL for Static Linkage Checker


suztomo:
@garrettjonesgoogle Would you review the pull request?

As per discussion with @elharo , we start adding test cases to read a class file using BCEL.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
```diff
@@ -0,0 +1,117 @@
...
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ClassDumperTest {
+
+  // We're sure that FirestoreGrpc class comes from this class file because
+  // this project (cloud-opensource-java) doesn't have dependency for Cloud Firestore
+  private String classFileName = "testdata/grpc-google-cloud-firestore-v1beta1-0.28.0_FirestoreGrpc.class";
```
garrettjonesgoogle: This appears to be a constant, so use UPPER_SNAKE_CASE for the name.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
```diff
@@ -0,0 +1,117 @@
...
+  public void testListDeclaredMethods() throws IOException {
+    List<MethodAndSignature> signatures = ClassDumper.listDeclaredMethods(classFileInputStream, classFileName);
+
+    Truth.assertThat(signatures).hasSize(45);
+
+    // getRunQueryMethod is a method defined in FirestoreGrpc class
+    // About type parameters:
+    // While getRunQueryMethod's string representation contains type parameters:
+    //   "public static io.grpc.MethodDescriptor getRunQueryMethod() [Signature: ()Lio/grpc/MethodDescriptor<Lcom/google/firestore/v1beta1/RunQueryRequest;Lcom/google/firestore/v1beta1/RunQueryResponse;>;][RuntimeInvisibleAnnotations]";
+    // , these parameters don't appear in signature field of BCEL Method class
```
garrettjonesgoogle: This is just type erasure, right?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -0,0 +1,180 @@
...
+   * @throws IOException when there is problem in reading classFileStream
+   */
+  public static List<MethodAndSignature> listDeclaredMethods(InputStream classFileStream, String fileName)
+      throws IOException {
+    final ClassParser parser = new ClassParser(classFileStream, fileName);
+    final JavaClass javaClass = parser.parse();
+    Method[] methods = javaClass.getMethods();
+
+    List<MethodAndSignature> signatures = Arrays.stream(methods)
+        .map(method -> new MethodAndSignature(method.getName(), method.getSignature()))
```
garrettjonesgoogle: So it appears that BCEL uses non-standard terminology. Here is Oracle's definition:

https://docs.oracle.com/javase/tutorial/java/javaOO/methods.html

"method signature" is supposed to be the combination of the method name and the parameters, it isn't just supposed to be the parameter types (as BCEL seems to use "signature" for). I'd prefer our tool to use the correct terminology. That would imply `MethodSignature` for the class name instead of `MethodAndSignature`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -0,0 +1,180 @@
...
+import org.apache.bcel.classfile.ClassParser;
+import org.apache.bcel.classfile.Constant;
+import org.apache.bcel.classfile.ConstantMethodref;
+import org.apache.bcel.classfile.ConstantNameAndType;
+import org.apache.bcel.classfile.ConstantPool;
+import org.apache.bcel.classfile.JavaClass;
+import org.apache.bcel.classfile.Method;
+
+/**
+ * This class reads Java class file to analyze following attributes:
```
garrettjonesgoogle: s/reads Java class file/reads a Java class file/

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ConstantPoolMethodref.java
```diff
@@ -0,0 +1,79 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.opensource.classpath;
+
+import com.google.common.base.MoreObjects;
+import java.util.Objects;
+
+/**
+ * A representation of an Methodref entry in constant pool.
```
garrettjonesgoogle: s/an Methodref/a Methodref/, s/in constant pool/in a constant pool/.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ConstantPoolMethodref.java
```diff
@@ -0,0 +1,79 @@
...
+ * In Java class file ConstantPool section,
```
garrettjonesgoogle: s/In Java .../In the Java .../

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ConstantPoolMethodref.java
```diff
@@ -0,0 +1,79 @@
...
+ * a methodref entry has a pointer (index) of the actual value of class name, method name and its
+ * type information
+ * On the other hand, this class holds these values store directly so that we can read them easily
+ */
+class ConstantPoolMethodref {
```
garrettjonesgoogle: This class seems to just be `MethodAndSignature` (which I recommend naming `MethodSignature` in another comment) plus the fully-qualified class name.

Since this class doesn't store indexes, I don't think `Methodref` is a good way to describe this.

I feel like maybe this class should be named `FullyQualifiedMethodSignature`, and should hold a String for the fully-qualified class name and a `MethodSignature` instance for the method name + parameter list.

garrettjonesgoogle: This class seems to just be `MethodAndSignature` (which I recommend naming `MethodSignature` in another comment) plus the fully-qualified class name.

Since this class doesn't store indexes, I don't think `Methodref` is a good way to describe this.

I feel like maybe this class should be named `FullyQualifiedMethodSignature`, and should hold a String for the fully-qualified class name and a `MethodSignature` instance for the method name + parameter list.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -0,0 +1,180 @@
...
+   *
+   * @param classFileStream stream of a class file
+   * @param fileName name of the file that contains class
+   * @return List of all methodref entries defined in ConstantPool
+   * @throws IOException when there is problem in reading classFileStream
+   */
+  @VisibleForTesting
+  static List<ConstantPoolMethodref> listConstantPoolMethodref(InputStream classFileStream,
+      String fileName) throws IOException {
+    ClassParser parser = new ClassParser(classFileStream, fileName);
```
garrettjonesgoogle: I find it odd that we need to pass both the class file stream and the file name that the class file stream is derived from - do you know why `ClassParser` requires that?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -0,0 +1,180 @@
...
+   */
+  public static List<ConstantPoolMethodref> listOwningConstantPoolMethodref(InputStream classFileStream,
+      String fileName) throws IOException {
+    ClassParser parser = new ClassParser(classFileStream, fileName);
+    JavaClass javaClass = parser.parse();
+    List<ConstantPoolMethodref> methodrefs = listConstantPoolMethodRef(javaClass);
+    Set<String> owningClass = Sets.newHashSet(javaClass.getClassName());
+    List<ConstantPoolMethodref> owningMethodrefs =  methodrefs
+            .stream()
+            .filter(methodref -> owningClass.contains(methodref.getClassName()))
```
garrettjonesgoogle: I don't understand why you're creating a set with a single element, then checking that a single string is in that set - why aren't you using string equality?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -0,0 +1,180 @@
...
+            .collect(Collectors.toList());
+    return owningMethodrefs;
+  }
+
+  /**
+   *  Lists the methodref entries not owned by the defined class in class file
+   *  The list the class attemts to invoke the methods defined elsewhere
```
garrettjonesgoogle: s/attemts/attempts/

The grammar here doesn't quite work.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -0,0 +1,180 @@
...
+   *
+   * @param classFileStream stream of a class file
+   * @param fileName name of the file that contains class
+   * @return methodref entries external to the class
+   * @throws IOException when there is problem in reading classFileStream
+   */
+  public static List<ConstantPoolMethodref> listExternalConstantPoolMethodref(
```
garrettjonesgoogle: I don't think we need to include "constant pool" in this method name; I feel like that is an implementation detail. I would suggest `listExternalMethodReferences`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -0,0 +1,180 @@
...
+
+  /**
+   *  Lists the methodref entries owned by the defined class in class file
+   *
+   * @param classFileStream stream of a class file
+   * @param fileName name of the file that contains class
+   * @return methodref entries owned by the class
+   * @throws IOException when there is problem in reading classFileStream
+   */
+  public static List<ConstantPoolMethodref> listOwningConstantPoolMethodref(InputStream classFileStream,
```
garrettjonesgoogle: I don't think we need to include "constant pool" in this method name; I feel like that is an implementation detail. Also, I think it would make more sense for this method to be named the "opposite" of the other one and use "internal" instead of "owing." Thus, I would suggest `listInternalMethodReferences`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -0,0 +1,180 @@
...
+   * @param classFileStream stream of a class file
+   * @param fileName name of the file that contains class
+   * @return method and signature entries defined in the class file
+   * @throws IOException when there is problem in reading classFileStream
+   */
+  public static List<MethodAndSignature> listDeclaredMethods(InputStream classFileStream, String fileName)
+      throws IOException {
+    final ClassParser parser = new ClassParser(classFileStream, fileName);
+    final JavaClass javaClass = parser.parse();
+    Method[] methods = javaClass.getMethods();
```
garrettjonesgoogle: Does this return private methods in the result?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -0,0 +1,180 @@
...
+ * This class reads Java class file to analyze following attributes:
+ *
+ * <ol>
+ *   <li>source (defined methods) via Method fields in the class, and</li>
+ *   <li>targets (what's attempted to be invoked) via ConstantPool field of the class</li>
+ * </ol>
+ */
+class ClassDumper {
+
+  private static List<ConstantPoolMethodref> listConstantPoolMethodRef(JavaClass javaClass) {
```
garrettjonesgoogle: I would recommend `listMethodReferences`

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -0,0 +1,180 @@
...
+            .filter(constant -> constant.getTag() == Const.CONSTANT_Methodref)
+            .collect(Collectors.toList());
+    for (Constant constant : methodrefConstants) {
+      ConstantMethodref constantMethodref = (ConstantMethodref) constant;
+      String className = constantMethodref.getClass(constantPool);
+      int nameAndTypeIndex = constantMethodref.getNameAndTypeIndex();
+      Constant constantNameAndTypeRaw = constantPool.getConstant(nameAndTypeIndex);
+      if (!(constantNameAndTypeRaw instanceof ConstantNameAndType)) {
+        throw new RuntimeException(
+            "Failed to lookup nameAndType constant indexed " + nameAndTypeIndex);
```
garrettjonesgoogle: In what circumstances does this happen?

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
suztomo: I think yes, it's related to type erasure.
The comment is to clarify that we don't expected type parameters in MethodSignature.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Yes, I'll update the name.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
garrettjonesgoogle: I think we need to treat them specially. We could have a situation where a public method becomes private, in which case it was a breaking change. Providing an error message that it doesn't exist could potentially be confusing. This is distinct from a public method being renamed to something else - the symbol is completely gone. So, we might want to return it and track that it is private somehow. (I would suggest making this a later enhancement though; I was just curious at this point.)

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
garrettjonesgoogle: Right - that's implied by type erasure. I think you can just mention type erasure, then anyone who is curious can go look that up.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
garrettjonesgoogle: Ok, good to know.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
garrettjonesgoogle: But when will a class file not be compliant with this? It seems like a case we "shouldn't" hit, which means we potentially have a corrupted class file or something. (If that is true, we might want to describe the error message in that way, i.e. that the class file is corrupted.) Or is my assumption incorrect?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -0,0 +1,182 @@
...
+      String descriptor = constantNameAndType.getSignature(constantPool);
+      FullyQualifiedMethodSignature methodref = new FullyQualifiedMethodSignature(className,
+          methodName, descriptor);
+      methodReferences.add(methodref);
+    }
+    return methodReferences;
+  }
+
+  /**
+   *  Lists all Methodref entries defined in a constant pool table of a class file
```
garrettjonesgoogle: This seems to be describing the method in terms of its implementation, but I think the first sentence should be focused on the the caller cares about. Something like: "Lists all method references from the class file." Then, after that you could talk about the implementation. In particular, I don't think we should talk about `Methodref` as a concept (that is part of BCEL and is an implementation detail for us). 

(Adjust the first sentences for the javadocs of the other methods in the same way)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSignature.java
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * A representation of a tuple of method name and descriptor (type information) in a Method entry
+ * in a class file. The descriptor part helps to distinguish method entries when method overloading
+ * is used
+ */
+public class MethodSignature {
+  private String methodName;
+  private String descriptor;
+
+  public MethodSignature(String methodName, String descriptor) {
```
garrettjonesgoogle: I don't think "descriptor" is quite right. Maybe "parameterListDescriptor", because this is just the parameter list, it doesn't describe the whole method signature.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
suztomo: Sure.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+  @Test
+  public void testListConstantPool() throws IOException {
+    List<String> constantPool = ClassDumper.listConstantPool(classFileInputStream,
+        EXAMPLE_CLASS_FILE);
+    Truth.assertThat(constantPool).hasSize(491);
+  }
+
+  @Test
+  public void testListMethodref() throws IOException {
```
garrettjonesgoogle: rename to `testListMethodReferences`

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSignature.java
suztomo: It includes return types too e.g., "(Lcom/google/protobuf/Message;)Lio/grpc/MethodDescriptor$Marshaller;"

dimtegbydsn

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Nice. I'll follow it.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
suztomo: Yes

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Sure, I'll make it later enhancement.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
suztomo: Now it's

>     // Because of type erasure, type parameters don't appear in signature of BCEL Method class

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Now these first sentences are simplified and without implementation detail:

>   *  Lists all method references from the class file. The output corresponds to
>  *  CONSTANT_Methodref_info entries its constant pool table

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -0,0 +1,186 @@
...
+      FullyQualifiedMethodSignature methodref = new FullyQualifiedMethodSignature(className,
+          methodName, descriptor);
+      methodReferences.add(methodref);
+    }
+    return methodReferences;
+  }
+
+  /**
+   *  Lists all method references from the class file. The output corresponds to
+   *  CONSTANT_Methodref_info entries its constant pool table
```
garrettjonesgoogle: Nit: end the javadoc with a period (applies to all javadocs in your PR)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSignature.java
garrettjonesgoogle: Oh thanks! I learned something :-)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Sure.

suztomo: Fixed.

suztomo:
@garrettjonesgoogle the nit is now fixed.

Issue #135, opened 2018-10-09
Run static linkage checker by one Maven coordinate


suztomo:
Example coordinate: "com.google.cloud:google-cloud-bigtable:0.66.0-alpha"

- Make [JarFile](https://docs.oracle.com/javase/8/docs/api/java/util/jar/JarFile.html) available via Maven resolver
- Create Path objects
- Call StaticLinkageChecker.findUnresolvedMethodReferencesInJars

Does it show any practical linkage error?

Once this is done, then next step is to apply this for BOM

PR #140, opened 2018-10-11
Coordinate to jars

suztomo:
StaticLinkageChecker to resolve Maven coordinate to jars using DependencyGraphBuilder.getTransitiveDependencies


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -52,25 +59,47 @@
...
    * @throws IOException when there is a problem in reading a jar file
    * @throws ClassNotFoundException when there is a problem in reading a class from a jar file
+   * @throws RepositoryException when there is a problem in resolving the Maven coordinate to jar
+   *     files
    */
-  public static void main(String[] arguments) throws IOException, ClassNotFoundException {
+  public static void main(String[] arguments)
+      throws IOException, ClassNotFoundException, RepositoryException {
+    if (arguments.length < 1) {
+      System.out.println("No argument specified.");
```
elharo: System.err

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -52,25 +59,47 @@
...
+      return;
+    }
+    List<Path> jarFilePaths;
+    if (arguments.length == 1) {
```
elharo: what if there's a single jar file? What if there are multiple maven coordinates? We probably need a more complete CLI here with arguments and such.

suztomo:
#135

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Let me implement it with commons-cli.

suztomo: Done

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Done

suztomo:
@elharo Now it accepts command line arguments (commons-cli). For example:

```
-j /Users/suztomo/.m2/repository/io/grpc/grpc-core/1.15.1/grpc-core-1.15.1.jar,/Users/suztomo/.m2/repository/io/grpc/grpc-context/1.15.1/grpc-context-1.15.1.jar
```

Without any argument, it outputs help:

```
No jar files to scan.
usage: StaticLinkageChecker
 -c,--coordinate <arg>   Maven coordinates (separated by ',')
 -j,--jars <arg>         Jar files (separated by ',')
```

dependencies/pom.xml
```diff
@@ -23,6 +23,11 @@
   </properties>
 
   <dependencies>
+    <dependency>
```
elharo: what other dependencies does this pull in?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -52,25 +65,61 @@
...
-   * @param arguments jar files available in the file system
    * @throws IOException when there is a problem in reading a jar file
    * @throws ClassNotFoundException when there is a problem in reading a class from a jar file
+   * @throws RepositoryException when there is a problem in resolving the Maven coordinate to jar
+   *     files
    */
-  public static void main(String[] arguments) throws IOException, ClassNotFoundException {
+  public static void main(String[] arguments)
+      throws IOException, ClassNotFoundException, RepositoryException {
+    Options options = new Options();
```
elharo: This logic is complex enough to pull into a separate method.

dependencies/pom.xml
elharo: Looks like this adds no transitive compile scope dependencies.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
garrettjonesgoogle: +1 I usually like to have option parsing and the program logic in separate functions

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -84,6 +133,30 @@ public static void main(String[] arguments) throws IOException, ClassNotFoundExc
...
+   *
+   * @param coordinate Maven coordinate of an artifact to check its dependencies
+   * @return list of absolute paths to jar files
+   * @throws RepositoryException when there is a problem in retrieving jar files
+   */
+  static List<Path> coordinateToJarPaths(String coordinate) throws RepositoryException {
+    DefaultArtifact artifact = new DefaultArtifact(coordinate);
+    DependencyGraph transitiveDependencies =
+        DependencyGraphBuilder.getTransitiveDependencies(artifact);
+    List<DependencyPath> dependencyPaths = transitiveDependencies.list();
```
garrettjonesgoogle: If there are multiple versions of an artifact on the classpath, will this process them both?

dependencies/pom.xml
suztomo: Yes.

```
$ mvn dependency:tree
...
[INFO] +- com.google.cloud.tools.opensource:dependency-lister:jar:1.0-SNAPSHOT:compile
[INFO] |  +- commons-cli:commons-cli:jar:1.4:compile
[INFO] |  +- org.apache.bcel:bcel:jar:6.2:compile
[INFO] |  +- org.apache.maven:maven-resolver-provider:jar:3.5.4:compile
...
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Fixed

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Yes, the list would includes both in that case.

However, the order may not be what Maven uses (BFS)...

garrettjonesgoogle: This is precisely why I think we should not have our own app try to figure out the jars; we should get the literal classpath passed in from Maven. We want to check exactly what the user will see when they run their app or use their library.

suztomo: > Yes, the list would includes both in that case.

This was incorrect. It only list one for an artifact.

suztomo: For grpc-auth 1.15.1 , https://storage.googleapis.com/cloud-opensource-java-dashboard/dashboard/target/dashboard/io.grpc_grpc-auth_1.15.1.html

 DependencyGraphBuilder.getTransitiveDependencies gives:

```
/Users/suztomo/.m2/repository/io/grpc/grpc-auth/1.15.1/grpc-auth-1.15.1.jar, 
 /Users/suztomo/.m2/repository/io/grpc/grpc-core/1.15.1/grpc-core-1.15.1.jar,
 /Users/suztomo/.m2/repository/io/grpc/grpc-context/1.15.1/grpc-context-1.15.1.jar,
 /Users/suztomo/.m2/repository/com/google/code/gson/gson/2.7/gson-2.7.jar,
 /Users/suztomo/.m2/repository/com/google/guava/guava/20.0/guava-20.0.jar,
 /Users/suztomo/.m2/repository/com/google/errorprone/error_prone_annotations/2.2.0/error_prone_annotations-2.2.0.jar,
 /Users/suztomo/.m2/repository/com/google/code/findbugs/jsr305/3.0.0/jsr305-3.0.0.jar,
 /Users/suztomo/.m2/repository/org/codehaus/mojo/animal-sniffer-annotations/1.17/animal-sniffer-annotations-1.17.jar,
 /Users/suztomo/.m2/repository/io/opencensus/opencensus-api/0.12.3/opencensus-api-0.12.3.jar,
 /Users/suztomo/.m2/repository/io/opencensus/opencensus-contrib-grpc-metrics/0.12.3/opencensus-contrib-grpc-metrics-0.12.3.jar,
 /Users/suztomo/.m2/repository/com/google/auth/google-auth-library-credentials/0.9.0/google-auth-library-credentials-0.9.0.jar]
```

garrettjonesgoogle: I see. It looks like at least for the cases we have coded for, it's correct, so maybe we can go with this for now. I'm still concerned about edge cases long term though, and I don't trust us to match Maven's class path in 100% of cases.

suztomo:
@garrettjonesgoogle @elharo Updated code and responded to comment. Can you check?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -75,6 +76,45 @@
    */
   public static void main(String[] arguments)
       throws IOException, ClassNotFoundException, RepositoryException {
+    List<Path> jarFilePaths = parseArguments(arguments);
+    if (jarFilePaths == null) {
+      System.err.println("Could not parse arguments to jar files.");
```
garrettjonesgoogle: This seems redundant with what `parseArguments` already prints out.

suztomo: removed.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
garrettjonesgoogle: @elharo do you have an opinion on this matter?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -84,6 +106,75 @@ public static void main(String[] arguments) throws IOException, ClassNotFoundExc
...
+      }
+    } catch (ParseException ex) {
+      System.err.println("Failed to parse command line arguments: " + ex.getMessage());
+      formatter.printHelp("StaticLinkageChecker", options);
+      return null;
+    }
+    if (jarFilePaths.size() < 1) {
+      System.err.println("No jar files to scan.");
+      formatter.printHelp("StaticLinkageChecker", options);
+      return null;
```
garrettjonesgoogle: I might prefer to actually throw an exception here - we can't do anything useful in the calling function after this point. (same with the `return null` a few lines up too)

suztomo: Removed redundancy with raising RuntimeException

suztomo:
@garrettjonesgoogle Can you check the fix? Now the error message looks like this:

```
Failed to parse command line arguments: Unrecognized option: -x
No jar files to scan
Exception in thread "main" java.lang.RuntimeException: Could not list jar files for given argument.
	at com.google.cloud.tools.opensource.classpath.StaticLinkageChecker.parseArguments(StaticLinkageChecker.java:142)
	at com.google.cloud.tools.opensource.classpath.StaticLinkageChecker.main(StaticLinkageChecker.java:79)
usage: StaticLinkageChecker
 -c,--coordinate <arg>   Maven coordinates (separated by ',')
 -j,--jars <arg>         Jar files (separated by ',')
```

garrettjonesgoogle:
After the "Failed to parse command line arguments" message, shouldn't it print the help and throw an exception? I would suggest throwing `IllegalArgumentException` from `parseArguments` instead of `RuntimeException`, and then at the top level, don't print the stack trace for `IllegalArgumentException`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
elharo: I'm not sure what you mean by the "literal classpath passed in from Maven". We can handle multiple classpath's and if we can get maven to tell us which classpath it's using under what circumstances, then we can use that with this code. Off the top of my head, I don't know how to get Maven to tell us that.

suztomo:
Changed exception to IllegalArgumentException. The strange order of the help message and the stack trace are due to difference on stdout and stderr. Screenshot explains better:

screen shot 2018-10-11 at 14 29 39


>  at the top level, don't print the stack trace

Not printing stacktrace would require following code. This fails "initialization at declaration" and ignores an exception.

```
    List<Path> jarFilePaths;
    try {
      jarFilePaths = parseArguments(arguments);
    } catch (IllegalArgumentException ex) {
      // Error messages are printed in parseArguments
      return;
    }
```

I prefer

```
jarFilePaths = parseArguments(arguments);
```

Let JVM to print stacktrace if any. @garrettjonesgoogle  What do you think?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
elharo: We started with Maven plugins. However we rapidly ran into several related problems:

1. The plugins are not designed to be invoked as a library. You have to spawn a process.
2. The plugins only operate on a source code project with a pom.xml, not an artifact in the repo. 
3. The output of the plugins isn't trivially parseable. It's designed to be read, not processed.

suztomo:
@garrettjonesgoogle @elharo Updated files and answered questions. Can you check

suztomo:
@elharo Merged 'origin/master' into coordinate_to_jars

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -84,6 +101,72 @@ public static void main(String[] arguments) throws IOException, ClassNotFoundExc
...
   }
 
+  /**
+   * Parses arguments to get list of jar file paths.
+   *
+   * @param arguments command-line arguments
+   * @return list of the absolute paths to jar files
+   * @throws RepositoryException when there is a problem in resolving the Maven coordinate to jar
+   */
+  static List<Path> parseArguments(String[] arguments) throws RepositoryException {
```
elharo: @VisibleForTesting if need be, otherwise private

suztomo: Missed this comment. I'll  add the annotation in next PR.

PR #141, opened 2018-10-11
Sort the global list of dependencies

elharo:
@garrettjonesgoogle @suztomo Some of the things we pull in are quite surprising. E.g Avalon?!

We may want to add some exploratory tools to this so we can easily dig in and find out where detritus like this is coming from.

### DIFF
--- a/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
+++ b/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
@@ -124,6 +124,7 @@ public class DashboardTest {
       Nodes li = document.query("//li");
       Assert.assertTrue(li.size() > 100);
       ArrayList<String> coordinateList = new ArrayList<>();
+
       for (int i = 0; i < li.size(); i++) {
         String coordinates = li.get(i).getValue();
         // fails if these are not valid Maven coordinates
@@ -138,8 +139,6 @@ public class DashboardTest {
         Assert.assertEquals("Coordinates are not sorted: " + i, sorted.get(i),
             coordinateList.get(i));
       }
-      
-      Assert.assertEquals("Coordinates are not sorted", sorted, coordinateList);
       
       Nodes updated = document.query("//p[@id='updated']");
       Assert.assertEquals("didn't find updated" + document.toXML(), 1, updated.size());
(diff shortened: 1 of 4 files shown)


dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -121,10 +124,22 @@ public void testDashboard()
...
       for (int i = 0; i < li.size(); i++) {
         String coordinates = li.get(i).getValue();
         // fails if these are not valid Maven coordinates
         new DefaultArtifact(coordinates);
+        coordinateList.add(coordinates);
+      }
+      
+      ArrayList<String> sorted = new ArrayList<>(coordinateList);
+      Comparator<String> comparator = new SortWithoutVersion();
+      Collections.sort(sorted, comparator );
```
suztomo: Unnecessary space after `comparator`

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -121,10 +124,22 @@ public void testDashboard()
...
+      
+      for (int i = 0; i < sorted.size(); i++) {
+        Assert.assertEquals("Coordinates are not sorted: ", sorted.get(i),
+            coordinateList.get(i));
```
garrettjonesgoogle: It seems like we could have much more concise assertions by using the `Truth` library

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
elharo: fixed

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
elharo: Specific suggestions? I looked there and didn't find anything on point. I did have to replace a general equals check on the list with an iterated check on each element to find the exact elements that were sorted out of order when my test failed.

garrettjonesgoogle: I think you have to combine a couple things, like `containsExactly` and `inOrder`. I'm not sure what the error messaging looks like when it fails.

Consider this optional if you want to look into it.

suztomo: `Truth.assertThat(coordinateList).containsExactlyElementsIn(sorted).inOrder();`?
http://google.github.io/truth/api/latest/com/google/common/truth/Ordered.html

PR #143, opened 2018-10-11
Refactor to do all parsing and querying of POMs before rather than du…

elharo:
@suztomo Refactor to do all parsing and querying of POMs before rather than during report generation. This will enable us to produce more reports, especially global reports, without hitting Maven central repeatedly to retrieve the same information.

### DIFF
--- a/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardUnavailableArtifactTest.java
+++ b/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardUnavailableArtifactTest.java
@@ -114,7 +114,7 @@ public class DashboardUnavailableArtifactTest {
     table.add(validArtifactResult);
     table.add(errorArtifactResult);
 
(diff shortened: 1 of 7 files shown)


dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardUnavailableArtifactTest.java
```diff
@@ -83,7 +92,7 @@ public void testDashboardForRepositoryException() {
         errorArtifactResult.getResult(DashboardMain.TEST_NAME_UPPER_BOUND));
     Assert.assertEquals(
         "The error artifact result should contain error message",
-        "Could not find artifact io.grpc:nonexistent:jar:1.15.0 in central (http://repo1.maven.org/maven2/)",
+        "foo",
```
suztomo: foo?

suztomo: found at line 76

PR #147, opened 2018-10-12
add global upper bounds to front page

elharo:
@suztomo @garrettjonesgoogle This code adds a global upper bounds (how many dependencies should be upgraded to a more recent release used somewhere in the BOM's full dependency tree) result to the front page. 

Next step is to add details about suggested upgrades to the individual artifact pages.


dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardUnavailableArtifactTest.java
```diff
@@ -75,8 +75,10 @@ public void testDashboardForRepositoryException() {
     map.put(validArtifact, new ArtifactInfo(graph1, graph2));
     map.put(nonExistentArtifact, new ArtifactInfo(new RepositoryException("foo")));
     
+    ArtifactCache cache = new ArtifactCache();
+    cache.setInfoMap(map);
     List<ArtifactResults> artifactResults =
-        DashboardMain.generateReports(configuration, outputDirectory, map);
+        DashboardMain.generateReports(configuration, outputDirectory, cache );
```
garrettjonesgoogle: Remove extra space before the `)`

PR #148, opened 2018-10-12
combine duplicate code

elharo:
@suztomo


dashboard/src/main/resources/templates/dashboard.ftl
```diff
@@ -50,63 +50,37 @@
   <body>
     <h1>Google Cloud Platform Dependency Dashboard</h1>
     <h2>Artifact Details</h2>
+
```
suztomo: Great use of macro.

PR #156, opened 2018-10-15
externalize stylesheet

### LINKED ISSUE #152: External CSS stylesheet
That can be shared between the main dashboard page and the individual artifact pages.

elharo:
@suztomo fix #152 

It feels like this should be possible from Maven alone, but it doesn't seem to be. It's hard to get a reference to the correct output directory.

### DIFF
--- a/dashboard/src/main/resources/templates/component.ftl
+++ b/dashboard/src/main/resources/templates/component.ftl
@@ -1,7 +1,7 @@
 <html lang="en-US">
   <head>
     <title>Google Cloud Platform Dependency Analysis Report for ${groupId}:${artifactId}:${version}</title>
-    <link rel="stylesheet" type="text/css" href="dashboard.css">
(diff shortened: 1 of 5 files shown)


dashboard/src/main/resources/templates/component.ftl
```diff
@@ -1,6 +1,7 @@
 <html lang="en-US">
   <head>
     <title>Google Cloud Platform Dependency Analysis Report for ${groupId}:${artifactId}:${version}</title>
+    <link rel="stylesheet" type="text/css" href="dashboard.css">
```
suztomo: Circle CI complains:

```
nu.xom.ParsingException: The element type "link" must be terminated by the matching end-tag "</link>". at line 5, column 5 in file:///home/circleci/cloud-opensource-java/dashboard/target/dashboard/dashboard.htmlv
```

PR #161, opened 2018-10-15
maven resolver 1.3.1

### LINKED ISSUE #138: Upgrade to Maven resolver 1.3.1
when it's released. 1.3.0 just came out but it has a NullPointerException

elharo:
@suztomo fix #138

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -20,14 +20,10 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <resolverVersion>1.3.1</resolverVersion>
   </properties>
 
   <dependencies>
-    <dependency>
-      <groupId>commons-cli</groupId>
-      <artifactId>commons-cli</artifactId>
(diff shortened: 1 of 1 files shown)


dependencies/pom.xml
```diff
@@ -41,22 +37,27 @@
...
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
       <version>4.5.6</version>
     </dependency>
     <dependency>
-      <groupId>org.apache.maven.resolver</groupId>
-      <artifactId>maven-resolver-connector-basic</artifactId>
-      <version>1.1.1</version>
+      <groupId>commons-cli</groupId>
```
suztomo: Isn't this pom supposed to have alphabetical order?

elharo: No, it isn't. The order of dependencies in pom.xml effects which transitive dependency versions are pulled in. Thus it's useful to put the more critical and fragile dependencies first.

PR #164, opened 2018-10-15
Level-order (BFS) to resolve maven dependency graph

### LINKED ISSUE #162: Ability to list jar files given list of artifact (dependency) as if it's resolved by Maven
Ability to list jar files, given a list of artifacts (dependencies) as if it's resolved by Maven.

Example:

input:

```
["com.google.cloud:google-cloud-datastore:1.37.1",
 "com.google.guava:guava:25.1-jre"]
```

output:

```
   ["com.google.cloud:google-cloud-datastore:1.37.1",
    "com.google.guava:guava:25.1-jre",                 // BFS (level-order) picks this dependency before ones below google-cloud-datastore
    "com.google.cloud:google-cloud-core:jar:1.48.0",   // The first child of google-cloud-datastore
    ...

```

The output is then to be used by StaticLinkageChecker to confirm two artifacts and their dependencies are not having linkage conflict.

Memo: does DependencyGraph have such attribute? https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/5f47c1e4c69e48086f8d38eeae5c7d68b808a131/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java#L49

suztomo:
Level-order (BFS) to resolve maven dependency graph, and ability to get dependency graph from 2 or more artifacts.

Fixes #162

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -48,7 +48,6 @@ import org.eclipse.aether.resolution.DependencyResolutionException;
 import com.google.cloud.tools.opensource.dependencies.Artifacts;
 import com.google.cloud.tools.opensource.dependencies.DependencyGraph;
 import com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder;
-import com.google.cloud.tools.opensource.dependencies.DependencyPath;
 import com.google.cloud.tools.opensource.dependencies.DependencyTreeFormatter;
 import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
 import com.google.cloud.tools.opensource.dependencies.Update;
@@ -148,9 +147,8 @@ public class DashboardMain {
     
     for (Artifact artifact : artifacts) {
       try {
-        // Pre-order is for showing dependency tree
         DependencyGraph completeDependencies =
-            DependencyGraphBuilder.getCompleteDependenciesInPreorder(artifact);
+            DependencyGraphBuilder.getCompleteDependencies(artifact);
         globalDependencies.add(completeDependencies);
         
         // picks versions according to Maven rules
@@ -196,9 +194,8 @@ public class DashboardMain {
       Map<Artifact, Artifact> globalUpperBoundFailures = findUpperBoundsFailures(
           collectLatestVersions(globalDependencies), transitiveDependencies);
 
-      List<DependencyPath> dependenciesInPreorder = completeDependencies.list();
       String dependencyTree =
-          DependencyTreeFormatter.formatDependencyPaths(dependenciesInPreorder);
+          DependencyTreeFormatter.formatDependencyPaths(completeDependencies.list());
       Template report = configuration.getTemplate("/templates/component.ftl");
 
       Map<String, Object> templateData = new HashMap<>();
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -16,19 +16,6 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
-import com.google.cloud.tools.opensource.dependencies.Artifacts;
-import com.google.cloud.tools.opensource.dependencies.DependencyGraph;
-import com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder;
-import com.google.cloud.tools.opensource.dependencies.DependencyPath;
-import com.google.cloud.tools.opensource.dependencies.DependencyTreeFormatter;
-import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
-import com.google.cloud.tools.opensource.dependencies.Update;
-import com.google.cloud.tools.opensource.dependencies.VersionComparator;
-import com.google.common.annotations.VisibleForTesting;
-import freemarker.template.Configuration;
-import freemarker.template.Template;
-import freemarker.template.TemplateException;
-import freemarker.template.Version;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -46,12 +33,27 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.TreeMap;
+
+import freemarker.template.Configuration;
+import freemarker.template.Template;
+import freemarker.template.TemplateException;
+import freemarker.template.Version;
 import org.eclipse.aether.RepositoryException;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.artifact.DefaultArtifact;
 import org.eclipse.aether.collection.DependencyCollectionException;
 import org.eclipse.aether.resolution.ArtifactDescriptorException;
 import org.eclipse.aether.resolution.DependencyResolutionException;
+
+import com.google.cloud.tools.opensource.dependencies.Artifacts;
+import com.google.cloud.tools.opensource.dependencies.DependencyGraph;
+import com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder;
+import com.google.cloud.tools.opensource.dependencies.DependencyPath;
+import com.google.cloud.tools.opensource.dependencies.DependencyTreeFormatter;
+import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
+import com.google.cloud.tools.opensource.dependencies.Update;
+import com.google.cloud.tools.opensource.dependencies.VersionComparator;
+import com.google.common.annotations.VisibleForTesting;
 
 public class DashboardMain {
   public static final String TEST_NAME_UPPER_BOUND = "Upper Bounds";
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -157,7 +157,7 @@ public class DashboardMain {
   
         ArtifactInfo info = new ArtifactInfo(completeDependencies, transitiveDependencies);
         infoMap.put(artifact, info);
-      } catch (DependencyCollectionException | DependencyResolutionException ex) {
+      } catch (RepositoryException ex) {
         ArtifactInfo info = new ArtifactInfo(ex);
         infoMap.put(artifact, info);
       }
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -75,8 +75,9 @@ public class DashboardMain {
     
     copyCss(output);
 
+    // TODO should pass in maven coordinates as argument
     DefaultArtifact bom =
-        new DefaultArtifact("com.google.cloud:cloud-oss-bom:pom:0.66.0-SNAPSHOT");
+        new DefaultArtifact("com.google.cloud:cloud-oss-bom:pom:0.67.0-SNAPSHOT");
     List<Artifact> managedDependencies = RepositoryUtility.readBom(bom);
 
     ArtifactCache cache = loadArtifactInfo(managedDependencies);
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -16,6 +16,19 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
+import com.google.cloud.tools.opensource.dependencies.Artifacts;
+import com.google.cloud.tools.opensource.dependencies.DependencyGraph;
+import com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder;
+import com.google.cloud.tools.opensource.dependencies.DependencyPath;
+import com.google.cloud.tools.opensource.dependencies.DependencyTreeFormatter;
+import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
+import com.google.cloud.tools.opensource.dependencies.Update;
+import com.google.cloud.tools.opensource.dependencies.VersionComparator;
+import com.google.common.annotations.VisibleForTesting;
+import freemarker.template.Configuration;
+import freemarker.template.Template;
+import freemarker.template.TemplateException;
+import freemarker.template.Version;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -33,26 +46,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.TreeMap;
-
-import freemarker.template.Configuration;
-import freemarker.template.Template;
-import freemarker.template.TemplateException;
-import freemarker.template.Version;
 import org.eclipse.aether.RepositoryException;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.artifact.DefaultArtifact;
 import org.eclipse.aether.collection.DependencyCollectionException;
 import org.eclipse.aether.resolution.ArtifactDescriptorException;
 import org.eclipse.aether.resolution.DependencyResolutionException;
-
-import com.google.cloud.tools.opensource.dependencies.Artifacts;
-import com.google.cloud.tools.opensource.dependencies.DependencyGraph;
-import com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder;
-import com.google.cloud.tools.opensource.dependencies.DependencyTreeFormatter;
-import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
-import com.google.cloud.tools.opensource.dependencies.Update;
-import com.google.cloud.tools.opensource.dependencies.VersionComparator;
-import com.google.common.annotations.VisibleForTesting;
 
 public class DashboardMain {
   public static final String TEST_NAME_UPPER_BOUND = "Upper Bounds";
@@ -147,8 +146,9 @@ public class DashboardMain {
     
     for (Artifact artifact : artifacts) {
       try {
+        // Pre-order is for showing dependency tree
         DependencyGraph completeDependencies =
-            DependencyGraphBuilder.getCompleteDependencies(artifact);
+            DependencyGraphBuilder.getCompleteDependenciesInPreorder(artifact);
         globalDependencies.add(completeDependencies);
         
         // picks versions according to Maven rules
@@ -194,8 +194,9 @@ public class DashboardMain {
       Map<Artifact, Artifact> globalUpperBoundFailures = findUpperBoundsFailures(
           collectLatestVersions(globalDependencies), transitiveDependencies);
 
+      List<DependencyPath> dependenciesInPreorder = completeDependencies.list();
       String dependencyTree =
-          DependencyTreeFormatter.formatDependencyPaths(completeDependencies.list());
+          DependencyTreeFormatter.formatDependencyPaths(dependenciesInPreorder);
       Template report = configuration.getTemplate("/templates/component.ftl");
 
       Map<String, Object> templateData = new HashMap<>();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
@@ -174,8 +174,8 @@ public class DependencyGraphBuilder {
   }
 
   private static final class LevelOrderQueueItem {
-    DependencyNode dependencyNode;
-    Stack<DependencyNode> parentNodes;
+    final DependencyNode dependencyNode;
+    final Stack<DependencyNode> parentNodes;
 
     LevelOrderQueueItem(DependencyNode dependencyNode,
         Stack<DependencyNode> parentNodes) {
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
@@ -147,22 +147,6 @@ public class DependencyGraphBuilder {
     
     return graph;
   }
-
-  /**
-   * Lists compile time dependencies for the artifact shown in pre-order.
-   * Although Maven uses BFS to build dependency list, this method is useful for printing the
-   * dependencies of the artifact as tree.
-   *
-   * @param artifact Maven artifact
-   * @return list of dependencies for the artifact in pre-order
-   */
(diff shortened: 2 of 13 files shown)


dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphIntegrationTest.java
```diff
@@ -104,7 +104,7 @@ public void testJaxen()
     Truth.assertThat(updates).hasSize(5);
     
     List<DependencyPath> conflicts = graph.findConflicts();
-    Truth.assertThat(conflicts).hasSize(33);
+    Truth.assertThat(conflicts).hasSize(30);
```
suztomo: Pre-order and level-order generated different size of conflict. Noted the items in
https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/162

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -147,8 +148,9 @@ private static ArtifactCache loadArtifactInfo(List<Artifact> artifacts) {
     
     for (Artifact artifact : artifacts) {
       try {
+        // Pre-order is for showing dependency tree
         DependencyGraph completeDependencies =
-            DependencyGraphBuilder.getCompleteDependencies(artifact);
+            DependencyGraphBuilder.getCompleteDependenciesInPreorder(artifact);
```
suztomo: Once #165 is fixed, we don't need "preorder" functions.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -18,15 +18,18 @@
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedList;
```
elharo: Despite what data structures textbooks say, you almost never need a linked list, and I can't think of any cases where java.util.LinkedList is appropriate. Default to ArrayList.

https://www.youtube.com/watch?v=YQs6IC-vgmo

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -86,6 +89,35 @@ private static DependencyNode resolveCompileTimeDependencies(Artifact artifact)
     return node;
   }
 
+  private static DependencyNode resolveCompileTimeDependencies(List<Artifact> artifacts)
+      throws DependencyCollectionException, DependencyResolutionException {
+    RepositorySystemSession session = RepositoryUtility.newSession(system);
```
elharo: move this closer to where it's first needed so it isn't constructed if this returns early from the cache

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -109,9 +141,9 @@ public static DependencyGraph getCompleteDependencies(Artifact artifact)
       throws DependencyCollectionException, DependencyResolutionException {
     
     // root node
-    DependencyNode node = resolveCompileTimeDependencies(artifact);  
+    DependencyNode node = resolveCompileTimeRootDependencies(artifact);
     DependencyGraph graph = new DependencyGraph();
-    fullPreorder(new Stack<>(), node, graph);
+    fullLevelorder(node, graph);
```
elharo: fullLevelOrder

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -124,70 +156,98 @@ public static DependencyGraph getCompleteDependencies(Artifact artifact)
...
   public static DependencyGraph getTransitiveDependencies(Artifact artifact)
       throws DependencyCollectionException, DependencyResolutionException {
-    
     // root node
-    DependencyNode node = resolveCompileTimeDependencies(artifact);  
+    DependencyNode node = resolveCompileTimeRootDependencies(artifact);
     DependencyGraph graph = new DependencyGraph();
-    preorder(new Stack<>(), node, graph);
-    
+    levelorder(node, graph);
```
elharo: levelOrder

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -124,70 +156,98 @@ public static DependencyGraph getCompleteDependencies(Artifact artifact)
...
+    DependencyGraph graph = new DependencyGraph();
+    levelorder(node, graph);
+    return graph;
+  }
+
+  private static class LevelOrderQueueItem {
+    DependencyNode dependencyNode;
+    Stack<DependencyNode> parentNodes;
+
+    DependencyNode getDependencyNode() {
```
elharo: don't have to use getters in an inner class, but if you do the fields can be private

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -124,70 +156,98 @@ public static DependencyGraph getCompleteDependencies(Artifact artifact)
...
+      throws DependencyCollectionException, DependencyResolutionException {
+    // root node is dummy (artifact: null) and has dependencies as children
+    DependencyNode node = resolveCompileTimeDependencies(artifacts);
+    DependencyGraph graph = new DependencyGraph();
+    levelorder(node, graph);
+    return graph;
+  }
+
+  private static class LevelOrderQueueItem {
+    DependencyNode dependencyNode;
```
elharo: final

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphIntegrationTest.java
elharo: it would be worth understanding why. What changed in the output?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Good to know. I'm picking ArrayDeque then.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Done

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Done

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Done

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: removed getters

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: done

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphIntegrationTest.java
suztomo: Checking...

suztomo: When it comes to a cyclic dependency, there's a difference in the returned graph of getCompleteDependencies between pre-order and level-order, because the two algorithm grow their dependency trees in different order. When they stop building dependency trees due to a cyclic dependency, the resulting (incomplete) trees look different.

`findConflicts` uses the tree to count the number of conflicting artifacts, the result differs between (incomplete) tree by pre-order and another tree by level-order.

https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/162#issuecomment-430063293

Among the dependency involved in this `testJaxen`, [dom4j:1.5.2](https://search.maven.org/artifact/dom4j/dom4j/1.5.2/jar) and [jaxen:1.1-beta-4](https://search.maven.org/artifact/jaxen/jaxen/1.1-beta-4/jar) have cyclic dependency.

suztomo: When it comes to a cyclic dependency, there's a difference in the returned graph of getCompleteDependencies between pre-order and level-order, because the two algorithm grow their dependency trees in different order. When they stop building dependency trees due to a cyclic dependency, the resulting (incomplete) trees look different.

`findConflicts` uses the tree to count the number of conflicting artifacts, the result differs between an incomplete tree by pre-order and another incomplete tree by level-order.

https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/162#issuecomment-430063293

Among the dependency involved in this `testJaxen`, [dom4j:1.5.2](https://search.maven.org/artifact/dom4j/dom4j/1.5.2/jar) and [jaxen:1.1-beta-4](https://search.maven.org/artifact/jaxen/jaxen/1.1-beta-4/jar) have cyclic dependency.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatter.java
```diff
@@ -61,15 +65,58 @@ private static void printDependencyTree(String coordinate)
...
+    }
+  }
+
+  /**
+   * Builds ListMultiMap that represents a Maven dependency tree of parent-children relationship.
+   * The root node is represented as an empty {@link DependencyPath} and its children are
+   * the values for the root node. As Maven dependency is retrieved via BFS, the order of children
+   * matters.
+   *
+   * @param dependencyPaths a list of dependency path without assuming any order
```
elharo: why "without assuming any order"? It's a list so it's ordered when called. Would the output be the same if the list were shuffled before this were called?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -86,6 +89,35 @@ private static DependencyNode resolveCompileTimeDependencies(Artifact artifact)
     return node;
   }
 
+  private static DependencyNode resolveCompileTimeDependencies(List<Artifact> artifacts)
+      throws DependencyCollectionException, DependencyResolutionException {
+    String key = artifacts.stream().map(Artifacts::toCoordinates).collect(Collectors.joining(","));
+    if (cache.containsKey(key)) {
+      return cache.get(key);
```
garrettjonesgoogle: Will we ever encounter a key more than once that is created by joining artifacts by comma?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -124,70 +156,90 @@ public static DependencyGraph getCompleteDependencies(Artifact artifact)
...
-      if (!"system".equals(node.getDependency().getScope())) {
-        forPath.add(node.getArtifact());
-      }
-    }
-    graph.addPath(forPath);
-    
-    for (DependencyNode child : current.getChildren()) {
-      preorder((Stack<DependencyNode>) path.clone(), child, graph);
+
+  static DependencyGraph getTransitiveDependencies(List<Artifact> artifacts)
```
garrettjonesgoogle: This doesn't appear to be called anywhere, except for its test

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -124,70 +156,90 @@ public static DependencyGraph getCompleteDependencies(Artifact artifact)
...
+          System.err.println("Likely cycle in " + parentNodes);
+          System.err.println("Child " + dependencyNode);
+          return;
+        }
+        parentNodes.push(dependencyNode);
+        graph.addPath(forPath);
+
+        if (!"system".equals(dependencyNode.getDependency().getScope())) {
+          try {
+            dependencyNode = resolveCompileTimeRootDependencies(dependencyNode.getArtifact());
```
garrettjonesgoogle: I can't tell what the difference of intent is between `levelOrder` and `fullLevelOrder` but this seems to be an important difference - what does this line accomplish?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatter.java
```diff
@@ -61,15 +65,58 @@ private static void printDependencyTree(String coordinate)
...
   public static String formatDependencyPaths(List<DependencyPath> dependencyPaths) {
     StringBuilder stringBuilder = new StringBuilder();
+    // Printing text representing a tree requires traversing the items in pre-order
+    ListMultimap<DependencyPath, DependencyPath> tree = buildDependencyPathTree(dependencyPaths);
+    // Empty dependency path is to retrieve children of root node
+    formatDependencyPathTreePreOrder(stringBuilder, tree, new DependencyPath());
+    return stringBuilder.toString();
+  }
+
+  private static void formatDependencyPathTreePreOrder(
```
garrettjonesgoogle: wasn't "pre-order" renamed to "full level order" in other places?

suztomo: When printing dependency tree in text format, pre-order is easier.

As it's confusing, I renamed the function to formatDependencyPaths.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatter.java
suztomo: Previous implementation of tree formatting was relying on the pre-order of `dependencyPaths`.



> Would the output be the same if the list were shuffled before this were called?

Yes, now it doesn't depend on the order of the item. Changed "List" to "Collection" to express this property as type.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Added javadoc to `levelOrder`.

`resolveCompileTimeRootDependencies` itself doesn't dig further when a dependency is `optional: true`. The function `fullLevelOrder`, on the other hand, calls `resolveCompileTimeRootDependencies` every time it finds a dependency so that it can fetch dependencies of a `optional: true` dependency. 

Example is noted in

> jaxen has dom4j as dependency (optional: true

https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/162#issuecomment-430063293

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -60,7 +63,7 @@
   // caching cuts time by about a factor of 4.
   private static final Map<String, DependencyNode> cache = new HashMap<>();
 
-  private static DependencyNode resolveCompileTimeDependencies(Artifact artifact)
+  private static DependencyNode resolveCompileTimeRootDependencies(Artifact artifact)
```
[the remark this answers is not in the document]
elharo: Ping @suztomo

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
elharo: OK. We do need to watch out for adding code in advance of actual use though. Surprisingly often we end up not needing the function we think we do. Also, smaller PRs are easier to review.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
elharo: Oh, I meant make the fields final. You finalized the class (which is fine).

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -124,70 +157,98 @@ public static DependencyGraph getCompleteDependencies(Artifact artifact)
...
+   * resolveFullDependency flag is true, then it resolves the dependency of the artifact of the each
+   * node in the dependency tree; otherwise it just follows the given dependency tree starting with
+   * firstNode.
+   *
+   * @param firstNode node to start traversal
+   * @param graph graph to store {@link DependencyPath} instances
+   * @param resolveFullDependency flag to resolve dependency for each node in the tree. Useful for
+   *     building a complete tree of dependencies including <i>provided</i> scope
+   * @throws DependencyCollectionException when there is a problem in collecting dependency. This
+   *     happens only when resolveFullDependency is true
```
elharo: nit: period

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -124,70 +157,98 @@ public static DependencyGraph getCompleteDependencies(Artifact artifact)
...
+   * @throws DependencyResolutionException when there is a problem in resolving dependency. This
+   *     happens only when resolveFullDependency is true
```
elharo: nit: period

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -124,70 +157,98 @@ public static DependencyGraph getCompleteDependencies(Artifact artifact)
...
+          // TODO: change to logger
+          System.err.println("Infinite recursion resolving " + dependencyNode);
+          System.err.println("Likely cycle in " + parentNodes);
+          System.err.println("Child " + dependencyNode);
+          return;
+        }
+        parentNodes.push(dependencyNode);
+        graph.addPath(forPath);
+
+        if (resolveFullDependency && !"system".equals(dependencyNode.getDependency().getScope())) {
```
elharo: The system exclusion was needed for earlier, existing code. But it might be a problem for static link checking. Please file an issue to look into this.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Some background why this PR grew bigger than initial plan:

- I wanted getTransitiveDependencies(List<Artifact> artifacts)
- Noticed that the DependencyGraph has DependencyPath sorted by pre-order
-- Maven's dependency resolution is level-order
- tree formatting in text requires pre-order

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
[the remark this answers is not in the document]
suztomo: If setRoot is called, then the resulting root DependencyNode has non-null artifact.

If setRoot is not called, then the resulting root DependencyNode has null artifact but it has children DependencyNode.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Fixed.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Created 
https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/167

suztomo:
@elharo Applied the review. Would you check?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
garrettjonesgoogle: But if the cache is by individual artifacts, we could catch all the artifacts in the cache anyway, right? If we put comma-joined keys in, then we'd have the same artifacts represented in a bunch of different ways - the cache would be bigger and more complex to reason about.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
garrettjonesgoogle: Do we still need getTransitiveDependencies(List artifacts) ?

garrettjonesgoogle: The reason I'm pushing on this is that it seems like `resolveCompileTimeRootDependencies` is just the standard thing we want to do, and it seems like it should be named `resolveCompileTimeDependencies`. But, we have a separate `resolveCompileTimeDependencies`, which this function calls, but neither actually seem necessary yet.

suztomo: I also feel it's confusing. Let me revert the renaming of `resolveCompileTimeRootDependencies`.

There are 3 ways to build CollectionRequest. I wanted to separate the 1st case as `resolveCompileTimeRootDependencies`, and 3rd `resolveCompileTimeDependencies `

> There are three ways to create a dependency graph. First, only the root dependency can be given. Second, a root dependency and direct dependencies can be specified in which case the specified direct dependencies are merged with the direct dependencies retrieved from the artifact descriptor of the root dependency. And last, only direct dependencies can be specified in which case the root node of the resulting graph has no associated dependency.

https://maven.apache.org/resolver/apidocs/org/eclipse/aether/collection/CollectRequest.html

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Let me remove this cache. I don't see much use case either. We can implement one when necessary.

Issue #163, opened 2018-10-15
DependencyGraphBuilder to be able to include "optional" dependency


suztomo:
This is to resolve false positive error from optional Log4j dependency in #151. (Note that this doesn't resolve grpc-netty-shaded's dependency on lzma-java)

As per discussion with @elharo , DependencyGraphBuilder to be able to include "optional" dependency (at least for now). Check DependencySelector and Dependency.isOptional method.

suztomo:
```
com.google.cloud:google-cloud-bigtable:jar:0.66.0-alpha (scope: compile, optional: false)
  com.google.cloud:google-cloud-core:jar:1.48.0 (scope: compile, optional: false)
    com.google.http-client:google-http-client:jar:1.24.1 (scope: compile, optional: false)
      org.apache.httpcomponents:httpclient:jar:4.5.3 (scope: compile, optional: false)
        commons-logging:commons-logging:jar:1.2 (scope: compile, optional: false)
          log4j:log4j:jar:1.2.17 (this doesn't appear in the dependency tree from `resolveCompileTimeRootDependencies`.
```

Probably because https://search.maven.org/artifact/commons-logging/commons-logging/1.2/jar has log4j as "optional".

<img width="1277" alt="screen shot 2018-10-17 at 14 16 23" src="https://user-images.githubusercontent.com/28604/47107496-7abbaa00-d217-11e8-942e-9d33b4b07078.png">

PR #169, opened 2018-10-17
Round robin table

suztomo:
This is a tool to generate a table like this from BOM:

screen shot 2018-10-17 at 13 17 59

Each cell represents the result of static linkage check using the class path from the 2 artifacts.


suztomo:
Talked with @elharo . Closing this PR for now. To focus on false positives #163 . I'll revisit this once we remediate more false positives.

PR #177, opened 2018-10-18
Adding JLBP-11: Stay up to date with compatible dependencies

garrettjonesgoogle:


library-best-practices/JLBP-11.md
```diff
@@ -0,0 +1,7 @@
+[JLBP-11] Frequent release cadence for non-breaking changes
```
elharo: Phrase imperatively: Release non-breaking changes frequently

library-best-practices/JLBP-11.md
```diff
@@ -0,0 +1,7 @@
...
+-----------------------------------------------------------
+
+- Keep compatible dependencies up-to-date.
+- Recommendation: Perform a release at least once every 6-weeks.
```
elharo: Perform a release --> Release

Every six weeks is very quick for projects in maintenance mode. I might suggest "at least twice a year and preferably every two months".

How feasible this is depends a lot on the automation of the build, test, release process. I wonder if we should say something about that?

library-best-practices/JLBP-11.md
```diff
@@ -0,0 +1,7 @@
...
+  - Don't be more than 6 weeks out of date with your dependencies.
```
elharo: Perhaps measure in versions rather than time? One minor release?

library-best-practices/JLBP-11.md
garrettjonesgoogle: Hmm I realized that my original title (and thus your variant) has the wrong implications - it implies that libraries should just keep adding things, regardless of their value. I don't actually care how frequently libraries push out new features - it could be never for all I care.

The more important thing is making releases when dependencies make new releases. How about: "Stay up to date with compatible dependencies"?

library-best-practices/JLBP-11.md
garrettjonesgoogle: I think the time dimension is important, because having the ecosystem generally up to date at each point in time results in ecosystem snapshots that are more consistent, and more likely to have compatibility across the board.

library-best-practices/JLBP-11.md
garrettjonesgoogle: I think projects in maintenance mode either need to be pulled out of the tree altogether, or actually have dependencies kept up to date. Otherwise, it's a huge source of pain for the ecosystem. (See: google-*-client-java libraries).

garrettjonesgoogle:
Feedback addressed, PTAL

garrettjonesgoogle:
Postscript: file https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/181

library-best-practices/JLBP-11.md
```diff
@@ -0,0 +1,12 @@
+[JLBP-11] Stay up to date with compatible dependencies
```
elharo: Is "compatible" needed in the title here? What's an incompatible dependency?

library-best-practices/README.md
```diff
@@ -9,6 +9,10 @@ made. We have found that following these rules will result in higher quality
 Java libraries with fewer dependency conflicts and other kinds of problems. The
 list is open-ended, so new ones may be added from time to time.
 
+*Note* This list is currently in "Beta," which means that we don't anticipate
```
elharo: I'd go stronger: This list is in alpha. It's an early draft and we're still developing it.

library-best-practices/JLBP-11.md
garrettjonesgoogle: This is to indicate that it's not a requirement to jump to new major versions in a short time frame.

library-best-practices/README.md
garrettjonesgoogle: Done in https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/191

PR #178, opened 2018-10-18
Adding JLBP-12: Make level of support and API stability clear

garrettjonesgoogle:


library-best-practices/JLBP-12.md
```diff
@@ -0,0 +1,8 @@
+[JLBP-12] Make level of support and level of API stability clear
```
elharo: level of support and API stability

library-best-practices/JLBP-12.md
```diff
@@ -0,0 +1,8 @@
...
+----------------------------------------------------------------
+
+- Surface stability should be clearly described.
```
elharo: Acrtve voice: Clearly state the stability clients can expect from the API surface. For example,

* This API is stable and suitable for production use, aside from packages marked internal. 
* This API is mostly stable except for the com.foo.canonical package which is changing rapidly and should not be relied on.
* This API is beta and will not change unless a bug or design flaw is uncovered.
* This API is in development. Anything may change at any time. Stable products should not depend on it.

library-best-practices/JLBP-12.md
```diff
@@ -0,0 +1,8 @@
...
+- Feedback mechanisms should be outlined (for example, file an issue).
```
elharo: Provide and identify a feedback mechanism. For example,

* File an issue at https://github.com/...
* Send email to dev@fooproject.example.com

library-best-practices/JLBP-12.md
```diff
@@ -0,0 +1,8 @@
...
+- Clearly describe when a library is abandoned.
```
elharo: This one is probably worth a JLBP of its own. &lt;cough>appengine-gcs-client&lt;/cough>

library-best-practices/JLBP-12.md
```diff
@@ -0,0 +1,8 @@
...
+- Clearly label 0.x libraries as unstable (as in, they can experience
```
elharo: seems redundant with above and JLBP on versioning

library-best-practices/JLBP-12.md
garrettjonesgoogle: I couldn't decide on this one - you're the tie breaker :-)

library-best-practices/JLBP-12.md
garrettjonesgoogle: I will include a variant of this.

garrettjonesgoogle:
Feedback addressed, PTAL

library-best-practices/JLBP-12.md
```diff
@@ -1,8 +1,16 @@
...
 
-- Surface stability should be clearly described.
-- Feedback mechanisms should be outlined (for example, file an issue).
-- Clearly describe when a library is abandoned.
-- Clearly label 0.x libraries as unstable (as in, they can experience
-  breakages and are not recommended for production use).
+- Clearly describe the lifecycle status (and corresponding level of expected support) of the library. Examples include:
+  - Incubating (no support expected)
+  - Stable (SLOs in place to handle issues)
+  - Maintenance only (only bugs will be addressed)
```
elharo: Maintenance only --> Maintenance

library-best-practices/JLBP-12.md
```diff
@@ -1,8 +1,16 @@
...
+  - Deprecated (no work is expected, use a replacement instead)
+  - Abandoned (no work is expected, no clear replacement)
```
elharo: These two don't fully match our internal instructions. See go/gcp-github  and https://github.com/googlearchive/

I also feel like there's one more relevant page I'm having trouble digging up ATM.

garrettjonesgoogle: I'm trying to keep the JLBPs generic enough to be used by non-Google libraries. Ideally, if there is a non-Google library we want to depend on, we should be able to check that library against these best practices. I'm currently thinking we might want to create a separate set of "GJLBPs" (Google Java Library Best Practices) that refer to Google-specific things, such as the requirement to sign binaries with Google's signature. 

I will merge these two into a single item so that it is less specific. The key point of this best practice is that library owners should make it clear to a human that they shouldn't expect additional work; it doesn't have to follow a rigid classification to comply with this best practice. For Google libraries, we might want to be more rigid, but I think that's separate.

garrettjonesgoogle:
More feedback addressed, PTAL

library-best-practices/JLBP-12.md
```diff
@@ -0,0 +1,20 @@
+[JLBP-12] Make level of support and API stability clear
+-------------------------------------------------------
+
+- Clearly describe the lifecycle status (and corresponding level of expected
```
elharo: delete "expected"

describe --> document

library-best-practices/JLBP-12.md
```diff
@@ -0,0 +1,20 @@
...
+  support) of the library. Examples include:
+  - Incubating (no support expected)
+  - Stable (SLOs in place to handle issues)
+  - Maintenance (only bugs will be addressed)
+  - Deprecated / Abandoned / Inactive (no work is expected)
+- Clearly state the API stability clients can expect from the library
```
elharo: Does the difference between API stability and behavioral stability matter for this JLBP? It seems libraries should document both of these together, for the most part.

library-best-practices/JLBP-12.md
garrettjonesgoogle: Ok.

library-best-practices/JLBP-12.md
garrettjonesgoogle: Yeah I think library owners should be clear about both. I will add that.

garrettjonesgoogle:
Feedback round 3 addressed, PTAL

PR #180, opened 2018-10-18
Creating glossary.md

garrettjonesgoogle:


library-best-practices/glossary.md
```diff
@@ -0,0 +1,84 @@
+Java Dependency Glossary
+------------------------
+
+### Types of conflicts and compatibility
+
+- **Linkage conflict**: The signature of a non-private method or class in a
```
elharo: It occurs to me we can define linkage conflicts without reference to compile time. I need to think about this more, but something like:

A class file references another method whose signature is not present in the classpath. Linkage conflicts typically manifest because the signature of a non-private method or class has changed in an incompatible way between compile time and runtime.

library-best-practices/glossary.md
```diff
@@ -0,0 +1,84 @@
...
+  dependency has changed in an incompatible way between the version supplied at
+  compile time and the version invoked at runtime. For example, a public method
+  may be removed from a class or a class may be made final. Linkage conflicts
+  detected at runtime manifest as ReflectiveOperationException,
```
elharo: quotes on ReflectiveOperationException

library-best-practices/glossary.md
```diff
@@ -0,0 +1,84 @@
...
+  - Sub-type: **Reflective linkage conflict**: A linkage conflict caused by a
+    reflective reference.
+- **Behavior conflict**: The class's implementation has changed in a way that
+  can break clients at runtime although all signatures remain compatible. For
+  example, if a method that has been returning mutable lists begins returning
+  immutable lists without updating its signature, dependents that mutate the
+  list will fail, possibly far away from the original call site. By contrast, a
+  change in return type from List to ImmutableList would be a linkage conflict.
+  - Opposite: **Behavior-compatible**.
+- **Version disagreement**: This is a class of failure reported by build systems
```
elharo: Can we add some blank lines in between paragraphs for clarity or will that mess up the formatting?

library-best-practices/glossary.md
```diff
@@ -0,0 +1,84 @@
...
+  with strict version checking on. The main example is dependency convergence
+  failures when using Maven version enforcer, which checks that all branches of
```
elharo: Maven enforcer plugin (it's just not for versions)

library-best-practices/glossary.md
```diff
@@ -0,0 +1,84 @@
...
+  a dependency tree agree on the version of every dependency. A failure of
+  dependency convergence doesn't guarantee a linkage or behavior conflict, and
+  the passing of a dependency convergence check doesn't guarantee there will be
```
elharo: If dependencies converge, static linkage errors will be detected at compile time. 

Reflective errors and behavior conflicts at this point indicate inadequate test coverage.

library-best-practices/glossary.md
```diff
@@ -0,0 +1,84 @@
...
+  no linkage or behavior conflicts. Consequently, version disagreement is a poor
+  proxy for real problems (linkage and behavior conflicts).
+  - Dependency convergence will fail even if the dependency tree only disagrees
```
elharo: I'm not sure the glossary should depend so heavily on the detailed rules of one plugin in one build system.

library-best-practices/glossary.md
```diff
@@ -0,0 +1,84 @@
...
+    or behavioral conflict.
+  - It is possible for dependency convergence to pass when a version of a
+    library is chosen at build time which is incompatible with another library
+    in the tree. This is possible by forcing Maven to pick the "wrong" version
+    using something like dependency management or dependency exclusion, which
+    silences the Maven enforcer.
+  - Opposite: **Version agreement**.
+- **Version alignment**: Said of the dependency tree of a Maven module. This
+  means that for any dependency of a module in that module's dependency tree,
+  all major Java build systems will select the same version of that dependency.
```
elharo: Again, the definition should be a function of the Java classes, not the build tools.

library-best-practices/glossary.md
```diff
@@ -0,0 +1,84 @@
...
+  - Linkage compatibility is the desired result from version alignment, but not
+    an inherent characteristic of the definition.
+
+#### Conflict relationships
+
+- A particular conflict cannot be both a linkage conflict and behavior conflict
+  at the same time (they are mutually exclusive).
+- A combination of jars at runtime can have any number of linkage conflicts and
+  behavior conflicts.
+- Version disagreement is a report by a tool, not an actual conflict. You can
```
elharo: "version disagreement failure" --> "version disagreement"

library-best-practices/glossary.md
```diff
@@ -0,0 +1,84 @@
...
+  have a version disagreement failure without either linkage or behavior
+  conflicts (a false positive report), or either linkage or behavior conflicts
+  without a version disagreement failure (a false negative report).
+
+### States of compatibility
+
+- **Linkage-compatible** (said of a particular version of A and a particular
+  version of B): When these versions are used together, there are no linkage
+  conflicts.
```
elharo: between A and B.

There can still be issues with C

library-best-practices/glossary.md
garrettjonesgoogle: I think the compile-time part helps people form a mental model in a concrete way, and I think it's worth keeping for the definition of "linkage conflict". Also, the conflicts are not limited to the absence of signatures; one case noted is the switch of a class to final, such that derived classes are not possible any more.

In addition to "linkage conflict," I think it's worth defining a more general (superordinate) idea of "linkage error," which would both linkage conflicts and completely missing references (which can be missing dependencies). WDYT?

library-best-practices/glossary.md
garrettjonesgoogle: Oops, will fix

library-best-practices/glossary.md
garrettjonesgoogle: I think the formatting will be unaffected because it is an unordered list.

Another option is for you to get used to list items sitting next to each other :-)

library-best-practices/glossary.md
garrettjonesgoogle: Will fix.

library-best-practices/glossary.md
garrettjonesgoogle: The passing of a dependency convergence check != dependencies converging. As noted elsewhere, dependency convergence checks will succeed when dependencyManagement is used, even if the tree still references multiple versions. Also, if exclusions are used so that the tree only references a single version per dependency and the dependency convergence check passes, the library trying to reference the excluded dependency can try to pull in a more recent version and use new features, and then there will be static linkage errors because the dependency tree is using an older version. In the theoretical scenario that the entire tree is compiled with identical dependencies and those are the same as the versions used in the final dependency tree, *only then* will you be guaranteed that you will have no static linkage errors at runtime. However, the text doesn't reference that idea, it only references the attempted check for convergence.

All this to say - the statement in the text still stands as-is.

library-best-practices/glossary.md
garrettjonesgoogle: If I remember correctly, Maven dependency convergence enforcement was originally going to form a huge part of our strategy; it was going to be the basis for our dashboard, and it was going to be part of our best practices. Now that things have changed and it is no longer part of our approach, we need to delineate it so that we can say that it is not a goal. I'm not sure how else to do that if we don't actually reference this particular plugin in this particular build system.

library-best-practices/glossary.md
garrettjonesgoogle: But the entire point of this concept relates to how build systems will form the classpath based on the dependency tree. If we take the build systems out of it, we have nothing.

library-best-practices/glossary.md
garrettjonesgoogle: Ok I'll add that.

library-best-practices/glossary.md
garrettjonesgoogle: Sure.

library-best-practices/glossary.md
```diff
@@ -0,0 +1,87 @@
...
+    aren't made within major versions), the upper bound of the versions in the
+    tree should be selected.
+    - This should fix all cases of linkage conflicts caused by purely additive
+      functionality.
+  - To achieve version alignment when multiple versions of a transitive
+    dependency are present, a direct dependency on the transitive dependency
+    will need to be added, in order to support Maven's dependency resolution
+    logic (which selects the version with the fewest hops on the dependency
+    tree).
+  - Linkage compatibility is the desired result from version alignment, but not
```
elharo: version alignment does not guarantee linkage compatibility because linkage errors occur between different artifacts. I suggest removing this bullet point.

library-best-practices/glossary.md
garrettjonesgoogle: Hmm that's a good general point. Should I just remove the definition of "version agreement" altogether?

library-best-practices/glossary.md
garrettjonesgoogle: We will use it when I create a JLBP for BOMs. Should I wait to add it until then?

garrettjonesgoogle: We will use it when I create a JLBP for BOMs. Should I wait to add it until then?

library-best-practices/glossary.md
garrettjonesgoogle: This is true, but if we remove this note, then people will wonder what the purpose is. It's like eating healthy: it won't guarantee that you will live longer with a higher quality of life, but that is the desired result of doing it.

library-best-practices/glossary.md
garrettjonesgoogle: https://docs.oracle.com/javase/specs/jls/se8/jls8.pdf

Looks like I have some light reading to do... 788 pages of it (for Java 8).

garrettjonesgoogle: I will add the non-signature stuff that needs to be mentioned, but I'm not sure what you have in mind for discussing byte code.

library-best-practices/glossary.md
elharo: Probably We may end up defining it there instead. Hard to tell until we get there.

garrettjonesgoogle: Well the pertinent question is whether I should remove it from this PR in the meantime.

elharo: I suggest removing it for now. We can always add it later when it's clearer.

garrettjonesgoogle:
Removed version alignment. I'll take care of other issues once you respond to the corresponding comment threads.

library-best-practices/glossary.md
```diff
@@ -0,0 +1,68 @@
...
+    code reference (non-reflective).
+  - Sub-type: **Reflective linkage conflict**: A linkage conflict caused by a
+    reflective reference.
+
+- **Behavior conflict**: The class's implementation has changed in a way that
+  can break clients at runtime although all signatures remain compatible. For
+  example, if a method that has been returning mutable lists begins returning
+  immutable lists without updating its signature, dependents that mutate the
+  list will fail, possibly far away from the original call site. By contrast, a
+  change in return type from List to ImmutableList would be a linkage conflict.
```
elharo: Is this a linkage conflict? Generally a return type can be made more specific without besaking existing code. Maybe flip so "change in return type from ImmutableList to List"

garrettjonesgoogle: Yes you're right, I will flip that.

garrettjonesgoogle: Wait, I didn't read carefully enough. I realized that this example is odd; it actually can break both directions. Part of the issue is that `ImmutableList` is not really substitutable for `List` even though it is derived from it.

- `List` -> `ImmutableList`: Behavior conflict, because code that used to modify the list will now throw exceptions.
- `ImmutableList` -> `List`: Static linkage conflict, because references to methods present on ImmutableList but absent from List will not be found.

We need to think of a clearer example that doesn't involve behavior conflicts.

elharo: I don't know. Both cases are covered (with the fix) and thus this example seems OK to me. But if you don't like it, maybe an example that doesn't change the signature at all? For instance a method that used to closes an input stream it was passed, no longer does that. Or a method that used to return non-negative numbers can now return negative numbers, or something like that?

library-best-practices/glossary.md
elharo: sounds like a good idea. We can always add it back in if we need it.

library-best-practices/glossary.md
garrettjonesgoogle: Oy, I also didn't read the larger context of this. You're right that it is actually a good example, because it demonstrates both things. I will keep it, and fix the linkage conflict part.

garrettjonesgoogle:
More feedback addressed, PTAL

PR #186, opened 2018-10-22
Adding JLBP-14: Do not use version ranges

garrettjonesgoogle:


library-best-practices/JLBP-14.md
```diff
@@ -0,0 +1,11 @@
+[JLBP-14] Use explicit versions, not version ranges
+---------------------------------------------------
+
+- Neither Maven nor Gradle have have a mechanism to locally lock versions like
+  package managers in other languages. If the pom.xml of your library specifies
+  a version range instead of an explicit version, it can pull in snapshot
```
elharo: I'm not sure it will get a snapshot version in this case.

library-best-practices/JLBP-14.md
```diff
@@ -0,0 +1,11 @@
+[JLBP-14] Use explicit versions, not version ranges
+---------------------------------------------------
+
+- Neither Maven nor Gradle have have a mechanism to locally lock versions like
```
elharo: Is the first sentence needed?

library-best-practices/JLBP-14.md
garrettjonesgoogle: Ok I wasn't sure myself - I think this statement was written by someone else and rode along to my PR. I'll just remove the reference to snapshots.

library-best-practices/JLBP-14.md
garrettjonesgoogle: No it's honestly not needed, since the JLBPs are focused on Java library authors. I will remove it.

garrettjonesgoogle:
Feedback addressed, PTAL

library-best-practices/JLBP-14.md
```diff
@@ -0,0 +1,12 @@
+[JLBP-14] Use explicit versions, not version ranges
+---------------------------------------------------
+
+- If the pom.xml of your library specifies a version range instead of an
+  explicit version, builds at different points in time can get different
+  dependency versions that could break your library unexpectedly.
+  - Version ranges can even pull in incomplete releases that will your
+    library's build (example:
+    https://github.com/GoogleCloudPlatform/appengine-gcs-client/issues/71).
+  - Thus, always use an explicit version.
```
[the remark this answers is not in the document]
garrettjonesgoogle: You make a good point. I'll change references to "single" version.

library-best-practices/JLBP-14.md
```diff
@@ -0,0 +1,12 @@
...
+- Single-element version ranges ("hard requirements") have a much different
```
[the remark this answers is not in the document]
garrettjonesgoogle: The "standard" way of specifying a version using a single version number is a "soft" requirement (it allows maven to select a higher version depending on its resolution rules). The "hard" requirement disallows anything but the exact version.

https://maven.apache.org/pom.html#Dependency_Version_Requirement_Specification

- 1.0: "Soft" requirement on 1.0 (just a recommendation, if it matches all other ranges for the dependency)
- [1.0]: "Hard" requirement on 1.0

library-best-practices/JLBP-14.md
```diff
@@ -0,0 +1,12 @@
+[JLBP-14] Use explicit versions, not version ranges
+---------------------------------------------------
+
+- If the pom.xml of your library specifies a version range instead of an
+  explicit version, builds at different points in time can get different
+  dependency versions that could break your library unexpectedly.
+  - Version ranges can even pull in incomplete releases that will your
```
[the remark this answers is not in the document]
garrettjonesgoogle: oops, will fix

library-best-practices/JLBP-14.md
```diff
@@ -0,0 +1,12 @@
+[JLBP-14] Use explicit versions, not version ranges
+---------------------------------------------------
+
+- If the pom.xml of your library specifies a version range instead of an
+  explicit version, builds at different points in time can get different
```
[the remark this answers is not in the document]
garrettjonesgoogle: will fix

garrettjonesgoogle:
Feedback addressed, PTAL

library-best-practices/JLBP-14.md
```diff
@@ -0,0 +1,16 @@
+[JLBP-14] Do not use version ranges
+-----------------------------------
+
+- If the pom.xml of your library specifies a version range instead of a
+  single version for any particular dependency, builds at different points
+  in time can get different versions of that dependency (as your dependency
```
elharo: perhaps, can get --> see

library-best-practices/JLBP-14.md
```diff
@@ -0,0 +1,16 @@
...
+  releases new versions), which could break your library unexpectedly.
```
elharo: could --> can

library-best-practices/JLBP-14.md
```diff
@@ -0,0 +1,16 @@
...
+  - Version ranges can even pull in incomplete releases that break your
+    library's build (example:
```
elharo: Random thought: did this qualify as a static linkage error or it a different kind of problem?

library-best-practices/JLBP-14.md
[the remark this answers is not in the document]
elharo: Weird. I never heard of anything like that before. It doesn't seem to explain what a hard or soft requirement is. We should probably double check with maven and make sure that isn't some legacy detail that isn't true any more.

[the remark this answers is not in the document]
elharo: https://books.sonatype.com/mvnref-book/reference/pom-relationships-sect-project-dependencies.html#pom-relationships-sect-version-ranges

This appears to be very uncommonly used. I wonder if we should recommend against hard requirements?

library-best-practices/JLBP-14.md
garrettjonesgoogle: I think the "can" part is still necessary; if a dependency never actually updates, you won't see different versions. I will s/get/see/ though.

library-best-practices/JLBP-14.md
garrettjonesgoogle: Ok

library-best-practices/JLBP-14.md
garrettjonesgoogle: I would say this is a different type of problem; the dependency tree that was specified couldn't be fetched.

library-best-practices/JLBP-14.md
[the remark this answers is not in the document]
garrettjonesgoogle: grpc-java is currently using it. I don't think it's a bad thing; it is a way to enforce something like the fact that grpc-auth 1.15.0 must be used with grpc-core 1.15.0.

PR #196, opened 2018-10-30
add a pre-1.0 report

### LINKED ISSUE #170: Pre 1.0 Dependencies
JLBP-8 suggests moving libraries to 1.0. We should generate a list of dependencies that are not 1.0 and work on moving them to 1.0. In many cases I suspect teams are unnecessarily resisting finalizing their APIs.

Arguably we shouldn't even publish info about pre 1.0 libraries on cloud.google.com and developers.google.com.

elharo:
fix #170 
wip

### DIFF
--- a/dashboard/src/main/resources/templates/dashboard.ftl
+++ b/dashboard/src/main/resources/templates/dashboard.ftl
@@ -76,8 +76,8 @@
     <#assign x = 0>
     <ul id="unstable">
       <#list latestArtifacts as artifact, version>
-        <#assign x++>
         <#if version[0] == '0'>
+          <#assign x++>
           <li>${artifact}:${version}</li>
         </#if>
       </#list>
--- a/dashboard/src/main/resources/templates/dashboard.ftl
+++ b/dashboard/src/main/resources/templates/dashboard.ftl
@@ -73,17 +73,17 @@
       Otherwise replace the dependency with something else.
     </p> 
     
-    <#assign x = 0>
+    <#assign unstableCount = 0>
     <ul id="unstable">
       <#list latestArtifacts as artifact, version>
         <#if version[0] == '0'>
-          <#assign x++>
+          <#assign unstableCount++>
           <li>${artifact}:${version}</li>
         </#if>
       </#list>
     </ul>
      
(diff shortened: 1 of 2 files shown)


elharo:
@garrettjonesgoogle I think the test failure is flaky. Working on that in a searate issue.

dashboard/src/main/resources/templates/dashboard.ftl
```diff
@@ -49,18 +49,39 @@
...
+      Otherwise replace the dependency with something else.
+    </p> 
+    
+    <ul id="unstable">
+      <#list latestArtifacts as artifact, version>
+        <#if version[0] == '0'>
+          <li>${artifact}:${version}</li>
+        </#if>
+      </#list>
+    </ul>
```
garrettjonesgoogle: It appears that this will just be empty if there are no 0.x libraries; do we want to print a "yay, you have no unstable deps" message in that case?

elharo: Ideally, though this doesn't seem likely to happen for some time, and putting logic like this in FreeMarker is way more convoluted than it should be.

garrettjonesgoogle: If we don't resolve it though, it's one of those things that will pop up later and it will confuse us until we remember what happened.

elharo: done

dashboard/src/main/resources/templates/dashboard.ftl
```diff
@@ -73,14 +73,21 @@
       Otherwise replace the dependency with something else.
     </p> 
     
+    <#assign x = 0>
     <ul id="unstable">
       <#list latestArtifacts as artifact, version>
+        <#assign x++>
```
garrettjonesgoogle: Why is this `x++` not inside the `if` condition?

Would it be easier to pass in a boolean to the template so there is less logic in it?

elharo: This is  a local variable. Passing it in from java code couple the two more tightly together.

PR #206, opened 2018-11-02
Dependency in dashboard shown as nested ul tags

### LINKED ISSUE #202: Tree control for Dependencies
Turn the Dependencies section on the individual component pages into a collapsing tree widget (expanded by default) so it's easier to navigate.

suztomo:
Fixes #202


suztomo:
Now it passes Circle CI, and adjusted the margin of the elements:
screen shot 2018-11-02 at 12 04 13

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -16,6 +16,9 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
+import com.google.cloud.tools.opensource.dependencies.DependencyPath;
```
elharo: please sort the imports

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -206,7 +209,10 @@ private static ArtifactResults generateArtifactReport(Configuration configuratio
       templateData.put("upperBoundFailures", upperBoundFailures);
       templateData.put("globalUpperBoundFailures", globalUpperBoundFailures);
       templateData.put("lastUpdated", LocalDateTime.now());
-      templateData.put("dependencyTree", dependencyTree);
+      // In `dependencyTree`, the root node is represented by empty DependencyPath
+      // Casting avoids Freemarker's error on `AbstractListMultimap.get` in CircleCI
+      templateData.put("dependencyTree", (ArrayListMultimap) dependencyTree);
+      templateData.put("dependencyTreeRoot", new DependencyPath());
```
elharo: I don't see why you need this here.

dashboard/src/main/resources/css/dashboard.css
```diff
@@ -10,6 +10,11 @@
    background-color: gray;
    font-weight: bold;
  }
+
+ p.DEPENDENCY_TREE_NODE {
```
elharo: p, not li?

dashboard/src/main/resources/templates/component.ftl
```diff
@@ -110,8 +110,28 @@
     </#if>
 
     <h2>Dependencies</h2>
-    <pre class="dependency-tree">${dependencyTree}</pre>
-    
+
+    <#macro formatDependencyNode tree currentNode parent>
+      <#-- tree: ListMultimap<DependencyPath, DependencyPath>, currentNode: DependencyPath -->
```
elharo: Ideally we shouldn't need to pass in currentNode from Java. Can we just create it here instead?

dashboard/src/main/resources/templates/component.ftl
```diff
@@ -110,8 +110,28 @@
...
+      <#if currentNode.size() gt 0>
+        <#-- root node does not have any leaf -->
+        <#if parent.size() gt 0>
+          <#assign label = 'parent: ' + parent.getLeaf() />
+        <#else>
+          <#assign label = 'root' />
+        </#if>
+        <p class="DEPENDENCY_TREE_NODE" title="${label}">${currentNode.getLeaf()}</p>
```
elharo: I don't think you need the p here. You can just use the li element.

dashboard/src/main/resources/templates/component.ftl
suztomo: FreeMarkers' ObjectConstructor worked.
https://freemarker.apache.org/docs/api/freemarker/template/utility/ObjectConstructor.html

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Sorted.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: removed

dashboard/src/main/resources/css/dashboard.css
suztomo: We need an HTML tag to hold the label (title) to show the mouse-hover. Adding title to li would make the scope of the li too big.

dashboard/src/main/resources/templates/component.ftl
suztomo: We need an HTML tag to hold the label (title) to show the mouse-hover. Adding title to li would make the scope of the li too big.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -187,26 +191,29 @@ private static ArtifactResults generateArtifactReport(Configuration configuratio
...
+      templateData.put("objectConstructor", new ObjectConstructor());
       templateData.put("groupId", artifact.getGroupId());
       templateData.put("artifactId", artifact.getArtifactId());
       templateData.put("version", artifact.getVersion());
       templateData.put("updates", convergenceIssues);
       templateData.put("upperBoundFailures", upperBoundFailures);
       templateData.put("globalUpperBoundFailures", globalUpperBoundFailures);
       templateData.put("lastUpdated", LocalDateTime.now());
-      templateData.put("dependencyTree", dependencyTree);
+      // In `dependencyTree`, the root node is represented by empty DependencyPath
```
elharo: comment no longer applies

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -187,26 +191,29 @@ private static ArtifactResults generateArtifactReport(Configuration configuratio
...
           collectLatestVersions(globalDependencies), transitiveDependencies);
 
-      String dependencyTree =
-          DependencyTreeFormatter.formatDependencyPaths(completeDependencies.list());
+      ListMultimap<DependencyPath, DependencyPath> dependencyTree =
+          DependencyTreeFormatter.buildDependencyPathTree(completeDependencies.list());
       Template report = configuration.getTemplate("/templates/component.ftl");
 
       Map<String, Object> templateData = new HashMap<>();
+      templateData.put("objectConstructor", new ObjectConstructor());
```
elharo: I'm surprised we need this. I guess it's a limitation of FreeMarker.

dashboard/src/main/resources/templates/component.ftl
```diff
@@ -110,8 +110,30 @@
...
```
elharo: I may have messed this up when I wrote some of this. It's weird that a node has a leaf. A node should either be or not be a leaf.

dashboard/src/main/resources/templates/component.ftl
```diff
@@ -110,8 +110,30 @@
...
+      <ul>
+        <#list tree.get(currentNode) as childNode>
+          <li class="DEPENDENCY_TREE_NODE">
+            <@formatDependencyNode tree childNode currentNode />
+          </li>
+        </#list>
+      </ul>
+    </#macro>
+    <@formatDependencyNode dependencyTree
+        objectConstructor('com.google.cloud.tools.opensource.dependencies.DependencyPath')
```
elharo: If we passed nulls here and wrote logic to check for that, would it avoid the need to use objectConstructor?

suztomo: `tree.get(null)` would cause problem as the root node is represented by empty DependencyPath.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: yes, it's freemarker's limitation.

dashboard/src/main/resources/templates/component.ftl
suztomo: Every DependencyPath instance has a leaf. (the middle node is stored in `path`)

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatter.java
```diff
@@ -105,7 +105,7 @@ private static void formatDependencyPathTree(
    * @param dependencyPaths dependency path instances without assuming any order
    * @return ListMultiMap representing a Maven dependency tree
```
elharo: If we stick with this representation for now, we should document how the tree is represented as a list multimap.

suztomo: Added description in Javadoc.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatter.java
```diff
@@ -98,14 +98,15 @@ private static void formatDependencyPathTree(
   }
 
   /**
-   * Builds ListMultiMap that represents a Maven dependency tree of parent-children relationship.
-   * The root node is represented as an empty {@link DependencyPath} and its children are
-   * the values for the root node.
+   * Builds ListMultimap that represents a Maven dependency tree of parent-children relationship.
```
elharo: Yes, but how? Now that this is a public method this needs to be described in sufficient detail for someone to understand the representation; that is, how a multimap turns into a tree, or how a tree is encoded in a multimap. They are not the same data structure, so how are we using this here? That's what needs to be clearer.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatter.java
```diff
@@ -98,14 +98,15 @@ private static void formatDependencyPathTree(
...
    *
    * @param dependencyPaths dependency path instances without assuming any order
-   * @return ListMultiMap representing a Maven dependency tree
+   * @return ListMultimap representing a Maven dependency tree of parent-children relationship. The
+   *     {@link DependencyPath} representing the root Maven artifact is available by querying the
+   *     tree with an empty {@link DependencyPath} instance. For a {@code node} in the returned
+   *     {@code tree}, its children are retrieved by {@code tree.get(node)}.
    */
-  private static ListMultimap<DependencyPath, DependencyPath> buildDependencyPathTree(
+  public static ListMultimap<DependencyPath, DependencyPath> buildDependencyPathTree(
```
elharo: We need a ListMultimap for order, but do we need to expose this in the API? Should this method simply return a Multimap instead?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatter.java
```diff
@@ -98,14 +98,15 @@ private static void formatDependencyPathTree(
...
-   * Builds ListMultiMap that represents a Maven dependency tree of parent-children relationship.
-   * The root node is represented as an empty {@link DependencyPath} and its children are
-   * the values for the root node.
+   * Builds ListMultimap that represents a Maven dependency tree of parent-children relationship.
    *
    * @param dependencyPaths dependency path instances without assuming any order
-   * @return ListMultiMap representing a Maven dependency tree
+   * @return ListMultimap representing a Maven dependency tree of parent-children relationship. The
+   *     {@link DependencyPath} representing the root Maven artifact is available by querying the
+   *     tree with an empty {@link DependencyPath} instance. For a {@code node} in the returned
```
elharo: Since a ListMultimap's values are ordered, why isn't the root simply the first value in the map? That would be easier to retrieve and seem more obvious.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatter.java
```diff
@@ -98,14 +98,15 @@ private static void formatDependencyPathTree(
...
   /**
-   * Builds ListMultiMap that represents a Maven dependency tree of parent-children relationship.
-   * The root node is represented as an empty {@link DependencyPath} and its children are
-   * the values for the root node.
+   * Builds ListMultimap that represents a Maven dependency tree of parent-children relationship.
    *
    * @param dependencyPaths dependency path instances without assuming any order
-   * @return ListMultiMap representing a Maven dependency tree
+   * @return ListMultimap representing a Maven dependency tree of parent-children relationship. The
+   *     {@link DependencyPath} representing the root Maven artifact is available by querying the
```
elharo: Multimap keys are nullable. Can we use null as a key here instead? I'm not sure but it's worth trying.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatter.java
elharo: To be more specific, I think you need to say something like:

Each node in the tree is a key in the multimap. The multiple values of that key are the children of that node.

suztomo: That's good explanation. Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatter.java
suztomo: First element in values of LinkedListMultimap worked good.

(Null as key would work too in the map in Java. Freemarker is not good at handling null though.)

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatter.java
```diff
@@ -98,19 +98,23 @@ private static void formatDependencyPathTree(
   }
 
   /**
-   * Builds ListMultiMap that represents a Maven dependency tree of parent-children relationship.
-   * The root node is represented as an empty {@link DependencyPath} and its children are
-   * the values for the root node.
+   * Builds ListMultimap that represents a Maven dependency tree of parent-children relationship.
+   * Each node in the tree has a corresponding key in the ListMultimap and the children of the node
```
elharo: Break into two sentences:

Each node in the tree has a corresponding key in the ListMultimap.
The value associated with that key is a list of the children of the node.

suztomo: Split the sentence.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -206,7 +211,9 @@ private static ArtifactResults generateArtifactReport(Configuration configuratio
       templateData.put("upperBoundFailures", upperBoundFailures);
       templateData.put("globalUpperBoundFailures", globalUpperBoundFailures);
       templateData.put("lastUpdated", LocalDateTime.now());
-      templateData.put("dependencyTree", dependencyTree);
+      // Explicit casting avoids Freemarker's error on `AbstractListMultimap.get` in CircleCI
```
netdpb: Can you explain this a bit more? How does Freemarker know anything about this cast?

netdpb:
Can you send me the whole stacktrace from that? It looks very strange to me.

suztomo:
@netdpb Email sent.

PR #211, opened 2018-11-02
Adding JLBP-18: Only shade dependencies as a last resort

garrettjonesgoogle:


library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,49 @@
+[JLBP-18] Use shading sparingly and carefully
```
elharo: Use shading --> shade

library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,49 @@
...
+---------------------------------------------
+
+The top answer to the question at
+https://softwareengineering.stackexchange.com/questions/297276/what-is-a-shaded-java-dependency
```
elharo: I would prefer this to be more self-contsined.

elharo:
@paul-hammant may have an opinion about this one.

library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,49 @@
...
+does a good job of explaining what shading is. In short, it is a process which
+relocates dependencies so that they don't clash with conflicting versions of the
+same dependencies in a dependency tree, then copies them all into a single jar
+with your own code.
+
+There are a number of problems with shading:
+
+- It can cause considerable size bloat, and if it is applied at multiple layers,
```
elharo: It can cause considerable bloat. If it is applied at multiple layers,

library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,49 @@
...
+  it can have a massive accumulating effect.
+- Users cannot upgrade dependencies shaded by other libraries (which means
+  security fixes to a transitive dependency need to wait for the shading library
+  to also roll in the security fix).
+- Service files can easily be messed up. They need special effort to be merged
```
elharo: I don't know what a service file is in this context.

library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,49 @@
...
+  because there is one per jar, always at the same path.
+- Shading doesn't relocate JNI code.
+- Classes loaded by reflection either can't be shaded automatically, or the
+  shader will replace all strings matching the path in the bytecode, leading to
```
elharo: will replace --> must replace

library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,49 @@
...
+  corruption of some constants.
+- It is easy to accidentally not relocate classes or other files, resulting in
+  an artifact that overlaps classes and files with the original dependency
+  (creating the situation described in [JLBP-5](JLBP-5.md) ).
+
+For these reasons, shading should be used sparingly, especially for libraries
+consumed by other libraries (because of the snowball effect).
+
+One area where shading has been put to good use is in frameworks, which need to
```
elharo: don't most frameworks in this situation use multiple classloaders?

library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,49 @@
...
+allow for user-space code to choose their own dependencies that might not be
+compatible with the framework's dependencies.
+
+If you do shade, you will need to have intimate knowledge of how your
```
elharo: will need to have --> need

library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,49 @@
...
+dependencies work, and you will probably need to read the source code. Make sure
```
elharo: delete probably

library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,49 @@
...
+you do all of the following:
+
+- Add a test to make sure that for all classes and other files that are copied
```
elharo: Add a test to make sure that all classes and other files copied
  into your jar from your dependencies are relocated

library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,49 @@
...
+  into your jar from your dependencies, they are all relocated.
+  - Other files include things like log4j.properties files.
+- For any transitive dependencies that are not relocated, promote them to direct
```
elharo: Promote transitive dependencies that are not relocated  to direct dependencies.

library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,49 @@
...
+  dependencies.
+- Make sure that no dependencies that appear on your own library surface are
```
elharo: Make sure no dependencies

library-best-practices/JLBP-18.md
garrettjonesgoogle: I think that might be a little terse for people not familiar with shading. I'd prefer to keep the existing form.

library-best-practices/JLBP-18.md
garrettjonesgoogle: That sounds reasonable, I'll make it self-contained.

elharo: OK, I'd still rewrite this though. "Sparingly" is a bit stilted and old-fashioned. "Carefully" is OK but when would we not recommend being careful. Can we be more specific here? Maybe even "Shade dependencies only as a last resort" or "Use shading only if multiple conflicting versions of a dependency are required"

library-best-practices/JLBP-18.md
garrettjonesgoogle: I think I'll go with "Only shade dependencies as a last resort". I think I want to avoid the command word at the beginning since we don't want anyone to fixate on the "shade" part that comes first and then think they need to do that (if they don't read carefully).

library-best-practices/JLBP-18.md
garrettjonesgoogle: Ok.

library-best-practices/JLBP-18.md
garrettjonesgoogle: Ok.

library-best-practices/JLBP-18.md
garrettjonesgoogle: ok.

library-best-practices/JLBP-18.md
garrettjonesgoogle: Ok.

library-best-practices/JLBP-18.md
garrettjonesgoogle: Ok.

library-best-practices/JLBP-18.md
garrettjonesgoogle: Ok.

library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,50 @@
+[JLBP-18] Only shade dependencies as a last resort
+--------------------------------------------------
+
+Shading is a process where selected classes (generally from dependencies)
+from the classpath of a library are relocated to different Java packages and
```
elharo: "the classpath of a library" --> a library's dependencies

since libraries technically don't have classpaths. Only runtime environments have classpaths.

library-best-practices/JLBP-18.md
```diff
@@ -0,0 +1,50 @@
...
+  (creating the situation described in [JLBP-5](JLBP-5.md) ).
+
+For these reasons, shading should be used sparingly, especially for libraries
+consumed by other libraries (because of the snowball effect).
+
+One area where shading has been put to good use is in frameworks, which need to
+allow for user-space code to choose their own dependencies that might not be
+compatible with the framework's dependencies.
+
+If you do shade, you need intimate knowledge of how your dependencies work, and
```
elharo: delete "do"

library-best-practices/JLBP-18.md
garrettjonesgoogle: Done

library-best-practices/JLBP-18.md
garrettjonesgoogle: Done

garrettjonesgoogle:
More feedback addressed, PTAL

PR #214, opened 2018-11-05
Readme for Static Linkage Checker

### LINKED ISSUE #212: Expand definition of static linkage errors
During the development of the static linkage checker, we uncovered a lot of detail about what exactly qualifies as a static linkage error and why or why not. This includes

* Provided, optional, and system dependencies. 
* How reachability determines whether a particular issue is or is not an error. 

Probably some other tings I'm not thinking of right now.

We need to capture all this knowledge in the repo, perhaps in the API docs,  the glossary, and/or a separate more detail doc about linkage errors.

suztomo:
I added concept of "class, method, and field reference", "entry point classes", "reachability", "dangling reference", and "static linkage error". They were useful when we discussed the concept of Static Linkage Checker.

Although this README introduces "static linkage error" specific to the tool, I tried to make it compatible with "static linkage conflict" (it's subclass of static linkage error).

Fixes #212

### DIFF
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -92,7 +92,7 @@ There are three use cases for Static Linkage Checker:
 - **Reachability** is the attribute of static linkage errors to indicate whether a linkage
   error caused by a reference is _reachable_ from the _entry point classes_ of the class usage
   graph; in other words, when a static linkage error is _reachable_, there exists a path of
-  references in the graph between one of entry point classes and the reference causing linkage
+  references in the graph from one of entry point classes to the reference causing linkage
   error.
   The path helps to diagnose why static linkage errors are introduced to the dependencies.
 
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -1,7 +1,7 @@
 Static Linkage Checker
 ======================
 
-Static Linkage Checker is a tool that to find [static linkage errors](
+Static Linkage Checker is a tool that finds [static linkage errors](
 ../library-best-practices/glossary.md) on a classpath and reports the errors to the console.
 It scans the class files in a classpath provided as input for references
 to other classes. Each reference is verified to find linkage conflicts
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -1,7 +1,8 @@
 # Static Linkage Checker
 
 Static Linkage Checker is a tool that finds [static linkage errors](
-../library-best-practices/glossary.md) on a classpath and reports the errors to the console.
+../library-best-practices/glossary.md#types-of-conflicts-and-compatibility)
+on a classpath and reports the errors to the console.
 It scans the class files in a classpath for references to other classes and
 reports any reference that cannot be satisfied in the classpath.
 It can report all such missing references or only those that are reachable from
@@ -21,17 +22,14 @@ There are two use cases for Static Linkage Checker:
 
 ### Approach
 
-1. The tool takes a classpath as required input, through either a list of jar and
-  class files in filesystems, a list of Maven coordinates, or a Maven BOM.
-  A Maven BOM is converted to a list of Maven coordinates, and a list of Maven
-  coordinates is resolved to a list of jar files that consists of the artifacts
-  and their dependencies.
+1. The tool takes a classpath as required input.
+
   The classpath is called as _linkage classpath_ on which the tool operates
   to find linkage errors and is separated from the runtime classpath of the tool itself.
 
 2. The tool extracts all _references_ from the all class files in the classpath.
 
-3. The tool records references cannot be satisfied in the classpath as
+3. The tool records references that cannot be satisfied in the classpath as
   static linkage errors.
   
 4. Optionally, the user can specify a subset of the classpath as _entry points_.
@@ -39,6 +37,16 @@ There are two use cases for Static Linkage Checker:
   from the classes in the entry points.
 
 5. At the end, the tool outputs a report on the linkage errors.
+
+### Input
+
+The tool takes a classpath through either a list of class and
+jar files in filesystems, a list of Maven coordinates, or a Maven BOM.
+
+A Maven BOM specified as a Maven coordinate is converted to a list of Maven coordinates.
+A list of Maven coordinates is resolved to a list of jar files
+that consists of the artifacts and their dependencies.
+The list of jar files forms a classpath, on which the tool operate.
 
 ### Output
 
@@ -52,9 +60,9 @@ source class and the destination class of the reference, and has one of the thre
 
   - _Missing field type_: a field reference has a static linkage conflict.
      
-### Class Usage Graph and Reachability
+### Class Reference Graph and Reachability
 
-In order to provide a diagnosis on the output report, the tool builds _class usage graphs_,
+In order to provide a diagnosis on the output report, the tool builds _class reference graphs_,
 and annotates linkage errors with _reachability_ from _entry point classes_.
 The tool allows users to choose the scope of entry point classes:
 
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -1,9 +1,8 @@
 Static Linkage Checker
 ======================
 
-Static Linkage Checker is a tool that performs a [static linkage check](
-../library-best-practices/glossary.md#static-linkage-check) on a classpath
-and reports static linkage errors to the console.
+Static Linkage Checker is a tool that to find [static linkage errors](
+../library-best-practices/glossary.md) on a classpath and reports the errors to the console.
 It scans the class files in a classpath provided as input for references
 to other classes. Each reference is verified to find linkage conflicts
 in the classpath.
@@ -15,11 +14,11 @@ classes through class usage graphs.
 There are two use cases for Static Linkage Checker:
 
 -  **For organizations** that provide multiple libraries developed by different teams,
-  the tool helps to ensure that there are no static linkage conflicts among the libraries and their
+  the tool helps to ensure that there are no static linkage errors among the libraries and their
   dependencies.
 
-- **For library/application developers** the tool assesses the risk of static linkage
-  conflicts in their projects, and will help to avoid incompatible versions of libraries
+- **For library/application developers** the tool finds static linkage
+  errors in their projects, and will help to avoid incompatible versions of libraries
   in their dependencies.
 
 ### Approach
@@ -31,7 +30,7 @@ There are two use cases for Static Linkage Checker:
 2. The tool extracts all _references_ from the all class files in the classpath.
 
 3. The tool checks the linkage compatibility of the references through the classpath, and records
-  static linkage conflicts.
+  static linkage errors.
   
 4. The tool traverses a _class usage graph_ to annotate the linkage errors with _reachability_.
 
@@ -39,21 +38,15 @@ There are two use cases for Static Linkage Checker:
 
 ### Output
 
-The tool reports zero or more static linkage errors for the linkage classpath, annotated
+The tool reports static linkage errors for the input linkage classpath, annotated
 with _reachability_. Each of the static linkage errors contains the information on the
 source class and the destination class of the reference, and has one of the three types:
 
-  - _Missing class type_: the destination class of a
-    class reference does not exist in the classpath. This error
-    happens when a class is removed in a different version of a library,
-    or there is a dependency missed when constructing the classpath.
-    The reference that causes a missing class error is called a _dangling reference_.
+  - _Missing class type_: a class reference causes a static missing class error.
 
-  - _Missing method type_: a method reference has a static
-    linkage conflict.
+  - _Missing method type_: a method reference has a static linkage conflict.
 
-  - _Missing field type_: a field reference has a static
-     linkage conflict.
+  - _Missing field type_: a field reference has a static linkage conflict.
      
 ### Class Usage Graph and Reachability
 
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -1,38 +1,42 @@
-Static Linkage Checker
-======================
+# Static Linkage Checker
 
 Static Linkage Checker is a tool that finds [static linkage errors](
 ../library-best-practices/glossary.md) on a classpath and reports the errors to the console.
-It scans the class files in a classpath provided as input for references
-to other classes.
-It then verifies each of the references to find linkage errors in the classpath.
-The tool also provides analysis on reachability to the errors from entry point
-classes through class usage graphs.
+It scans the class files in a classpath for references to other classes and
+reports any reference that cannot be satisfied in the classpath.
+It can report all such missing references or only those that are reachable from
+a given set of entry point classes.
 
 ### Use Cases
 
 There are two use cases for Static Linkage Checker:
 
--  **For organizations** that provide multiple libraries developed by different teams,
-  the tool helps to ensure that there are no static linkage errors among the libraries and their
-  dependencies.
-
 - **For library/application developers** the tool finds static linkage
   errors in their projects, and will help to avoid incompatible versions of libraries
   in their dependencies.
 
+-  **For organizations** that provide multiple libraries developed by different teams,
+  the tool helps to ensure that users depending on the libraries will not see any
+  static linkage errors at runtime.
+
 ### Approach
 
-1. Given an input, the tool prepares a _linkage classpath_, from which the tool inspects Java
-  classes.
-  Note that the linkage classpath is different from the runtime classpath of the tool itself.
+1. The tool takes a classpath as required input, through either a list of jar and
+  class files in filesystems, a list of Maven coordinates, or a Maven BOM.
+  A Maven BOM is converted to a list of Maven coordinates, and a list of Maven
+  coordinates is resolved to a list of jar files that consists of the artifacts
+  and their dependencies.
+  The classpath is called as _linkage classpath_ on which the tool operates
+  to find linkage errors and is separated from the runtime classpath of the tool itself.
 
 2. The tool extracts all _references_ from the all class files in the classpath.
 
-3. The tool checks the linkage compatibility of the references through the classpath, and records
+3. The tool records references cannot be satisfied in the classpath as
   static linkage errors.
   
-4. The tool traverses a _class usage graph_ to annotate the linkage errors with _reachability_.
+4. Optionally, the user can specify a subset of the classpath as _entry points_.
+  In that case, the tool will list only those references that are reachable
+  from the classes in the entry points.
 
 5. At the end, the tool outputs a report on the linkage errors.
 
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -43,16 +43,16 @@ The tool reports zero or more static linkage errors for the linkage classpath, a
 with _reachability_. Each of the static linkage errors contains the information on the
 source class and the destination class of the reference, and has one of the three types:
 
-  - _Missing class type_ is for errors when where the destination class of a
+  - _Missing class type_: the destination class of a
     class reference does not exist in the classpath. This error
     happens when a class is removed in a different version of a library,
     or there is a dependency missed when constructing the classpath.
     The reference that causes a missing class error is called a _dangling reference_.
 
-  - _Missing method type_ is for errors when a method reference has a static
+  - _Missing method type_: a method reference has a static
     linkage conflict.
 
-  - _Missing field type_ is for errors when a field reference has a static
+  - _Missing field type_: a field reference has a static
      linkage conflict.
      
 ### Class Usage Graph and Reachability
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -1,9 +1,11 @@
 Static Linkage Checker
 ======================
 
-To solve linkage conflicts caused by dependencies, Static Linkage Checker identifies the linkage
-conflicts by reading the class files in jar files, and verifying availability of the implementation
-of classes, methods and fields referenced by the files.
+Static Linkage Checker identifies the
+[linkage conflicts](../library-best-practices/glossary.md#types-of-conflicts-and-compatibility)
+that would be caused by 3rd-party dependencies of Java projects at runtime. The tool scans
+the class files in dependencies, and verifies availability of the implementation of the classes,
+methods, and fields referenced by the files.
 
 ### Use Cases
 
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -2,7 +2,7 @@ Static Linkage Checker
 ======================
 
 Static Linkage Checker is the tool to perform [static linkage check](
-../library-best-practices/glossary.md#types-of-conflicts-and-compatibility#static_linkage_check).
+../library-best-practices/glossary.md#types-of-conflicts-and-compatibility#static-linkage-check).
 The tool identifies static linkage conflicts that would be caused by Java projects and their
 dependencies at runtime. Given a classpath provided from
 user (e.g., a Maven project and its dependency), it scans the class files in the classpath
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -1,27 +1,26 @@
 Static Linkage Checker
 ======================
 
-Static Linkage Checker is a tool that inspects classpaths for [static linkage check](
-../library-best-practices/glossary.md#static-linkage-check).
-The tool identifies static linkage errors that appear in Java projects at runtime.
+Static Linkage Checker is a tool that performs a [static linkage check](
+../library-best-practices/glossary.md#static-linkage-check) on a classpath
+and reports static linkage errors to the console.
 It scans the class files in a classpath provided as input for references
-to other classes. Then it verifies compatibility of the methods and fields
-in the referenced classes, with regard to their type signatures and accessors.
+to other classes. Each reference is verified to find linkage conflicts
+in the classpath.
+The tool also provides analysis on reachability to the errors from entry point
+classes through class usage graphs.
 
 ### Use Cases
 
-There are three use cases for Static Linkage Checker:
+There are two use cases for Static Linkage Checker:
 
 -  **For organizations** that provide multiple libraries developed by different teams,
   the tool helps to ensure that there are no static linkage conflicts among the libraries and their
-  dependencies when some of them are used at the same time.
+  dependencies.
 
-- **For library developers**, the tool ensures that the users of the library will not suffer
-  static linkage conflicts caused solely by the dependencies of the library.
-
-- (future plan) **For application developers** consuming libraries, the tool will
-  assess the risk of static linkage conflicts among the application's dependencies, and will help
-  to avoid incompatible versions of libraries.
+- **For library/application developers** the tool will assess the risk of static linkage
+  conflicts in their projects, and will help to avoid incompatible versions of libraries
+  in their dependencies.
 
 ### Approach
 
@@ -41,11 +40,24 @@ There are three use cases for Static Linkage Checker:
 ### Output
 
 The tool reports zero or more static linkage errors for the linkage classpath, annotated
-with _reachability_.
+with _reachability_. Each of the static linkage errors contains the information on the
+source class and the destination class of the reference, and has one of the three types:
 
+  - _Missing class type_ is for errors when where the destination class of a
+    class reference does not exist in the classpath. This error
+    happens when a class is removed in a different version of a library,
+    or there is a dependency missed when constructing the classpath.
+    The reference that causes a missing class error is called a _dangling reference_.
+
+  - _Missing method type_ is for errors when a method reference has a static
+    linkage conflict.
+
+  - _Missing field type_ is for errors when a field reference has a static
+     linkage conflict.
+     
 ### Class Usage Graph and Reachability
 
-In order to provide diagnosis on the output report, the tool builds _class usage graphs_,
+In order to provide a diagnosis on the output report, the tool builds _class usage graphs_,
 and annotates linkage errors with _reachability_ from _entry point classes_.
 The tool allows users to choose the scope of entry point classes:
 
@@ -53,10 +65,10 @@ The tool allows users to choose the scope of entry point classes:
     target project, it ensures that the current functionality used in the dependencies will not
     cause static linkage errors.
     The output may fail to report potential static linkage errors, which would be introduced
-    by starting to use a new class in one of the dependencies.
+    by starting to use a previously unreachable class in one of the dependencies.
 
   - **Direct dependencies of the target project**: when the scope of the entry point is the all
-    classes in direct dependencies of the target project, it ensures that functionality of the
+    classes in the direct dependencies of the target project, it ensures that functionality of the
     dependencies will not cause static linkage errors. The output may contain linkage errors for
-    unused classes from user's perspective.
+    unreachable classes from user's perspective.
 
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -18,7 +18,7 @@ There are two use cases for Static Linkage Checker:
   the tool helps to ensure that there are no static linkage conflicts among the libraries and their
   dependencies.
 
-- **For library/application developers** the tool will assess the risk of static linkage
+- **For library/application developers** the tool assesses the risk of static linkage
   conflicts in their projects, and will help to avoid incompatible versions of libraries
   in their dependencies.
 
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -1,0 +1,85 @@
+Static Linkage Checker
+======================
+
+To solve linkage conflicts caused by dependencies, Static Linkage Checker identifies the linkage
+conflicts by reading the class files in jar files, and verifying availability of the implementation
+of classes, methods and fields referenced by the files.
+
+### Use Cases
+
+There are three use cases for Static Linkage Checker:
+
+-  **For organizations** that provide multiple libraries (through BOM) developed by different teams,
+  the tool helps to ensure that there is no static linkage conflicts among the libraries and their
+  dependencies.
+
+- **For library developers**, the tool ensures that the users of the library will not suffer the
+  static linkage conflicts caused solely by the dependencies of the library.
+
+- (future plan) **For application developers** consuming 3rd-party libraries, the tool will
+  assess th risk of static linkage conflicts among the application's dependencies, and will help
+  to use appropriate versions of libraries to avoid runtime errors.
+
+### Approach
+
+1. Given an input, the tool prepares a _linkage classpath_, from which the tool inspects Java
+  classes.
+  
+
+2. The tool extracts all references from the class files in the classpath.
+
+3. The tool checks the linkage compatibility of the references through the classpath.
+  
+4. To annotate _reachability_ on the linkage errors, the tool traverses a _class usage graph_.
+
+5. Once the tool finishes traversal, then it outputs the report on linkage errors.
+
+
+### Class Usage Graph and Reachability
+
+In order to provide to better diagnosis on the output report, the tool builds _class usage
+graphs_ when running the analysis. The nodes (vertex) of the graph correspond to Java classes
+and the directed edges are references between classes. For example, when 'Class A' has reference
+to 'Class B' on calling a method X, the tool builds a graph holding an edge between the two nodes:
+
+    [Class A] --(method X of class B)-> [Class B]
+
+In this case, 'Class A' is called the _source class_ of the reference and 'Class B' is called
+the _destination class_.
+
+- **A reference** between two classes is either _class reference_, _method reference_
+  or _field reference_.
+
+  A _method reference_ indicates the source class uses the (static or non-static) method of
+  the destination class.
+
+  A _field reference_ indicates the source class accesses the (static or non-static) field of
+  the destination class.
+
+  A _class reference_ indicates the source class uses the destination class without specific field 
+  or method (e.g., inheritance).
+
+  A reference is called _linkage conflicting_ when there is a linkage conflict
+  (see [glossary.md](../library-best-practices/glossary.md)) caused by the reference.
+  When the destination class does not exist in the linkage classpath, the reference is called
+  _dangling reference_. When a reference does not have linkage error, then the reference is 
+  called _linkage compatible_,
+
+- **Reachability** is the attribute of static linkage errors to indicate whether a linkage
+  error caused by a reference is reachable from the _entry point classes_ of the class usage
+  graph; in other words, there exists a path of references in the graph between a entry point class
+  and the reference causing linkage error. The path helps to diagnose why static linkage
+  errors are introduced to the dependencies.
+
+- **Static Linkage Errors** are reported by the tool. A static linkage error is either
+  a _dangling reference_ or a _static linkage conflict_.
+
+  A _dangling reference_ is a reference where its destination class does not exist in the 
+  linkage classpath. This error happens when a class is removed in different version of a library or
+  there is a dependency missed when constructing the linkage classpath.
+  
+  A _static linkage conflict_ is described in [glossary.md](../library-best-practices/glossary.md).
+  In short, this error happens when the destination class's implementation is not compatible from
+  what it was available at the time of compilation of the source class, in a way that such 
+  incompatibility causes `NoSuchFieldException`, `MethodNotFoundException`, or `LinkageError` when
+  Java Virtual Machine tries to use the implementation.
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -17,7 +17,7 @@ There are three use cases for Static Linkage Checker:
   static linkage conflicts caused solely by the dependencies of the library.
 
 - (future plan) **For application developers** consuming 3rd-party libraries, the tool will
-  assess th risk of static linkage conflicts among the application's dependencies, and will help
+  assess the risk of static linkage conflicts among the application's dependencies, and will help
   to use appropriate versions of libraries to avoid runtime errors.
 
 ### Approach
@@ -26,7 +26,7 @@ There are three use cases for Static Linkage Checker:
   classes.
   
 
-2. The tool extracts all references from the class files in the classpath.
+2. The tool extracts all _references_ from the class files in the classpath.
 
 3. The tool checks the linkage compatibility of the references through the classpath.
   
@@ -35,17 +35,40 @@ There are three use cases for Static Linkage Checker:
 5. Once the tool finishes traversal, then it outputs the report on linkage errors.
 
 
+### Output
+
+- **Static Linkage Errors** are reported by the tool. A static linkage error is either
+  a _dangling reference_ or a _static linkage conflict_.
+
+  A _dangling reference_ is a reference where its destination class does not exist in the 
+  linkage classpath. This error happens when a class is removed in a different version of a library
+  , or there is a dependency missed when constructing the linkage classpath.
+  
+  A _static linkage conflict_ is described in [glossary.md](../library-best-practices/glossary.md).
+  In short, this error happens when the destination class's implementation is not compatible from
+  what it was available at the time of compilation of the source class, in a way that such 
+  incompatibility causes `NoSuchFieldException`, `MethodNotFoundException`, or `LinkageError` when
+  Java Virtual Machine tries to use the implementation.
+
+
 ### Class Usage Graph and Reachability
 
-In order to provide to better diagnosis on the output report, the tool builds _class usage
-graphs_ when running the analysis. The nodes (vertex) of the graph correspond to Java classes
-and the directed edges are references between classes. For example, when 'Class A' has reference
-to 'Class B' on calling a method X, the tool builds a graph holding an edge between the two nodes:
+- **Linkage Classpath** is a list of jar files from which the tool inspects Java classes when
+  running the tool against a project. A linkage classpath corresponds to the transitive dependencies
+  induced by the direct dependencies in the target project's pom.xml file. This includes the
+  dependencies marked as `provided` or `optional: true`.
+
+- **Class Usage Graph**
+
+  In order to provide better diagnosis on the output report, the tool builds _class usage
+  graphs_ when running the analysis. The nodes (vertex) of the graph correspond to Java classes
+  and the directed edges are references between classes. For example, when 'Class A' has reference
+  to 'Class B' on calling a method X, the tool builds a graph holding an edge between two nodes:
 
     [Class A] --(method X of class B)-> [Class B]
 
-In this case, 'Class A' is called the _source class_ of the reference and 'Class B' is called
-the _destination class_.
+  In this case, 'Class A' is called the _source class_ of the reference and 'Class B' is called
+  the _destination class_.
 
 - **A reference** between two classes is either _class reference_, _method reference_
   or _field reference_.
@@ -67,19 +90,19 @@ the _destination class_.
 
 - **Reachability** is the attribute of static linkage errors to indicate whether a linkage
   error caused by a reference is reachable from the _entry point classes_ of the class usage
-  graph; in other words, there exists a path of references in the graph between a entry point class
+  graph; in other words, there exists a path of references in the graph between an entry point class
   and the reference causing linkage error. The path helps to diagnose why static linkage
   errors are introduced to the dependencies.
 
-- **Static Linkage Errors** are reported by the tool. A static linkage error is either
-  a _dangling reference_ or a _static linkage conflict_.
+- **Entry Point Classes** are classes in the class usage graph and used to verify the reachability
+  of static linkage errors. Users have choices on the scope of entry point classes:
 
-  A _dangling reference_ is a reference where its destination class does not exist in the 
-  linkage classpath. This error happens when a class is removed in different version of a library or
-  there is a dependency missed when constructing the linkage classpath.
-  
-  A _static linkage conflict_ is described in [glossary.md](../library-best-practices/glossary.md).
-  In short, this error happens when the destination class's implementation is not compatible from
-  what it was available at the time of compilation of the source class, in a way that such 
-  incompatibility causes `NoSuchFieldException`, `MethodNotFoundException`, or `LinkageError` when
-  Java Virtual Machine tries to use the implementation.
+  - **Classes in the target project**: when entry point are only the classes in the target project,
+    it ensures the current functionality used in the dependencies will not cause static linkage 
+    errors. The output may fail to report potential static linkage errors, which would be introduced
+    by starting to use a new class in one of the dependencies.
+
+  - **Direct dependencies of the target project**: when entry point are all classes in direct
+    dependencies of the target project, it ensures that functionality of the dependencies will
+    not cause static linkage errors. The output may contain linkage error for unused classes from
+    user's perspective.
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -4,8 +4,8 @@ Static Linkage Checker
 Static Linkage Checker is a tool that finds [static linkage errors](
 ../library-best-practices/glossary.md) on a classpath and reports the errors to the console.
 It scans the class files in a classpath provided as input for references
-to other classes. Each reference is verified to find linkage conflicts
-in the classpath.
+to other classes.
+It then verifies each of the references to find linkage errors in the classpath.
 The tool also provides analysis on reachability to the errors from entry point
 classes through class usage graphs.
 
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -77,11 +77,10 @@ The tool allows users to choose the scope of entry point classes:
     by starting to use a previously unreachable class in one of the dependencies.
 
   - **With direct dependencies of the target project**: when the scope of the entry point is
-    the classes in the target project and the all
-    classes in the direct dependencies of the project, it ensures that functionality of the
-    dependencies will not cause static linkage errors. The output may contain linkage errors for
-    unreachable classes from user's perspective.
+    the classes in the target project and the all classes in the direct dependencies of the project,
+    it ensures that functionality of the dependencies will not cause static linkage errors.
+    The output may contain linkage errors for unreachable classes from user's perspective.
 
-  - **(All classes in the input class path)**: when reachability check is off, then
+  - **All classes in the input class path**: when reachability check is off, then
     all static linkage errors from all classes in the classpath, regardless of the reachability,
     are reported.
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -65,9 +65,9 @@ _missing class_, _missing method_, or _missing field_.
      
 ### Class Reference Graph and Reachability
 
-In order to provide a diagnosis on the output report, the tool builds _class reference graphs_,
+In order to provide a diagnosis in the output report, the tool builds a _class reference graph_,
 and annotates linkage errors with _reachability_ from _entry point classes_.
-Optionally the tool outputs a report only including _reachable_ static linkage errors.
+Optionally the tool outputs a report including only _reachable_ static linkage errors.
 The tool allows users to choose the scope of entry point classes:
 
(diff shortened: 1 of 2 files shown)


dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+  A _method reference_ indicates the source class uses the (static or non-static) method of
+  the destination class.
+
+  A _field reference_ indicates the source class accesses the (static or non-static) field of
+  the destination class.
+
+  A _class reference_ indicates the source class uses the destination class without specific field 
+  or method (e.g., inheritance).
+
+  A reference is called _linkage conflicting_ when there is a linkage conflict
```
garrettjonesgoogle: "linkage conflicting" feels grammatically odd to me.

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+  (see [glossary.md](../library-best-practices/glossary.md#types-of-conflicts-and-compatibility)
```
garrettjonesgoogle:
```suggestion
  (see [glossary.md](../library-best-practices/glossary.md#types-of-conflicts-and-compatibility) )
```

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+  caused by the reference.
+  When the destination class does not exist in the linkage classpath, the reference is called
+  _dangling reference_. When a reference does not have linkage error, then the reference is
```
garrettjonesgoogle:
```suggestion
  a _dangling reference_. When a reference does not have a linkage error, then the reference is 
```

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+  A reference is called _linkage conflicting_ when there is a linkage conflict
+  (see [glossary.md](../library-best-practices/glossary.md#types-of-conflicts-and-compatibility)
+  caused by the reference.
+  When the destination class does not exist in the linkage classpath, the reference is called
+  _dangling reference_. When a reference does not have linkage error, then the reference is 
+  called _linkage compatible_,
+
+- **Reachability** is the attribute of static linkage errors to indicate whether a linkage
+  error caused by a reference is _reachable_ from the _entry point classes_ of the class usage
+  graph; in other words, when a static linkage error is _reachable_, there exists a path of
```
garrettjonesgoogle:
```suggestion
  graph. In other words, when a static linkage error is _reachable_, there exists a path of
```

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+  references in the graph from one of entry point classes to the reference causing linkage
+  error.
+  The path helps to diagnose why static linkage errors are introduced to the dependencies.
+
+- **Entry Point Classes** are classes in the class usage graph and used to verify the reachability
```
garrettjonesgoogle:
```suggestion
- **Entry Point Classes** are classes in the class usage graph that are used to verify the reachability
```

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+  of static linkage errors. Users of the tool have choices on the scope of entry point classes:
+
+  - **Classes in the target project**: when the scope of entry point is only the classes in the
```
garrettjonesgoogle:
```suggestion
  - **Classes in the target project**: when the scope of the entry point is only the classes in the
```

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+    target project, it ensures the current functionality used in the dependencies will not cause
```
garrettjonesgoogle:
```suggestion
    target project, it ensures that the current functionality used in the dependencies will not cause
```

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+    static linkage errors.
+    The output may fail to report potential static linkage errors, which would be introduced
+    by starting to use a new class in one of the dependencies.
+
+  - **Direct dependencies of the target project**: when the scope of entry point is the all classes
+    in direct dependencies of the target project, it ensures that functionality of the dependencies
+    will not cause static linkage errors. The output may contain linkage error for unused classes
```
garrettjonesgoogle:
```suggestion
    will not cause static linkage errors. The output may contain linkage errors for unused classes
```

dependencies/README.md
suztomo: How about "A reference is called to have a linkage conflict" ?

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+- **A reference** between two classes is either _class reference_, _method reference_
+  or _field reference_.
+
+  A _method reference_ indicates the source class uses the (static or non-static) method of
+  the destination class.
+
+  A _field reference_ indicates the source class accesses the (static or non-static) field of
+  the destination class.
+
+  A _class reference_ indicates the source class uses the destination class without specific field
```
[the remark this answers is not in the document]
suztomo: Updated

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+  In this case, 'Class A' is called the _source class_ of the reference and 'Class B' is called
+  the _destination class_.
+
+- **A reference** between two classes is either _class reference_, _method reference_
+  or _field reference_.
+
+  A _method reference_ indicates the source class uses the (static or non-static) method of
+  the destination class.
+
+  A _field reference_ indicates the source class accesses the (static or non-static) field of
```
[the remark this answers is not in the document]
suztomo: Updated

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+    [Class A] --(method X of class B)-> [Class B]
+
+  In this case, 'Class A' is called the _source class_ of the reference and 'Class B' is called
+  the _destination class_.
+
+- **A reference** between two classes is either _class reference_, _method reference_
+  or _field reference_.
+
+  A _method reference_ indicates the source class uses the (static or non-static) method of
```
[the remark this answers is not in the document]
suztomo: Updated

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+  A _static linkage conflict_ is described in [glossary.md](../library-best-practices/glossary.md).
+  In short, this error happens when the destination class's implementation is not compatible from
+  what it was available at the time of compilation of the source class, in a way that such 
+  incompatibility causes `NoSuchFieldException`, `MethodNotFoundException`, or `LinkageError` when
+  Java Virtual Machine tries to use the implementation.
+
+
+### Class Usage Graph and Reachability
+
+- **Linkage Classpath** is a list of jar files from which the tool inspects Java classes when
```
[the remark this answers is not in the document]
suztomo: The term "linkage classpath" is introduced to distinguish the classpath of the tool (StaticLinkageChecker) itself and the classpath of the subject.

So without the mention of the tool, it's just a "classpath". I'd like to hear more from you.

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+4. The tool traverses a _class usage graph_ to annotate the linkage errors with _reachability_.
+
+5. At the end, the tool outputs the report on linkage errors.
+
+### Output
+
+- **Static Linkage Errors** are reported by the tool. A static linkage error is either
+  a _dangling reference_ or a _static linkage conflict_.
+
+  A _dangling reference_ is a reference where its destination class does not exist in the
```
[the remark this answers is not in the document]
suztomo: "Unsatisfied" sounds to include the case where type signature/accessor mismatch.

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+  a _dangling reference_ or a _static linkage conflict_.
+
+  A _dangling reference_ is a reference where its destination class does not exist in the 
+  linkage classpath. This error happens when a class is removed in a different version of a library
+  , or there is a dependency missed when constructing the linkage classpath.
+  
+  A _static linkage conflict_ is described in [glossary.md](../library-best-practices/glossary.md).
+  In short, this error happens when the destination class's implementation is not compatible from
+  what it was available at the time of compilation of the source class, in a way that such 
+  incompatibility causes `NoSuchFieldException`, `MethodNotFoundException`, or `LinkageError` when
```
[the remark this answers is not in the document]
suztomo: Yes, this document defines "static linkage error". I didn't find the term in "library-best-practices/glossary.md" though.

>  The messy part is that the tool can't really distinguish between the two causes of missing classes

That's right. I feel it's unnatural to call something missing as conflict. Let's continue discussion on this.

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+1. Given an input, the tool prepares a _linkage classpath_, from which the tool inspects Java
+  classes.
+
+2. The tool extracts all _references_ from the class files in the classpath.
+
+3. The tool checks the linkage compatibility of the references through the classpath.
+  
+4. The tool traverses a _class usage graph_ to annotate the linkage errors with _reachability_.
+
+5. At the end, the tool outputs the report on linkage errors.
```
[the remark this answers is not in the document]
suztomo: Fixed

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+There are three use cases for Static Linkage Checker:
+
+-  **For organizations** that provide multiple libraries (through BOM) developed by different teams,
+  the tool helps to ensure that there is no static linkage conflicts among the libraries and their
+  dependencies when some of them are used at the same time.
+
+- **For library developers**, the tool ensures that the users of the library will not suffer the
+  static linkage conflicts caused solely by the dependencies of the library.
+
+- (future plan) **For application developers** consuming 3rd-party libraries, the tool will
```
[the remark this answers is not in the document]
suztomo: Removed "3rd-party".

I placed "future plan" to indicate this use case is out of scope of 2018 Q4 target.

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Use Cases
+
+There are three use cases for Static Linkage Checker:
+
+-  **For organizations** that provide multiple libraries (through BOM) developed by different teams,
+  the tool helps to ensure that there is no static linkage conflicts among the libraries and their
+  dependencies when some of them are used at the same time.
+
+- **For library developers**, the tool ensures that the users of the library will not suffer the
```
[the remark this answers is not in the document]
suztomo: Fixed

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+[linkage conflicts](../library-best-practices/glossary.md#types-of-conflicts-and-compatibility)
+that would be caused by 3rd-party dependencies of Java projects at runtime. The tool scans
+the class files for references to other classes, and verifies the availability of the implementation
+through the classpath of the dependencies.
+
+### Use Cases
+
+There are three use cases for Static Linkage Checker:
+
+-  **For organizations** that provide multiple libraries (through BOM) developed by different teams,
```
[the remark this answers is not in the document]
suztomo: That is true. Removed the mention of BOM.

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
+Static Linkage Checker
+======================
+
+Static Linkage Checker identifies the
+[linkage conflicts](../library-best-practices/glossary.md#types-of-conflicts-and-compatibility)
+that would be caused by 3rd-party dependencies of Java projects at runtime. The tool scans
+the class files for references to other classes, and verifies the availability of the implementation
+through the classpath of the dependencies.
```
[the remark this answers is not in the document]
suztomo: How about this?

Given a classpath provided from
user (e.g., a Maven project and its dependency), the tool scans the class files in the classpath
for references to other classes, and verifies the availability of the implementation through
the classpath.

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+2. The tool extracts all _references_ from the class files in the classpath.
+
+3. The tool checks the linkage compatibility of the references through the classpath.
+  
+4. The tool traverses a _class usage graph_ to annotate the linkage errors with _reachability_.
+
+5. At the end, the tool outputs the report on linkage errors.
+
+### Output
```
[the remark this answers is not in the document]
elharo: I agree. I think we need to separate the definitions from specific tooling. If they don't sync up, at least one, possibly both, is somewhat incorrect.

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+-  **For organizations** that provide multiple libraries (through BOM) developed by different teams,
+  the tool helps to ensure that there is no static linkage conflicts among the libraries and their
+  dependencies when some of them are used at the same time.
+
+- **For library developers**, the tool ensures that the users of the library will not suffer the
+  static linkage conflicts caused solely by the dependencies of the library.
+
+- (future plan) **For application developers** consuming 3rd-party libraries, the tool will
+  assess the risk of static linkage conflicts among the application's dependencies, and will help
+  to use appropriate versions of libraries to avoid runtime errors.
```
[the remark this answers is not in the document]
suztomo: Updated.

dependencies/README.md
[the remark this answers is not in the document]
suztomo: Let me try to update them, and add to existing glossary.md.

dependencies/README.md
```diff
@@ -0,0 +1,112 @@
...
+that would be caused by 3rd-party dependencies of Java projects at runtime. The tool scans
+the class files for references to other classes, and verifies the availability of the implementation
+through the classpath of the dependencies.
+
+### Use Cases
+
+There are three use cases for Static Linkage Checker:
+
+-  **For organizations** that provide multiple libraries (through BOM) developed by different teams,
+  the tool helps to ensure that there is no static linkage conflicts among the libraries and their
```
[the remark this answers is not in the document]
suztomo: Updated.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,77 @@ Java Dependency Glossary
...
+  dependencies in the project's pom.xml file.
+
+- **Class usage graph**: a directed graph that represents the relationship between
+  classes in the linkage classpath. The nodes (vertex) of the graph correspond to
+  Java classes and the directed edges are references between classes.
+  For example, when 'Class A' has reference to 'Class B' on calling a method X,
+  the class usage graph holds an edge between two nodes:
+
+  ```
+  [Class A] --(method X of class B)-> [Class B]
```
elharo: There can be multiple paths from A to B under this definition.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,77 @@ Java Dependency Glossary
...
+  ```
+
+  In this case, 'Class A' is called the _source class_ of the reference and
+  'Class B' is called the _destination class_.
+
+- **A reference**: in the class usage graph, the relationship between two 
+  classes is either _class reference_, _method reference_ or _field reference_.
+
+  A _method reference_ indicates that the source class uses the (static or
```
elharo: uses --> invokes

library-best-practices/glossary.md
```diff
@@ -66,3 +66,77 @@ Java Dependency Glossary
...
+  non-static) method of the destination class.
+
+  A _field reference_ indicates that the source class accesses the (static or
```
elharo: the (static --> a (static

library-best-practices/glossary.md
```diff
@@ -66,3 +66,77 @@ Java Dependency Glossary
...
+  non-static) field of the destination class.
+
+  A _class reference_ indicates that the source class uses the destination
+  class without referencing a specific field or method (e.g., class inheritance).
+
+  A reference is called to have a _linkage conflict_
```
elharo: called --> said

suztomo:
@elharo @garrettjonesgoogle I moved terms to glossary.md. Static linkage check is explained as a (generic) _process_ in glossary, and Static Linkage Checker is the tool to perform the process.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,77 @@ Java Dependency Glossary
...
+  when there is a static linkage conflict caused by the reference.
```
elharo: This definition is circular. Maybe, "when the target of the reference is not found" but then that requires the target node to be a field or a method, not just a class. I'm not sure what to do here.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,77 @@ Java Dependency Glossary
...
+
+- **Reachability** is the attribute of static linkage errors to indicate
+  whether a linkage error caused by a reference is _reachable_ from the _entry
+  point classes_ of the class usage graph. In other words, when a static
```
elharo: entry point classes are separate from the class usage graph, not part of it.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,77 @@ Java Dependency Glossary
...
+  linkage error is _reachable_, there exists a path of references in the graph
+  from one of the entry point classes to the reference causing linkage error.
+  The path helps to diagnose why static linkage errors are introduced to the
```
elharo: why --> how

library-best-practices/glossary.md
```diff
@@ -66,3 +66,77 @@ Java Dependency Glossary
...
+  linkage classpath.
+
+- **Entry point Classes** are classes in the class usage graph that are used
```
elharo: " are classes in the class usage graph that are used" --> classes used

dependencies/README.md
```diff
@@ -0,0 +1,62 @@
+Static Linkage Checker
+======================
+
+Static Linkage Checker is the tool to perform [static linkage check](
+../library-best-practices/glossary.md#types-of-conflicts-and-compatibility#static-linkage-check).
+The tool identifies static linkage conflicts that would be caused by Java projects and their
+dependencies at runtime. Given a classpath provided from
+user (e.g., a Maven project and its dependency), it scans the class files in the classpath
+for references to other classes, and verifies the availability of the implementation through
```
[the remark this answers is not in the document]
suztomo: Updated.

dependencies/README.md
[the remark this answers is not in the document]
suztomo: The terms have been moved to glossary.md.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,74 @@ Java Dependency Glossary
...
   A and the version of B are linkage-compatible.
+
+
+### Static Linkage Check
+
+- **Static linkage check**: a process to identify the existence of static
+  linkage errors for a given classpath, by scanning Java classes and
+  verifying the availability in the classpath.
+
+- **Static linkage report**: the outcome of a static linkage check. It reports
```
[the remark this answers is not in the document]
suztomo: Removed.

library-best-practices/glossary.md
suztomo: Yes, multiple edges between two classes are normal. Clarified in the document.

Such _parallel edges_ are supported in Guava's [Network](https://github.com/google/guava/wiki/GraphsExplained#network)

library-best-practices/glossary.md
```diff
@@ -66,3 +66,77 @@ Java Dependency Glossary
...
+  - _Missing field type_ is the error when a field reference has a static
+     linkage conflict.
+
+- **Linkage classpath**: the classpath which a static linkage check
+  runs against. For a Maven project, a linkage classpath consists of the
+  project's class files and the transitive dependencies induced by the direct
+  dependencies in the project's pom.xml file.
+
+- **Class usage graph**: a directed graph that represents the relationship between
+  classes in the linkage classpath. The nodes (vertex) of the graph correspond to
```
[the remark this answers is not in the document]
suztomo: Updated

library-best-practices/glossary.md
```diff
@@ -66,3 +66,77 @@ Java Dependency Glossary
...
+    or there is a dependency missed when constructing the linkage classpath.
+    The reference that causes a missing class error is called a _dangling reference_.
+
+  - _Missing method type_ is the error when a method reference has a  static
+    linkage conflict.
+
+  - _Missing field type_ is the error when a field reference has a static
+     linkage conflict.
+
+- **Linkage classpath**: the classpath which a static linkage check
```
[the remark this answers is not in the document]
suztomo: Removed "linkage classpath" from glossary. Yes, it's only needed for the tool.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,77 @@ Java Dependency Glossary
...
+  runs against. For a Maven project, a linkage classpath consists of the
+  project's class files and the transitive dependencies induced by the direct
+  dependencies in the project's pom.xml file.
+
+- **Class usage graph**: a directed graph that represents the relationship between
```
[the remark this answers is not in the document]
suztomo: Yes, it's cyclic. Updated.

Added description:

> To avoid unnecessary graph traversal, self-loops (references between the same class) are omitted.

library-best-practices/glossary.md
suztomo: Updated

library-best-practices/glossary.md
suztomo: Updated

library-best-practices/glossary.md
suztomo: Updated

library-best-practices/glossary.md
suztomo: Removed the sentence. I couldn't find good sentence here. "Static linkage error" section explains the concept.

library-best-practices/glossary.md
suztomo: Removed from "of the class usage graph".

dependencies/README.md
```diff
@@ -0,0 +1,62 @@
+Static Linkage Checker
+======================
+
+Static Linkage Checker is the tool to perform [static linkage check](
+../library-best-practices/glossary.md#types-of-conflicts-and-compatibility#static-linkage-check).
+The tool identifies static linkage conflicts that would be caused by Java projects and their
+dependencies at runtime. Given a classpath provided from
+user (e.g., a Maven project and its dependency), it scans the class files in the classpath
```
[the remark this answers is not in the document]
suztomo: Updated.

dependencies/README.md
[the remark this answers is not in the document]
garrettjonesgoogle: That sounds good.

dependencies/README.md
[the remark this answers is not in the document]
garrettjonesgoogle: Sorry, I should have said it is redefining a static linkage *conflict*. 

I think "conflict" is still fine as a term in the context of two libraries wanting two versions of the same dependency, one of which is missing a class. They are "in conflict" about what the right answer is.

dependencies/README.md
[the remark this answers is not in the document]
garrettjonesgoogle: A couple things to say:

- After thinking about the term more carefully, I don't know that "linkage classpath" has very much sense to it.
- If you want to talk about the classpath that the tool is analyzing (to distinguish from the tool's classpath), I feel like "input classpath" or "subject classpath" would be clearer. The definition would be pretty simple: "A classpath analyzed by a tool, in contrast to the classpath of the tool itself." It doesn't need to reference Static Linkage Checker at all - this term could be used by other tools, and thus is appropriate for the glossary.
- There might also be value in having a term for a type of classpath that includes `provided` + `optional` dependencies, which doesn't need to make any mention of tools. Maybe "effective classpath" or "rendered classpath" or something else might work?

dependencies/README.md
garrettjonesgoogle: In that case, the sentence would become self-referential, and might as well be removed.

dependencies/README.md
```diff
@@ -0,0 +1,62 @@
+Static Linkage Checker
+======================
+
+Static Linkage Checker is a tool that inspects classpaths for [static linkage check](
```
garrettjonesgoogle: "Static Linkage Checker is a tool performs a [static linkage check](...) on a classpath."

dependencies/README.md
```diff
@@ -0,0 +1,62 @@
...
+../library-best-practices/glossary.md#static-linkage-check).
+The tool identifies static linkage errors that appear in Java projects at runtime.
```
garrettjonesgoogle: This makes it sound like the Static Linkage Checker executes the code. I feel like the information in this sentence is already implied by "static linkage check" and can be removed. Maybe you can add the following to the prior sentence: "... and reports static linkage errors to the console."

dependencies/README.md
```diff
@@ -0,0 +1,62 @@
...
+It scans the class files in a classpath provided as input for references
+to other classes. Then it verifies compatibility of the methods and fields
+in the referenced classes, with regard to their type signatures and accessors.
```
garrettjonesgoogle: I feel like this last sentence is repeating information from "static linkage check." I think it might be more relevant to mention entry point classes and reachability here (since "static linkage check" doesn't imply those).

dependencies/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Use Cases
+
+There are three use cases for Static Linkage Checker:
+
+-  **For organizations** that provide multiple libraries developed by different teams,
+  the tool helps to ensure that there are no static linkage conflicts among the libraries and their
+  dependencies when some of them are used at the same time.
```
garrettjonesgoogle: Remove "when some of them are used at the same time." - this is implied by "among the libraries and their dependencies".

dependencies/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+- **For library developers**, the tool ensures that the users of the library will not suffer
```
garrettjonesgoogle:
```suggestion
- **For library developers**, the tool ensures that users of the library will not suffer
```

dependencies/README.md
```diff
@@ -0,0 +1,62 @@
...
+  static linkage conflicts caused solely by the dependencies of the library.
+
+- (future plan) **For application developers** consuming libraries, the tool will
```
garrettjonesgoogle: > I placed "future plan" to indicate this use case is out of scope of 2018 Q4 target.

Honestly, completion of any part of the tool is out of scope of 2018 Q4 target.

As for usage by library developers and application developers, I feel like there is really no difference. We either support both or neither.

I feel like "future plan" makes use for calling out future functionality; it doesn't seem like tool documentation is a good place talk about future support for different classes of users.

dependencies/README.md
```diff
@@ -0,0 +1,62 @@
...
+5. At the end, the tool outputs a report on the linkage errors.
+
+### Output
+
+The tool reports zero or more static linkage errors for the linkage classpath, annotated
+with _reachability_.
+
+### Class Usage Graph and Reachability
+
+In order to provide diagnosis on the output report, the tool builds _class usage graphs_,
```
garrettjonesgoogle:
```suggestion
In order to provide a diagnosis on the output report, the tool builds _class usage graphs_,
```

dependencies/README.md
```diff
@@ -0,0 +1,62 @@
...
+and annotates linkage errors with _reachability_ from _entry point classes_.
+The tool allows users to choose the scope of entry point classes:
+
+  - **Classes in the target project**: when the scope of the entry point is only the classes in the
+    target project, it ensures that the current functionality used in the dependencies will not
+    cause static linkage errors.
+    The output may fail to report potential static linkage errors, which would be introduced
+    by starting to use a new class in one of the dependencies.
```
garrettjonesgoogle:
```suggestion
    by starting to use a previously unreachable class in one of the dependencies.
```

dependencies/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+  - **Direct dependencies of the target project**: when the scope of the entry point is the all
+    classes in direct dependencies of the target project, it ensures that functionality of the
```
garrettjonesgoogle:
```suggestion
    classes in the direct dependencies of the target project, it ensures that the functionality of the
```

dependencies/README.md
```diff
@@ -0,0 +1,62 @@
...
+    dependencies will not cause static linkage errors. The output may contain linkage errors for
+    unused classes from user's perspective.
```
garrettjonesgoogle:
```suggestion
    unreachable classes from user's perspective.
```

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
 - **Linkage-matchable version** (said of a particular version of A in relation
   to all versions of B): There exists some version of B such that the version of
   A and the version of B are linkage-compatible.
+
+
+### Static Linkage Check
```
garrettjonesgoogle:
```suggestion
### Static Linkage Checking
```

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
...
+  ```
+
+  In this case, 'Class A' is called the _source class_ of the reference and
+  'Class B' is called the _destination class_.
+
+  In general, there can be multiple (parallel) edges between two nodes when
+  a class is calling two or more methods and fields on another class.
+  Self-loops (references between the same class) are possible and
+  common, however, it is safe to omit such references during static linkage checks,
+  because the self-loop references are ensured to be linkage compatible by Java compiler.
```
garrettjonesgoogle:
```suggestion
  because the self-loop references are ensured to be linkage compatible by the Java compiler.
```

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
...
+
+- **A reference**: in the class usage graph, the relationship between two 
+  classes is either _class reference_, _method reference_ or _field reference_.
```
garrettjonesgoogle:
```suggestion
  classes is either a _class reference_, _method reference_ or _field reference_.
```

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
...
+  A _method reference_ indicates that the source class invokes a (static or
+  non-static) method of the destination class.
+
+  A _field reference_ indicates that the source class accesses a (static or
+  non-static) field of the destination class.
+
+  A _class reference_ indicates that the source class uses the destination
+  class without referencing a specific field or method (e.g., class inheritance).
+
+- **Static linkage error**: a reported error for a reference as the result of
```
garrettjonesgoogle: I feel like this notion belongs in the "Types of conflicts and compatibility" section, although probably with some modification so that it fits. In particular, we need need a general notion of an error beyond just conflicts (the missing case), without any reference to running a tool.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
...
+    or there is a dependency missed when constructing the classpath.
+    The reference that causes a missing class error is called a _dangling reference_.
+
+  - _Missing method type_ is for errors when a method reference has a static
+    linkage conflict.
+
+  - _Missing field type_ is for errors when a field reference has a static
+     linkage conflict.
+
+- **Reachability** is the attribute of static linkage errors to indicate
```
garrettjonesgoogle: I think we are tying our terms too much to the "static" type of linkage errors; for several  of these terms, there is no inherent reason that they should be limited to static linkage errors - they could apply to dynamic linkage errors too. 

Also, I feel like "reachability" should be defined as an attribute of references, not linkage errors.

library-best-practices/glossary.md
elharo: I feel like we should be defining states here; that is, static linkage and static linkage errors; not processes such as a check

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
 - **Linkage-matchable version** (said of a particular version of A in relation
   to all versions of B): There exists some version of B such that the version of
   A and the version of B are linkage-compatible.
+
+
+### Static Linkage Check
+
+- **Static linkage check**: a process to identify the existence of _static
```
elharo: I don't think this definition is needed since it's not fundamental.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
...
+
+  ```
+  [Class A] --(method X of class B)-> [Class B]
+  ```
+
+  In this case, 'Class A' is called the _source class_ of the reference and
+  'Class B' is called the _destination class_.
+
+  In general, there can be multiple (parallel) edges between two nodes when
+  a class is calling two or more methods and fields on another class.
```
elharo: a class references two or more methods and fields of another class.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
...
+  Self-loops (references between the same class) are possible and
+  common, however, it is safe to omit such references during static linkage checks,
```
elharo: common. However,

library-best-practices/glossary.md
elharo: active voice: the Java compiler ensures self-references are linkage compatible.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
...
+  non-static) method of the destination class.
+
+  A _field reference_ indicates that the source class accesses a (static or
+  non-static) field of the destination class.
+
+  A _class reference_ indicates that the source class uses the destination
+  class without referencing a specific field or method (e.g., class inheritance).
+
+- **Static linkage error**: a reported error for a reference as the result of
+  a static linkage check.
```
elharo: Again, I think we can define the problem without reference to how or whether it's checked.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
...
+  A static linkage error contains the information on the source class and
```
elharo: This sounds like you're describing s class (contains information) rather than a state or better yet, a mathematical property of a certain list.

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
...
+  the destination class of the reference. Each static linkage error has
+  one of the three types:
+
+  - _Missing class type_ is for errors when where the destination class of a
```
elharo: delete "is for errors when where "

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
...
+    class reference does not exist in the classpath. This error
+    happens when a class is removed in a different version of a library,
+    or there is a dependency missed when constructing the classpath.
```
elharo: there is a dependency missed --> a dependency is missing

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
...
+    The reference that causes a missing class error is called a _dangling reference_.
+
+  - _Missing method type_ is for errors when a method reference has a static
```
elharo: delete "is for errors when"

library-best-practices/glossary.md
```diff
@@ -66,3 +66,72 @@ Java Dependency Glossary
...
+    linkage conflict.
+
+  - _Missing field type_ is for errors when a field reference has a static
```
elharo: delete "is for errors when"

library-best-practices/glossary.md
elharo: agreed

library-best-practices/glossary.md
```diff
@@ -3,14 +3,20 @@ Java Dependency Glossary
 
 ### Types of conflicts and compatibility
 
+- **Linkage error**: an error when a Java class in a classpath has reference to
```
garrettjonesgoogle:
```suggestion
- **Linkage error**: an error when a Java class in a classpath has a reference to
```

library-best-practices/glossary.md
```diff
@@ -3,14 +3,20 @@ Java Dependency Glossary
...
+  another class file for a class, field or method, and the reference can not be
```
garrettjonesgoogle:
```suggestion
  another class file for a class, field, or method, and the reference cannot be
```

library-best-practices/glossary.md
```diff
@@ -3,14 +3,20 @@ Java Dependency Glossary
...
+  satisfied with the available classes in the classpath.
+  A linkage error is caused by a reference to missing classes or a linkage conflict.
+  Linkage errors detected at runtime manifest as `ReflectiveOperationException`,
+  `NoClassDefFoundError`, `NoSuchFieldException`, `MethodNotFoundException`,
+  `LinkageError`, or other related exceptions.
+  - Sub-type: **Linkage conflict**
+
 - **Linkage conflict**: The signature, return type, modifiers, or throws
```
garrettjonesgoogle: I think we should indent "linkage conflict" since it is a subtype of "linkage error", and also add a definition for a term referring to missing things.

library-best-practices/glossary.md
```diff
@@ -66,3 +72,61 @@ Java Dependency Glossary
...
+  In this case, 'Class A' is called the _source class_ of the reference and
+  'Class B' is called the _destination class_.
+
+  In general, there can be multiple (parallel) edges between two nodes when
+  a class references two or more methods and fields of another class.
+  Self-loops (references between the same class) are possible and
+  common. However, it is safe to omit such references during static linkage checks,
+  because the Java compiler ensures that the self-loop references are linkage compatible.
+
+- **A reference**: in the class usage graph, the relationship between two
```
garrettjonesgoogle:
```suggestion
- **Reference**: in the class usage graph, the relationship between two 
```

library-best-practices/glossary.md
```diff
@@ -66,3 +72,61 @@ Java Dependency Glossary
 - **Linkage-matchable version** (said of a particular version of A in relation
   to all versions of B): There exists some version of B such that the version of
   A and the version of B are linkage-compatible.
+
+
+### Static Linkage Checking
+
+- **Static linkage**: a reference from a class to another class that is
```
garrettjonesgoogle: I think we should just treat "static" as a modifier for other terms, and not define "static linkage" at all.

library-best-practices/glossary.md
```diff
@@ -66,3 +72,61 @@ Java Dependency Glossary
...
+  statically found in the Java class file of the former class.
+
+- **Static linkage error**: a linkage error for a static linkage.
```
garrettjonesgoogle: Just define this as a sub-type of "linkage error".

library-best-practices/glossary.md
```diff
@@ -66,3 +72,61 @@ Java Dependency Glossary
...
+
+- **Static linkage check**: a process to identify the existence of _static
+  linkage conflicts_ for a given classpath, by scanning Java classes and
+  verifying the availability in the classpath.
+
+- **Class usage graph**: a possibly cyclic directed graph of references between classes
+  in the classpath. The nodes (vertices) of the graph correspond to
+  Java classes and the directed edges are references between classes.
+  For example, when 'Class A' has reference to 'Class B' on calling a method X,
```
garrettjonesgoogle:
```suggestion
  For example, when 'Class A' has invokes method X on 'Class B',
```

library-best-practices/glossary.md
```diff
@@ -66,3 +72,61 @@ Java Dependency Glossary
...
+  the class usage graph holds an edge between two nodes:
```
garrettjonesgoogle:
```suggestion
  the class usage graph holds an edge between the two nodes:
```

library-best-practices/glossary.md
garrettjonesgoogle: This sentence doesn't actually define a reference, it only says that there are sub-types. Could we give it a self-contained, one sentence definition?

library-best-practices/glossary.md
```diff
@@ -66,3 +72,61 @@ Java Dependency Glossary
...
+  a class references two or more methods and fields of another class.
+  Self-loops (references between the same class) are possible and
+  common. However, it is safe to omit such references during static linkage checks,
+  because the Java compiler ensures that the self-loop references are linkage compatible.
+
+- **A reference**: in the class usage graph, the relationship between two 
+  classes is either a _class reference_, _method reference_ or _field reference_.
+
+  A _method reference_ indicates that the source class invokes a (static or
+  non-static) method of the destination class.
```
garrettjonesgoogle: Maybe move the "static or non-static" to the end so we don't have a parenthetical in the middle. e.g. "the source class invokes a method of the destination class, which can be static or non-static."

library-best-practices/glossary.md
```diff
@@ -66,3 +72,61 @@ Java Dependency Glossary
...
+  A _class reference_ indicates that the source class uses the destination
+  class without referencing a specific field or method (e.g., class inheritance).
+
+
+- **Reachability** is the attribute of classes (nodes in the graph)
+  and references (edges in the graph) to indicate whether they are
+  _reachable_ from a class. For example, when a reference that causes
+  a linkage error is marked as _reachable_ from 'Class A', it means that
+  there exists a path of edges in the class usage graph from 'Class A'
+  to the reference causing linkage error.
```
garrettjonesgoogle:
```suggestion
  to the reference causing a linkage error.
```

library-best-practices/glossary.md
suztomo: Added. What do you think?

library-best-practices/glossary.md
suztomo: Removed "static or non-static"

library-best-practices/glossary.md
suztomo: Fixed

dependencies/README.md
```diff
@@ -0,0 +1,67 @@
+Static Linkage Checker
```
netdpb: Is this the best name for the tool? Should it be something more like CheckStaticLinks? check_static_links?

netdpb:
```suggestion
# Static Linkage Checker
```

dependencies/README.md
```diff
@@ -0,0 +1,67 @@
...
+======================
```
netdpb:
```suggestion
```

dependencies/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Static Linkage Checker is a tool that finds [static linkage errors](
+../library-best-practices/glossary.md) on a classpath and reports the errors to the console.
+It scans the class files in a classpath provided as input for references
```
netdpb:
```suggestion
It scans the class files in a classpath for references
```

dependencies/README.md
```diff
@@ -0,0 +1,67 @@
...
+to other classes.
+It then verifies each of the references to find linkage errors in the classpath.
```
netdpb:
```suggestion
and reports any references that cannot be found in the classpath.
```

dependencies/README.md
```diff
@@ -0,0 +1,67 @@
...
+The tool also provides analysis on reachability to the errors from entry point
```
netdpb:
```suggestion
It can report all such missing references or only those that are reachable from a given set of entry points.
```

dependencies/README.md
```diff
@@ -0,0 +1,67 @@
...
+classes through class usage graphs.
+
+### Use Cases
+
+There are two use cases for Static Linkage Checker:
+
+-  **For organizations** that provide multiple libraries developed by different teams,
+  the tool helps to ensure that there are no static linkage errors among the libraries and their
```
netdpb:
```suggestion
  the tool helps to ensure that users depending on the libraries will not see any linkage errors at runtime
```

dependencies/README.md
```diff
@@ -0,0 +1,67 @@
...
+  dependencies.
+
+- **For library/application developers** the tool finds static linkage
```
netdpb: I would put this one first. The other one is secondary; it's there only to make it less likely that a user using this tool will find errors that the library developer should have caught.

dependencies/README.md
```diff
@@ -0,0 +1,67 @@
...
+  errors in their projects, and will help to avoid incompatible versions of libraries
+  in their dependencies.
+
+### Approach
+
+1. Given an input, the tool prepares a _linkage classpath_, from which the tool inspects Java
```
netdpb: I would say that the required input is a class path. What do you think of this?

1.  The user specifies a class path on which to operate. This is separate from the class path of the tool itself.
2.  The tool lists all _static references_ from classes in the class path that cannot be found in the class path or in the JDK.
3.  Optionally, the user can specify a subset of the class path as "entry points"; in that case the tool will list only those missing references that are _reachable_ from the classes in the entry points.

I think the entry point subset should be specifiable as either a class path or a set of classes, but that's separate.

dependencies/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+  - _Missing field type_: a field reference has a static linkage conflict.
+     
+### Class Usage Graph and Reachability
+
+In order to provide a diagnosis on the output report, the tool builds _class usage graphs_,
+and annotates linkage errors with _reachability_ from _entry point classes_.
+The tool allows users to choose the scope of entry point classes:
+
+  - **Classes in the target project**: when the scope of the entry point is only the classes in the
```
netdpb: Is the input to this tool a project definition like a pom.xml file? I think it should not be. I think instead this tool should operate on pure class paths. That way we can write a Maven plugin that uses the tool, or a Gradle one, or a Bazel one.

library-best-practices/glossary.md
```diff
@@ -3,23 +3,38 @@ Java Dependency Glossary
...
-    compatibility are the same, a linkage conflict is when compilation would
-    fail if the libraries in the classpath were all built together from their
-    originating source code, or when reflection would fail.
-  - Opposite: **Linkage-compatible**.
-  - Sub-type: **Static linkage conflict**: A linkage conflict caused by a direct
-    code reference (non-reflective).
-  - Sub-type: **Reflective linkage conflict**: A linkage conflict caused by a
-    reflective reference.
+- **Linkage error**: an error when a Java class in a classpath references
+  another class (for the class itself, its field, or its method), and the reference
```
netdpb:
```suggestion
  another class (as a class literal, a field reference, or a method reference), and the reference
```

library-best-practices/glossary.md
```diff
@@ -3,23 +3,38 @@ Java Dependency Glossary
...
+  cannot be satisfied with the available classes in the classpath.
+  A linkage error is caused by a reference to missing classes or a linkage conflict.
```
netdpb: I think this line duplicates the "sub-types" below and can be deleted.

library-best-practices/glossary.md
```diff
@@ -3,23 +3,38 @@ Java Dependency Glossary
...
+  Linkage errors detected at runtime manifest as `ReflectiveOperationException`,
+  `NoClassDefFoundError`, `NoSuchFieldException`, `MethodNotFoundException`,
+  `LinkageError`, or other related exceptions.
+  - Sub-type: **Linkage conflict**
+  - Sub-type: **Missing class error**
+
+  - **Linkage conflict**: The signature, return type, modifiers, or throws
```
netdpb: I'm not sure this should be nested.

library-best-practices/glossary.md
```diff
@@ -3,23 +3,38 @@ Java Dependency Glossary
...
+    declaration of a non-private method or class in a dependency has changed in an
```
netdpb: "changed" is perhaps not right when referring to deleting a method.

Also, what about field references?

library-best-practices/glossary.md
```diff
@@ -3,23 +3,38 @@ Java Dependency Glossary
...
+      code reference (non-reflective).
+    - Sub-type: **Reflective linkage conflict**: A linkage conflict caused by a
+      reflective reference.
+
+  - **Missing class error**: an error when a class referenced does not exist
+    in the classpath. This error happens when a class is removed in a different
+    version of a library, or there is a dependency missed when constructing the classpath.
+
+  - **Static**: Said of a linkage error when the linkage error is caused by a
+    direct code reference (e.g., _static linkage error_ and _static linkage conflict_').
```
netdpb:
```suggestion
    direct code reference (e.g., _static linkage error_ and _static linkage conflict_).
```

library-best-practices/glossary.md
```diff
@@ -66,3 +81,55 @@ Java Dependency Glossary
 - **Linkage-matchable version** (said of a particular version of A in relation
   to all versions of B): There exists some version of B such that the version of
   A and the version of B are linkage-compatible.
+
+
+### Class usage graph
+
+- **Class usage graph**: a possibly cyclic directed graph of references between classes
```
netdpb: I'm not sure the terms related to the graph and reachability are general enough to go into the glossary. They're relevant to the "reachability" notion in the static linkage checker, so maybe should be defined there.

library-best-practices/glossary.md
```diff
@@ -66,3 +81,55 @@ Java Dependency Glossary
...
+  in the classpath. The nodes (vertices) of the graph correspond to
+  Java classes and the directed edges are references between classes.
+  For example, when 'Class A' invokes method X on 'Class B',
+  the class usage graph holds an edge between the two nodes:
+
+  ```
+  [Class A] --(method X of class B)-> [Class B]
+  ```
+
+  In this case, 'Class A' is called the _source class_ of the reference and
```
netdpb: Should we call these "using class" and "referenced class"?

library-best-practices/glossary.md
```diff
@@ -66,3 +81,55 @@ Java Dependency Glossary
...
+  'Class B' is called the _destination class_.
+
+  In general, there can be multiple (parallel) edges between two nodes when
+  a class references two or more methods and fields of another class.
+  Self-loops (references between the same class) are possible and
+  common. However, it is safe to omit such references when checking the reachability
```
netdpb: Again, here, we're leaking implementation considerations for the static linkage checker into a glossary.

library-best-practices/glossary.md
netdpb: And there we can maybe just simply say something like:

Only linkage errors that are reachable through some sequence of method calls, field accesses, and class literal references from the entry point set are reported.

(We will actually report more than that for now because we assume everything in a class is reached when anything is, but we could actually narrow it down to that set later, and that would be ideal.)

garrettjonesgoogle: I would prefer to keep these terms in the glossary in case we reference them outside of the static linkage checker (which I feel is pretty likely). Even if not, I think it's best for all terms to be defined in one place so that they are easier to find.

library-best-practices/glossary.md
```diff
@@ -3,29 +3,38 @@ Java Dependency Glossary
...
+    invoked at runtime. For example, a public method may be removed from a class
+    or a class may be made final.
+    - Or, another perspective: In cases where binary compatibility and source
+      compatibility are the same, a linkage conflict is when compilation would
+      fail if the libraries in the classpath were all built together from their
+      originating source code, or when reflection would fail.
+    - Opposite: **Linkage-compatible**.
+    - Sub-type: **Static linkage conflict**: A linkage conflict caused by a direct
+      code reference (non-reflective).
+    - Sub-type: **Reflective linkage conflict**: A linkage conflict caused by a
```
garrettjonesgoogle: Generally "dynamic" is the opposite of "static".

library-best-practices/glossary.md
```diff
@@ -101,32 +101,35 @@ Java Dependency Glossary
...
 
-- **A reference**: in the class usage graph, the relationship between two 
-  classes is either a _class reference_, _method reference_ or _field reference_.
-
-  A _method reference_ indicates that the source class invokes a (static or
-  non-static) method of the destination class.
-
-  A _field reference_ indicates that the source class accesses a (static or
-  non-static) field of the destination class.
+- **Reference**: a relationship from a class to another class on how the former
```
garrettjonesgoogle: How about: "A relationship from one class to another class such that the first class requires the presence the second class in a particular form in order to function."

dependencies/README.md
suztomo: For me, a noun (_static linkage checker_) sounds suitable than verb+objective (_check_static_link_).

Yet, something we can hold affection would be nice. For example _Maven_ is better name than  _Java Dependecy Management Tool_ (hypothetical name).
Now "_Saltlake_" just came up to my mind as the name of the tool. (Static Linkage Checker -> SLC -> Saltlake City -> Saltlake)

Talk to you tomorrow.

dependencies/README.md
suztomo: Updated.

dependencies/README.md
suztomo: I incorporated the enhancements. Do they look good now?

dependencies/README.md
suztomo: Added following in approach:

> The tool takes a classpath as required input, through either a list of jar and
  class files in filesystems, a list of Maven coordinates, or a Maven BOM.
  A Maven BOM is converted to a list of Maven coordinates, and a list of Maven
  coordinates is resolved to a list of jar files that consists of the artifacts
  and their dependencies.

library-best-practices/glossary.md
suztomo: Updated as

> (as a class literal, a field access, or a method invocation)

library-best-practices/glossary.md
suztomo: Right. Removed.

library-best-practices/glossary.md
suztomo: to discuss with @garrettjonesgoogle .

library-best-practices/glossary.md
suztomo: Added mention to removal and field.

library-best-practices/glossary.md
suztomo: Updated

library-best-practices/glossary.md
suztomo: Fixed

library-best-practices/glossary.md
suztomo: To discuss tomorrow.

library-best-practices/glossary.md
suztomo: "referenced class" sounds good. But "using class" doesn't for me.

How about "source class" and "referenced class"?

library-best-practices/glossary.md
suztomo: Right. Removed "However..."

library-best-practices/glossary.md
suztomo: Updated.
Still this part is not 100% sure though >  "in a particular form in order to function"

library-best-practices/glossary.md
```diff
@@ -66,3 +80,55 @@ Java Dependency Glossary
 - **Linkage-matchable version** (said of a particular version of A in relation
   to all versions of B): There exists some version of B such that the version of
   A and the version of B are linkage-compatible.
+
+
+### Class usage graph
```
suztomo: Class reference graph.

library-best-practices/glossary.md
```diff
@@ -66,3 +80,55 @@ Java Dependency Glossary
...
+
+- **Class usage graph**: a possibly cyclic directed graph of references
+  between classes. The nodes (vertices) of the graph correspond to
```
suztomo: bullet points.

- Each node is a Java class
- Each edge re...

dependencies/README.md
```diff
@@ -1,38 +1,42 @@
...
+  A Maven BOM is converted to a list of Maven coordinates, and a list of Maven
+  coordinates is resolved to a list of jar files that consists of the artifacts
+  and their dependencies.
+  The classpath is called as _linkage classpath_ on which the tool operates
+  to find linkage errors and is separated from the runtime classpath of the tool itself.
 
 2. The tool extracts all _references_ from the all class files in the classpath.
 
-3. The tool checks the linkage compatibility of the references through the classpath, and records
+3. The tool records references cannot be satisfied in the classpath as
```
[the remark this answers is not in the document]
suztomo: Fixed

library-best-practices/glossary.md
suztomo: Now it's 'source class' and 'target class' to follow terms in Guava Graph.

PR #225, opened 2018-11-06
include guava in bom

### LINKED ISSUE #220: Add Guava 20.0 to cloud-oss-bom

elharo:
fix #220 @garrettjonesgoogle

### DIFF
--- a/boms/cloud-oss-bom/pom.xml
+++ b/boms/cloud-oss-bom/pom.xml
@@ -42,6 +42,11 @@
         <artifactId>guava</artifactId>
         <version>${guava.version}</version>
       </dependency>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava-gwt</artifactId>
+        <version>${guava.version}</version>
+      </dependency>
 
       <!-- GRPC; specifically https://github.com/grpc/grpc-java -->
       <!-- We should import the officially-provided grpc BOM once it is created:
--- a/boms/cloud-oss-bom/pom.xml
+++ b/boms/cloud-oss-bom/pom.xml
@@ -29,6 +29,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <guava.version>20.0</guava.version>
     <google.cloud.java.version>0.67.0-alpha</google.cloud.java.version>
     <io.grpc.version>1.15.1</io.grpc.version>
     <protobuf.version>3.6.1</protobuf.version>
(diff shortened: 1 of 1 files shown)


boms/cloud-oss-bom/pom.xml
```diff
@@ -29,13 +29,20 @@
...
     <protobuf.version>3.6.1</protobuf.version>
   </properties>
 
   <dependencyManagement>
     <dependencies>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>${guava.version}</version>
+      </dependency>
```
garrettjonesgoogle: It appears that guava has more than one artifact, e.g. `guava-gwt`. Could you add them all here?

elharo: This is tricky for several reaosns:

1. guava-gwt is not in the google-cloud-java orbit.
2. What do we do about listenable future?
3. We claim that libraries in the BOM work together, but 26.0-jre doesn't work with all our other orbit dependencies. Not relevant immediately, but I expect it will become so.
4. Not all Guava artifacts have the same version number. Again, listenable future and also com.google.guava:failureaccess:1.0.

I suggest we just include the base artifact for now and try to resolve these issues in the eventual guava BOM.

garrettjonesgoogle: - (1)
  - (a) When Guava creates a BOM, guava-gwt will be in it, so when cloud-oss-bom imports Guava's BOM, guava-gwt will be there anyway.
  - (b) I think the choice of which artifacts are in the GCJ orbit is distinct from which artifacts of a library to include in a BOM. Once you decide that *any* artifact of a library is in an aggregated BOM, they *all* should be. The reason is that is probably *some* consumer out there that will use the artifact we care about together with the artifact(s) we don't care about; those consumers will need the artifacts to be at the same version.
- (2) & (4) If listenable future and failureaccess are released with guava and live in the guava repo, they should be part of guava's BOM (and thus we should include them here in preparation). The distinct version number isn't important. *However*, if they are released separately, then they probably wouldn't go into guava's BOM. 
- (3) We should be using 26.0-android of Guava.

To restate what I said yesterday: I think it makes sense to be incremental with "layers" (i.e. which libraries we pull in), but once we pull in a layer (a library), we should include all of its artifacts, otherwise some consumer somewhere will have trouble with making versions consistent.

garrettjonesgoogle: A search for guava packages also turns up `guava-testlib` and `guava-parent`: https://search.maven.org/search?q=g:com.google.guava

`guava-parent` probably isn't necessary, but anyone using Guava libraries and also testing using `guava-testlib` would want those versions to be the same. Could you add that too?

garrettjonesgoogle: google-cloud-java uses it: https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/pom.xml#L363

PR #238, opened 2018-11-12
Classes to represent references between classes

suztomo:
* Classes to represent symbolic references: `ClassSymbolReference`, `MethodSymbolReference`, `FieldSymbolReference` with AutoValue.

* The error classes (e.g., `LinkageErrorMissingField`) now have pointers to the references above.

* To represent a scan result of a jar, `SymbolTable` class is introduced to hold the set of these references above. StaticLinkageChecker.scanExternalSymbolTable generates a SymbolTable per jar file.

* new function `findLinkageErrors` is to start separating the logic of data extraction and validation. (previously, the validation was happening while data extraction progresses)

* `FullyQualifiedMethodReference` and `MethodSignature` classes are to be replaced with `MethodSymbolReference` class (not in this PR). So no need to review logics around FullyQualifiedMethodReference.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -16,10 +16,10 @@
 
 package com.google.cloud.tools.opensource.classpath;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Preconditions;
 import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
@@ -62,7 +62,7 @@ import org.apache.bcel.util.ClassPath;
 import org.apache.bcel.util.SyntheticRepository;
 
 /**
- * Class to read symbolic references in Java class files and to verify the availability
+ * Class to read symbol references in Java class files and to verify the availability
  * of references in them, through the input class path for a static linkage check.
  */
 class ClassDumper {
@@ -157,16 +157,15 @@ class ClassDumper {
 
   /**
    * Scans class files in the jar file and returns a {@link SymbolReferenceSet} populated with
-   * symbolic references.
+   * symbol references.
    *
    * @param jarFilePath absolute path to a jar file
-   * @return symbol references and classes defined in the jar file
    */
   static SymbolReferenceSet scanSymbolReferencesInJar(Path jarFilePath)
       throws ClassNotFoundException, IOException {
-    Preconditions.checkArgument(
+    checkArgument(
         jarFilePath.isAbsolute(), "The input jar file path is not an absolute path");
-    Preconditions.checkArgument(
+    checkArgument(
         Files.isReadable(jarFilePath), "The input jar file path is not readable");
 
     SymbolReferenceSet.Builder symbolTableBuilder = SymbolReferenceSet.builder();
@@ -217,7 +216,7 @@ class ClassDumper {
     return symbolTableBuilder.build();
   }
 
-  private static ConstantNameAndType constantNameAndTypeFromConstantCP(
+  private static ConstantNameAndType constantNameAndType(
       ConstantCP constantCP, ConstantPool constantPool) {
     int nameAndTypeIndex = constantCP.getNameAndTypeIndex();
     Constant constantAtNameAndTypeIndex = constantPool.getConstant(nameAndTypeIndex);
@@ -237,7 +236,7 @@ class ClassDumper {
       ConstantMethodref constantMethodref, ConstantPool constantPool, String sourceClassName) {
     String classNameInMethodReference = constantMethodref.getClass(constantPool);
     ConstantNameAndType constantNameAndType =
-        constantNameAndTypeFromConstantCP(constantMethodref, constantPool);
+        constantNameAndType(constantMethodref, constantPool);
     String methodName = constantNameAndType.getName(constantPool);
     String descriptor = constantNameAndType.getSignature(constantPool);
     MethodSymbolReference methodReference = MethodSymbolReference.builder()
@@ -254,7 +253,7 @@ class ClassDumper {
     // Either a class type or an interface type
     String classNameInFieldReference = constantFieldref.getClass(constantPool);
     ConstantNameAndType constantNameAndType =
-        constantNameAndTypeFromConstantCP(constantFieldref, constantPool);
+        constantNameAndType(constantFieldref, constantPool);
     String fieldName = constantNameAndType.getName(constantPool);
 
     FieldSymbolReference fieldSymbolReference = FieldSymbolReference.builder()
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -23,6 +23,7 @@ import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
+import com.google.common.collect.Sets;
 import com.google.common.reflect.ClassPath.ClassInfo;
 import java.io.File;
 import java.io.IOException;
@@ -176,15 +177,18 @@ class ClassDumper {
           fieldReferences.add(constantToFieldReference(constantFieldref, constantPool,
               sourceClassName));
           break;
+        case Const.CONSTANT_Class:
+          // TODO(suztomo): handle class reference
+          break;
         default:
           break;
       }
     }
-    ImmutableSet.Builder<String> definedClassNameBuilder =
-        symbolTableBuilder.definedClassNamesBuilder();
-    definedClassNameBuilder.add(sourceClassName).addAll(listInnerClassNames(javaClass));
 
-    return symbolTableBuilder.build();
+    Set<String> definedClassNames = Sets.newHashSet(listInnerClassNames(javaClass));
+    definedClassNames.add(sourceClassName);
+
+    return symbolTableBuilder.setDefinedClassNames(definedClassNames).build();
   }
 
   static ConstantNameAndType constantNameAndTypeFromConstantCP(
@@ -202,7 +206,7 @@ class ClassDumper {
     return (ConstantNameAndType) constantAtNameAndTypeIndex;
   }
 
-  static MethodSymbolReference constantToMethodReference(ConstantMethodref constantMethodref,
+  private static MethodSymbolReference constantToMethodReference(ConstantMethodref constantMethodref,
       ConstantPool constantPool, String sourceClassName) {
     String classNameInMethodReference = constantMethodref.getClass(constantPool);
     ConstantNameAndType constantNameAndType =
@@ -218,7 +222,7 @@ class ClassDumper {
     return methodReference;
   }
 
-  static FieldSymbolReference constantToFieldReference(ConstantFieldref constantFieldref,
+  private static FieldSymbolReference constantToFieldReference(ConstantFieldref constantFieldref,
       ConstantPool constantPool, String sourceClassName) {
     // Either a class type or an interface type
     String classNameInFieldReference = constantFieldref.getClass(constantPool);
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -146,8 +146,8 @@ class ClassDumper {
     return methodReferences.build();
   }
 
-  static SymbolsInFile scanClassSymbolTable(JavaClass javaClass) {
-    SymbolsInFile.Builder symbolTableBuilder = SymbolsInFile.builder();
+  static SymbolReferenceSet scanSymbolReferencesInClass(JavaClass javaClass) {
+    SymbolReferenceSet.Builder symbolTableBuilder = SymbolReferenceSet.builder();
     ImmutableSet.Builder<MethodSymbolReference> methodReferences =
         symbolTableBuilder.methodReferencesBuilder();
     ImmutableSet.Builder<FieldSymbolReference> fieldReferences =
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -24,6 +24,7 @@ import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
+import com.google.common.collect.SetMultimap;
 import com.google.common.reflect.ClassPath.ClassInfo;
 import java.io.File;
 import java.io.IOException;
@@ -75,20 +76,6 @@ class ClassDumper {
     return inputClasspath;
   }
 
-  private ClassDumper(
-      List<Path> inputClasspath,
-      SyntheticRepository syntheticRepository,
-      ClassLoader classLoader) throws ClassNotFoundException, IOException {
-    this.inputClasspath = ImmutableList.copyOf(inputClasspath);
-    this.syntheticRepository = syntheticRepository;
-    this.classLoader = classLoader;
-    this.jarFileToClasses = jarFilesToDefinedClasses(inputClasspath);
-  }
-
-  JavaClass loadJavaClass(String javaClassName) throws ClassNotFoundException {
-    return syntheticRepository.loadClass(javaClassName);
-  }
-
   static ClassDumper create(List<Path> jarFilePaths) throws IOException, ClassNotFoundException {
     // Creates classpath in the same order as inputClasspath for BCEL API
     String pathAsString =
@@ -107,7 +94,26 @@ class ClassDumper {
     URLClassLoader classLoaderFromJars =
         new URLClassLoader(jarFileUrls, ClassLoader.getSystemClassLoader());
 
-    return new ClassDumper(jarFilePaths, syntheticRepository, classLoaderFromJars);
+    return new ClassDumper(
+        jarFilePaths,
+        syntheticRepository,
+        classLoaderFromJars,
+        jarFilesToDefinedClasses(jarFilePaths));
+  }
+
+  private ClassDumper(
+      List<Path> inputClasspath,
+      SyntheticRepository syntheticRepository,
+      ClassLoader classLoader,
+      SetMultimap<Path, String> jarFileToClasses) {
+    this.inputClasspath = ImmutableList.copyOf(inputClasspath);
+    this.syntheticRepository = syntheticRepository;
+    this.classLoader = classLoader;
+    this.jarFileToClasses = ImmutableSetMultimap.copyOf(jarFileToClasses);
+  }
+
+  JavaClass loadJavaClass(String javaClassName) throws ClassNotFoundException {
+    return syntheticRepository.loadClass(javaClassName);
   }
 
   /**
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -16,6 +16,8 @@
 
 package com.google.cloud.tools.opensource.classpath;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableList;
@@ -42,6 +44,7 @@ import java.util.stream.Collectors;
 import org.apache.bcel.Const;
 import org.apache.bcel.classfile.Attribute;
 import org.apache.bcel.classfile.Constant;
+import org.apache.bcel.classfile.ConstantCP;
 import org.apache.bcel.classfile.ConstantFieldref;
 import org.apache.bcel.classfile.ConstantMethodref;
 import org.apache.bcel.classfile.ConstantNameAndType;
@@ -157,10 +160,10 @@ class ClassDumper {
     String sourceClassName = javaClass.getClassName();
     ConstantPool constantPool = javaClass.getConstantPool();
     Constant[] constants = constantPool.getConstantPool();
-    List<Constant> methodrefConstants = Arrays.stream(constants)
-        .filter(Predicates.notNull())
-        .collect(Collectors.toList());
-    for (Constant constant : methodrefConstants) {
+    for (Constant constant : constants) {
+      if (constant == null) {
+        continue;
+      }
       byte constantTag = constant.getTag();
       switch (constantTag) {
         case Const.CONSTANT_Methodref:
@@ -184,10 +187,9 @@ class ClassDumper {
     return symbolTableBuilder.build();
   }
 
-  static MethodSymbolReference constantToMethodReference(ConstantMethodref constantMethodref,
-      ConstantPool constantPool, String sourceClassName) {
-    String classNameInMethodReference = constantMethodref.getClass(constantPool);
-    int nameAndTypeIndex = constantMethodref.getNameAndTypeIndex();
+  static ConstantNameAndType constantNameAndTypeFromConstantCP(
+      ConstantCP constantCP, ConstantPool constantPool) {
+    int nameAndTypeIndex = constantCP.getNameAndTypeIndex();
     Constant constantAtNameAndTypeIndex = constantPool.getConstant(nameAndTypeIndex);
     if (!(constantAtNameAndTypeIndex instanceof ConstantNameAndType)) {
       // This constant_pool entry must be a CONSTANT_NameAndType_info
@@ -197,7 +199,14 @@ class ClassDumper {
               + nameAndTypeIndex
               + ". This class file is not compliant with CONSTANT_Methodref_info specification");
     }
-    ConstantNameAndType constantNameAndType = (ConstantNameAndType) constantAtNameAndTypeIndex;
+    return (ConstantNameAndType) constantAtNameAndTypeIndex;
+  }
+
+  static MethodSymbolReference constantToMethodReference(ConstantMethodref constantMethodref,
+      ConstantPool constantPool, String sourceClassName) {
+    String classNameInMethodReference = constantMethodref.getClass(constantPool);
+    ConstantNameAndType constantNameAndType =
+        constantNameAndTypeFromConstantCP(constantMethodref, constantPool);
     String methodName = constantNameAndType.getName(constantPool);
     String descriptor = constantNameAndType.getSignature(constantPool);
     MethodSymbolReference methodReference = MethodSymbolReference.builder()
@@ -213,17 +222,8 @@ class ClassDumper {
       ConstantPool constantPool, String sourceClassName) {
     // Either a class type or an interface type
     String classNameInFieldReference = constantFieldref.getClass(constantPool);
-    int nameAndTypeIndex = constantFieldref.getNameAndTypeIndex();
-    Constant constantAtNameAndTypeIndex = constantPool.getConstant(nameAndTypeIndex);
-    if (!(constantAtNameAndTypeIndex instanceof ConstantNameAndType)) {
-      // This constant_pool entry must be a CONSTANT_NameAndType_info
-      // as specified https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3.2
-      throw new RuntimeException(
-          "Failed to lookup nameAndType constant indexed "
-              + nameAndTypeIndex
-              + ". This class file is not compliant with field descriptor specification");
-    }
-    ConstantNameAndType constantNameAndType = (ConstantNameAndType) constantAtNameAndTypeIndex;
+    ConstantNameAndType constantNameAndType =
+        constantNameAndTypeFromConstantCP(constantFieldref, constantPool);
     String fieldName = constantNameAndType.getName(constantPool);
 
     FieldSymbolReference fieldSymbolReference = FieldSymbolReference.builder()
@@ -318,7 +318,7 @@ class ClassDumper {
     ImmutableList<FullyQualifiedMethodSignature> fullyQualifiedMethodSignatures = methods.stream()
         .map(method -> new FullyQualifiedMethodSignature(
             javaClass.getClassName(), method.getMethodName(), method.getDescriptor()))
-        .collect(ImmutableList.toImmutableList());
+        .collect(toImmutableList());
     return fullyQualifiedMethodSignatures;
   }
 
@@ -364,7 +364,7 @@ class ClassDumper {
     Method[] methods = javaClass.getMethods();
     ImmutableList<MethodSignature> signatures = Arrays.stream(methods)
         .map(method -> new MethodSignature(method.getName(), method.getSignature()))
-        .collect(ImmutableList.toImmutableList());
+        .collect(toImmutableList());
     return signatures;
   }
 
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -225,8 +225,8 @@ class ClassDumper {
     return (ConstantNameAndType) constantAtNameAndTypeIndex;
   }
 
-  private static MethodSymbolReference constantToMethodReference(ConstantMethodref constantMethodref,
-      ConstantPool constantPool, String sourceClassName) {
+  private static MethodSymbolReference constantToMethodReference(
+      ConstantMethodref constantMethodref, ConstantPool constantPool, String sourceClassName) {
     String classNameInMethodReference = constantMethodref.getClass(constantPool);
     ConstantNameAndType constantNameAndType =
         constantNameAndTypeFromConstantCP(constantMethodref, constantPool);
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -170,7 +170,7 @@ class ClassDumper {
 
     SymbolReferenceSet.Builder symbolTableBuilder = SymbolReferenceSet.builder();
     for (JavaClass javaClass : topLevelJavaClassesInJar(jarFilePath)) {
-      symbolTableBuilder.merge(scanSymbolReferencesInClass(javaClass));
+      symbolTableBuilder.addAll(scanSymbolReferencesInClass(javaClass));
     }
     return symbolTableBuilder.build();
   }
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -19,6 +19,7 @@ package com.google.cloud.tools.opensource.classpath;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
@@ -31,6 +32,7 @@ import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.CodeSource;
@@ -146,7 +148,31 @@ class ClassDumper {
     return methodReferences.build();
   }
 
-  static SymbolReferenceSet scanSymbolReferencesInClass(JavaClass javaClass) {
+  /**
+   * Scans class files in the jar file and returns a {@link SymbolReferenceSet} populated with
+   * symbolic references.
+   *
+   * @param jarFilePath absolute path to a jar file
+   * @return symbol references and classes defined in the jar file
+   */
+  static SymbolReferenceSet scanSymbolReferencesInJar(Path jarFilePath) {
+    Preconditions.checkArgument(
+        jarFilePath.isAbsolute(), "The input jar file path is not an absolute path");
+    Preconditions.checkArgument(
+        Files.isReadable(jarFilePath), "The input jar file path is not readable");
+
+    SymbolReferenceSet.Builder symbolTableBuilder = SymbolReferenceSet.builder();
+    try {
+      for (JavaClass javaClass : topLevelJavaClassesInJar(jarFilePath)) {
+        symbolTableBuilder.merge(scanSymbolReferencesInClass(javaClass));
+      }
+      return symbolTableBuilder.build();
+    } catch (ClassNotFoundException | IOException ex) {
+      throw new RuntimeException("Failed to scan jar file", ex);
+    }
+  }
+
+  private static SymbolReferenceSet scanSymbolReferencesInClass(JavaClass javaClass) {
     SymbolReferenceSet.Builder symbolTableBuilder = SymbolReferenceSet.builder();
     ImmutableSet.Builder<MethodSymbolReference> methodReferences =
         symbolTableBuilder.methodReferencesBuilder();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -45,6 +45,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import org.apache.bcel.Const;
 import org.apache.bcel.classfile.Attribute;
+import org.apache.bcel.classfile.ClassFormatException;
 import org.apache.bcel.classfile.Constant;
 import org.apache.bcel.classfile.ConstantCP;
 import org.apache.bcel.classfile.ConstantFieldref;
@@ -77,7 +78,7 @@ class ClassDumper {
   private ClassDumper(
       List<Path> inputClasspath,
       SyntheticRepository syntheticRepository,
-      ClassLoader classLoader) {
+      ClassLoader classLoader) throws ClassNotFoundException, IOException {
     this.inputClasspath = ImmutableList.copyOf(inputClasspath);
     this.syntheticRepository = syntheticRepository;
     this.classLoader = classLoader;
@@ -88,7 +89,7 @@ class ClassDumper {
     return syntheticRepository.loadClass(javaClassName);
   }
 
-  static ClassDumper create(List<Path> jarFilePaths) {
+  static ClassDumper create(List<Path> jarFilePaths) throws IOException, ClassNotFoundException {
     // Creates classpath in the same order as inputClasspath for BCEL API
     String pathAsString =
         jarFilePaths.stream().map(Path::toString).collect(Collectors.joining(File.pathSeparator));
@@ -133,7 +134,7 @@ class ClassDumper {
       if (!(constantAtNameAndTypeIndex instanceof ConstantNameAndType)) {
         // This constant_pool entry must be a CONSTANT_NameAndType_info
         // as specified https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.4.2
-        throw new RuntimeException(
+        throw new ClassFormatException(
             "Failed to lookup nameAndType constant indexed "
                 + nameAndTypeIndex
                 + ". This class file is not compliant with CONSTANT_Methodref_info specification");
@@ -217,7 +218,7 @@ class ClassDumper {
     if (!(constantAtNameAndTypeIndex instanceof ConstantNameAndType)) {
       // This constant_pool entry must be a CONSTANT_NameAndType_info
       // as specified https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.4.2
-      throw new RuntimeException(
+      throw new ClassFormatException(
           "Failed to lookup nameAndType constant indexed "
               + nameAndTypeIndex
               + ". This class file is not compliant with CONSTANT_Methodref_info specification");
@@ -485,21 +486,17 @@ class ClassDumper {
    * @return map of jar file paths to classes defined in them
    */
   private static ImmutableSetMultimap<Path, String> jarFilesToDefinedClasses(
-      List<Path> jarFilePaths) {
+      List<Path> jarFilePaths) throws IOException, ClassNotFoundException {
     ImmutableSetMultimap.Builder<Path, String> pathToClasses =
         ImmutableSetMultimap.builder();
 
     for (Path jarFilePath : jarFilePaths) {
-      try {
-        for (JavaClass javaClass: topLevelJavaClassesInJar(jarFilePath)) {
-          pathToClasses.put(jarFilePath, javaClass.getClassName());
-          // This does not take double-nested classes. As long as such classes are accessed
-          // only from the outer class, static linkage checker does not report false positives
-          // TODO(suztomo): enhance this so that it can work with double-nested classes
-          pathToClasses.putAll(jarFilePath, listInnerClassNames(javaClass));
-        }
-      } catch (IOException | ClassNotFoundException ex) {
-        throw new RuntimeException("There was problem in loading classes in jar file", ex);
+      for (JavaClass javaClass: topLevelJavaClassesInJar(jarFilePath)) {
+        pathToClasses.put(jarFilePath, javaClass.getClassName());
+        // This does not take double-nested classes. As long as such classes are accessed
+        // only from the outer class, static linkage checker does not report false positives
+        // TODO(suztomo): enhance this so that it can work with double-nested classes
+        pathToClasses.putAll(jarFilePath, listInnerClassNames(javaClass));
       }
     }
     return pathToClasses.build();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -42,6 +42,7 @@ import java.util.stream.Collectors;
 import org.apache.bcel.Const;
 import org.apache.bcel.classfile.Attribute;
 import org.apache.bcel.classfile.Constant;
+import org.apache.bcel.classfile.ConstantFieldref;
 import org.apache.bcel.classfile.ConstantMethodref;
 import org.apache.bcel.classfile.ConstantNameAndType;
 import org.apache.bcel.classfile.ConstantPool;
@@ -140,6 +141,97 @@ class ClassDumper {
       methodReferences.add(methodref);
     }
     return methodReferences.build();
+  }
+
+  static SymbolTable scanClassSymbolTable(JavaClass javaClass) {
+    SymbolTable.Builder symbolTableBuilder = SymbolTable.builder();
+    ImmutableSet.Builder<MethodSymbolReference> methodReferences =
+        symbolTableBuilder.methodReferencesBuilder();
+    ImmutableSet.Builder<FieldSymbolReference> fieldReferences =
+        symbolTableBuilder.fieldReferencesBuilder();
+
+    // TODO(suztomo): Read class references (inheritance) from javaClass file
+    ImmutableSet.Builder<ClassSymbolReference> classReferences =
+        symbolTableBuilder.classReferencesBuilder();
+
+    String sourceClassName = javaClass.getClassName();
+    ConstantPool constantPool = javaClass.getConstantPool();
+    Constant[] constants = constantPool.getConstantPool();
+    List<Constant> methodrefConstants = Arrays.stream(constants)
+        .filter(Predicates.notNull())
+        .collect(Collectors.toList());
+    for (Constant constant : methodrefConstants) {
+      byte constantTag = constant.getTag();
+      switch (constantTag) {
+        case Const.CONSTANT_Methodref:
+          ConstantMethodref constantMethodref = (ConstantMethodref) constant;
+          methodReferences.add(constantToMethodReference(constantMethodref, constantPool,
+              sourceClassName));
+          break;
+        case Const.CONSTANT_Fieldref:
+          ConstantFieldref constantFieldref = (ConstantFieldref) constant;
+          fieldReferences.add(constantToFieldReference(constantFieldref, constantPool,
+              sourceClassName));
+          break;
+        default:
+          break;
+      }
+    }
+    ImmutableSet.Builder<String> definedClassNameBuilder =
+        symbolTableBuilder.definedClassNamesBuilder();
+    definedClassNameBuilder.add(sourceClassName).addAll(listInnerClassNames(javaClass));
+
+    return symbolTableBuilder.build();
+  }
+
+  static MethodSymbolReference constantToMethodReference(ConstantMethodref constantMethodref,
+      ConstantPool constantPool, String sourceClassName) {
+    String classNameInMethodReference = constantMethodref.getClass(constantPool);
+    int nameAndTypeIndex = constantMethodref.getNameAndTypeIndex();
+    Constant constantAtNameAndTypeIndex = constantPool.getConstant(nameAndTypeIndex);
+    if (!(constantAtNameAndTypeIndex instanceof ConstantNameAndType)) {
+      // This constant_pool entry must be a CONSTANT_NameAndType_info
+      // as specified https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.4.2
+      throw new RuntimeException(
+          "Failed to lookup nameAndType constant indexed "
+              + nameAndTypeIndex
+              + ". This class file is not compliant with CONSTANT_Methodref_info specification");
+    }
+    ConstantNameAndType constantNameAndType = (ConstantNameAndType) constantAtNameAndTypeIndex;
+    String methodName = constantNameAndType.getName(constantPool);
+    String descriptor = constantNameAndType.getSignature(constantPool);
+    MethodSymbolReference methodReference = MethodSymbolReference.builder()
+        .setSourceClassName(sourceClassName)
+        .setMethodName(methodName)
+        .setTargetClassName(classNameInMethodReference)
+        .setDescriptor(descriptor)
+        .build();
+    return methodReference;
+  }
+
+  static FieldSymbolReference constantToFieldReference(ConstantFieldref constantFieldref,
+      ConstantPool constantPool, String sourceClassName) {
+    // Either a class type or an interface type
+    String classNameInFieldReference = constantFieldref.getClass(constantPool);
+    int nameAndTypeIndex = constantFieldref.getNameAndTypeIndex();
+    Constant constantAtNameAndTypeIndex = constantPool.getConstant(nameAndTypeIndex);
+    if (!(constantAtNameAndTypeIndex instanceof ConstantNameAndType)) {
+      // This constant_pool entry must be a CONSTANT_NameAndType_info
+      // as specified https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3.2
+      throw new RuntimeException(
+          "Failed to lookup nameAndType constant indexed "
+              + nameAndTypeIndex
+              + ". This class file is not compliant with field descriptor specification");
+    }
+    ConstantNameAndType constantNameAndType = (ConstantNameAndType) constantAtNameAndTypeIndex;
+    String fieldName = constantNameAndType.getName(constantPool);
+
+    FieldSymbolReference fieldSymbolReference = FieldSymbolReference.builder()
+        .setSourceClassName(sourceClassName)
+        .setFieldName(fieldName)
+        .setTargetClassName(classNameInFieldReference)
+        .build();
+    return fieldSymbolReference;
   }
 
   /**
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -23,7 +23,6 @@ import com.google.common.base.Predicates;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
-import com.google.common.collect.Sets;
 import com.google.common.reflect.ClassPath.ClassInfo;
 import java.io.File;
 import java.io.IOException;
@@ -185,10 +184,7 @@ class ClassDumper {
       }
     }
 
-    Set<String> definedClassNames = Sets.newHashSet(listInnerClassNames(javaClass));
-    definedClassNames.add(sourceClassName);
-
-    return symbolTableBuilder.setDefinedClassNames(definedClassNames).build();
+    return symbolTableBuilder.build();
   }
 
   static ConstantNameAndType constantNameAndTypeFromConstantCP(
@@ -471,10 +467,8 @@ class ClassDumper {
         ImmutableSetMultimap.builder();
 
     for (Path jarFilePath : jarFilePaths) {
-      String pathToJar = jarFilePath.toString();
-      SyntheticRepository repository = SyntheticRepository.getInstance(new ClassPath(pathToJar));
       try {
-        for (JavaClass javaClass: topLevelJavaClassesInJar(jarFilePath, repository)) {
+        for (JavaClass javaClass: topLevelJavaClassesInJar(jarFilePath)) {
           pathToClasses.put(jarFilePath, javaClass.getClassName());
           // This does not take double-nested classes. As long as such classes are accessed
           // only from the outer class, static linkage checker does not report false positives
@@ -504,8 +498,10 @@ class ClassDumper {
     return allClassesInJar;
   }
 
-  static ImmutableSet<JavaClass> topLevelJavaClassesInJar(Path jarFilePath,
-      SyntheticRepository repository) throws IOException, ClassNotFoundException {
+  static ImmutableSet<JavaClass> topLevelJavaClassesInJar(Path jarFilePath)
+      throws IOException, ClassNotFoundException {
+    String pathToJar = jarFilePath.toString();
+    SyntheticRepository repository = SyntheticRepository.getInstance(new ClassPath(pathToJar));
     ImmutableSet.Builder<JavaClass> javaClasses = ImmutableSet.builder();
     URL jarFileUrl = jarFilePath.toUri().toURL();
     for (ClassInfo classInfo : listTopLevelClassesFromJar(jarFileUrl)) {
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -147,8 +147,8 @@ class ClassDumper {
     return methodReferences.build();
   }
 
-  static SymbolTable scanClassSymbolTable(JavaClass javaClass) {
-    SymbolTable.Builder symbolTableBuilder = SymbolTable.builder();
+  static SymbolsInFile scanClassSymbolTable(JavaClass javaClass) {
+    SymbolsInFile.Builder symbolTableBuilder = SymbolsInFile.builder();
     ImmutableSet.Builder<MethodSymbolReference> methodReferences =
         symbolTableBuilder.methodReferencesBuilder();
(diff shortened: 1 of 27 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassSymbolReference.java
```diff
@@ -0,0 +1,36 @@
...
+  abstract String getTargetClassName();
+  abstract String getSourceClassName();
+
+  static Builder builder() {
+    return new AutoValue_ClassSymbolReference.Builder();
+  }
+
+  @AutoValue.Builder
+  abstract static class Builder {
+    abstract Builder setTargetClassName(String value);
```
elharo: value --> name

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassSymbolReference.java
```diff
@@ -0,0 +1,36 @@
...
+    abstract Builder setSourceClassName(String value);
```
elharo: value --> name

suztomo:
@elharo ping

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -142,6 +143,97 @@ static ClassDumper create(List<Path> jarFilePaths) {
...
+    // TODO(suztomo): Read class references (inheritance) from javaClass file
+    ImmutableSet.Builder<ClassSymbolReference> classReferences =
+        symbolTableBuilder.classReferencesBuilder();
+
+    String sourceClassName = javaClass.getClassName();
+    ConstantPool constantPool = javaClass.getConstantPool();
+    Constant[] constants = constantPool.getConstantPool();
+    List<Constant> methodrefConstants = Arrays.stream(constants)
+        .filter(Predicates.notNull())
+        .collect(Collectors.toList());
```
netdpb: What about this? Then you don't need the `methodRefConstants` variable at all, and you don't need to create a list object.

```suggestion
        .forEach(constant -> { ... });
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -142,6 +143,97 @@ static ClassDumper create(List<Path> jarFilePaths) {
...
+        .setMethodName(methodName)
+        .setTargetClassName(classNameInMethodReference)
+        .setDescriptor(descriptor)
+        .build();
+    return methodReference;
+  }
+
+  static FieldSymbolReference constantToFieldReference(ConstantFieldref constantFieldref,
+      ConstantPool constantPool, String sourceClassName) {
+    // Either a class type or an interface type
```
netdpb: These lines look identical to the start of `constantToMethodReference()`. Can you extract a method that takes a `ConstantCP` and `ConstantPool` and returns a `ConstantNameAndType`, throwing if it can't be cast? Then these two methods will have only what's different between methods and fields.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassSymbolReference.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.opensource.classpath;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+abstract class ClassSymbolReference {
```
netdpb: Let's use Javadoc for everything that's not private and not trivial.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FieldSymbolReference.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.opensource.classpath;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+abstract class FieldSymbolReference {
+  abstract String getTargetClassName();
```
netdpb: Would it be useful for `FieldSymbolReference` and `ClassSymbolReference` (and `MethodSymbolReference` too) to implement a common interface like `SymbolReference` that has a source and target class name? They all share those properties.

Their builders could all implement a common interface with those setters, too, although that requires the builder interface to have a self-bounded generic type like `Builder<B extends Builder<B>>` so the setters can return the right builder. That part is up to you; it's not a huge advantage.

But treating all the references as subtypes is maybe more useful.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorMissingClass.java
```diff
@@ -17,23 +17,20 @@
...
 import com.google.auto.value.AutoValue;
-
 /**
  * A missing class linkage error.
  */
 @AutoValue
 abstract class LinkageErrorMissingClass {
-  abstract String getTargetClassName();
-  abstract String getSourceClassName();
+  abstract ClassSymbolReference getReference();
```
netdpb: Do we still need the "LinkageError" classes at all? Are they each just going to hold onto one missing reference, or do you think they'll acquire other properties?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,90 @@ private static void printStaticLinkageErrors(
     return jarPaths.build();
   }
 
+  /**
+   * Finds linkage errors in the input classpath and generates a static linkage check report.
+   */
+  StaticLinkageCheckReport findLinkageErrors() {
+    ImmutableList<Path> jarFilePaths = classDumper.getInputClasspath();
+
+    ImmutableMap.Builder<Path, SymbolTable> jarToSymbols = ImmutableMap.builder();
```
netdpb: What about doing this in one step?

```
ImmutableMap<Path, SymbolTable> jarToSymbols =
    jarFilePaths.stream().collect(toImmutableMap(
        jarPath -> jarPath, jarPath -> scanExternalSymbolTable(jarPath));
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,90 @@ private static void printStaticLinkageErrors(
...
+    for (Path jarPath : jarFilePaths) {
+      jarToSymbols.put(jarPath, scanExternalSymbolTable(jarPath));
+    }
+
+    // Validate linkage error of each reference
+    ImmutableList.Builder<JarLinkageReport> jarLinkageReports = ImmutableList.builder();
```
netdpb: Again, you can do this in one step if you'd like:

```
jarToSymbols.entrySet().stream()
    .map(entry -> findInvalidReferences(entry.getKey(), entry.getValue()))
    .collect(toImmutableList());
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,90 @@ private static void printStaticLinkageErrors(
...
+    for (Map.Entry<Path, SymbolTable> entry : jarToSymbols.build().entrySet()) {
+      jarLinkageReports.add(findInvalidReferences(entry.getKey(), entry.getValue()));
+    }
+
+    if (reportOnlyReachable) {
+      // TODO: Optionally, report errors only reachable from entry point classes
```
netdpb: Maybe warn that this is unimplemented for now?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,90 @@ private static void printStaticLinkageErrors(
...
+
+  private JarLinkageReport findInvalidReferences(Path jarPath, SymbolTable jarSymbolTable) {
+    JarLinkageReport.Builder reportBuilder = JarLinkageReport.builder().setJarPath(jarPath);
+
+    // TODO(suztomo): implement validation for field, method and class references in the table
+    return reportBuilder.build();
+  }
+
+  /**
+   * Scans class files in the jar file and creates {@link SymbolTable} populated with class names
```
netdpb:
```suggestion
   * Scans class files in the jar file and returns a {@link SymbolTable} populated with class names
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,90 @@ private static void printStaticLinkageErrors(
...
+   * defined within the jar and symbolic references to other classes outside the jar.
+   *
+   * @param jarFilePath absolute path to a jar file.
```
netdpb:
```suggestion
   * @param jarFilePath absolute path to a jar file
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,90 @@ private static void printStaticLinkageErrors(
...
+   * @return Table containing symbol references and classes defined in the jar file.
```
netdpb:
```suggestion
   * @return symbol references and classes defined in the jar file
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,90 @@ private static void printStaticLinkageErrors(
...
+   */
+  static SymbolTable scanExternalSymbolTable(Path jarFilePath) {
+    SymbolTable.Builder symbolTableBuilder = SymbolTable.builder();
+    ImmutableSet.Builder<String> symbolTableClassNameBuilder = ImmutableSet.builder();
```
netdpb: If you declare `SymbolTable.Builder` to have methods that return these builders, then you don't even need these variables. Below, you can do things like `symbolTableBuilder.classReferencesBuilder().addAll(partialSymbolTable.getClassReferences())`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,90 @@ private static void printStaticLinkageErrors(
...
+    ImmutableSet.Builder<FieldSymbolReference> fieldSymbolReferences = ImmutableSet.builder();
+
+    String pathToJar = jarFilePath.toString();
+    SyntheticRepository repository = SyntheticRepository.getInstance(new ClassPath(pathToJar));
+
+    try {
+      for (JavaClass javaClass : ClassDumper.topLevelJavaClassesInJar(jarFilePath, repository)) {
+        SymbolTable partialSymbolTable = ClassDumper.scanClassSymbolTable(javaClass);
+
+        classSymbolReferences.addAll(partialSymbolTable.getClassReferences());
```
netdpb: If it makes sense, you can add a concrete method to `SymbolTable.Builder` that takes a `SymbolTable` and merges it, like `Builder merge(SymbolTable other)`. You can have any concrete methods you want on an AutoValue builder (or the class itself).

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,90 @@ private static void printStaticLinkageErrors(
...
+        methodSymbolReferences.addAll(partialSymbolTable.getMethodReferences());
+        fieldSymbolReferences.addAll(partialSymbolTable.getFieldReferences());
+        symbolTableClassNameBuilder.addAll(partialSymbolTable.getDefinedClassNames());
+      }
+      ImmutableSet<String> classesDefinedInJar = symbolTableClassNameBuilder.build();
+
+      ImmutableSet<ClassSymbolReference> externalClassReferences = classSymbolReferences.build()
+          .stream()
+          .filter(reference -> !classesDefinedInJar.contains(reference.getTargetClassName()))
+          .collect(ImmutableSet.toImmutableSet());
```
netdpb: These `Collector` methods are good to statically import.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,90 @@ private static void printStaticLinkageErrors(
...
+      ImmutableSet<MethodSymbolReference> externalMethodReferences = methodSymbolReferences.build()
+          .stream()
+          .filter(reference -> !classesDefinedInJar.contains(reference.getTargetClassName()))
```
netdpb: Check out `com.google.common.collect.Sets.filter()` for another way to do this.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,90 @@ private static void printStaticLinkageErrors(
...
+          .setMethodReferences(externalMethodReferences)
+          .setFieldReferences(externalFieldReferences)
+          .setDefinedClassNames(classesDefinedInJar)
+          .build();
+    } catch (ClassNotFoundException | IOException ex) {
+      throw new RuntimeException("Failed to scan jar file", ex);
+    }
+  }
+
+
```
netdpb: Delete extra blank line.

Can we use google-java-format for every PR? Specifically in the mode that formats only changed lines.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolTable.java
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.opensource.classpath;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableSet;
+
+/**
+ * A table of list of symbolic references and defined classes in a jar file or class file.
```
netdpb: How is this a table?

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassSymbolReferenceTest.java
```diff
@@ -0,0 +1,35 @@
...
+public class ClassSymbolReferenceTest {
+  @Test
+  public void testCreation() {
+    ClassSymbolReference classSymbolReference =
+        ClassSymbolReference.builder()
+            .setTargetClassName("ClassA")
+            .setSourceClassName("ClassB")
+            .build();
+
+    Assert.assertEquals("ClassA", classSymbolReference.getTargetClassName());
```
netdpb: I still feel that these tests are not worth your time.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,90 @@ private static void printStaticLinkageErrors(
...
+          .filter(reference -> !classesDefinedInJar.contains(reference.getTargetClassName()))
+          .collect(ImmutableSet.toImmutableSet());
+
+      return symbolTableBuilder
+          .setClassReferences(externalClassReferences)
+          .setMethodReferences(externalMethodReferences)
+          .setFieldReferences(externalFieldReferences)
+          .setDefinedClassNames(classesDefinedInJar)
+          .build();
+    } catch (ClassNotFoundException | IOException ex) {
```
elharo: It probably makes more sense to bubble these exceptions here rather than converting to a checked exception we might miss

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -193,6 +282,7 @@ private static void printStaticLinkageErrors(
   ImmutableList<FullyQualifiedMethodSignature> findUnresolvedMethodReferences()
       throws IOException, ClassNotFoundException {
     // TODO(suztomo): Separate logic between data retrieval and usage graph traversal. Issue #203
+    // This method, applicable only for method reference, is to be replaced with findLinkageErrors
```
elharo: todo:

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolTable.java
elharo: delete "list of"

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolTable.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+@AutoValue
+abstract class SymbolTable {
```
elharo: Javadoc with links to relevant section of JVM spec could be helpful here.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSymbolReference.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package com.google.cloud.tools.opensource.classpath;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+abstract class MethodSymbolReference {
+  abstract String getTargetClassName();
+  abstract String getMethodName();
```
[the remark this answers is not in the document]
suztomo: Added

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSymbolReference.java
```diff
@@ -0,0 +1,40 @@
...
+  abstract String getDescriptor();
```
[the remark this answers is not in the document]
suztomo: Added with url to JVM specification

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -47,14 +47,14 @@ static Builder builder() {
 
   @AutoValue.Builder
   abstract static class Builder {
-    abstract Builder setBomCoordinate(@Nullable String value);
-    abstract Builder setMavenCoordinates(List<String> value);
-    abstract Builder setJarFileList(List<Path> value);
+    abstract Builder setBomCoordinate(@Nullable String coordinate);
+    abstract Builder setMavenCoordinates(List<String> coordinates);
```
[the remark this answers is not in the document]
suztomo: Added javadoc.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -142,6 +143,97 @@ static ClassDumper create(List<Path> jarFilePaths) {
...
+          methodReferences.add(constantToMethodReference(constantMethodref, constantPool,
+              sourceClassName));
+          break;
+        case Const.CONSTANT_Fieldref:
+          ConstantFieldref constantFieldref = (ConstantFieldref) constant;
+          fieldReferences.add(constantToFieldReference(constantFieldref, constantPool,
+              sourceClassName));
+          break;
+        default:
+          break;
```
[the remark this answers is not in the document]
suztomo: Added TODO.

There're many unnecessary entries in constant pool so ignoring by default is fine.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -142,6 +143,97 @@ static ClassDumper create(List<Path> jarFilePaths) {
...
+      }
+    }
+    ImmutableSet.Builder<String> definedClassNameBuilder =
+        symbolTableBuilder.definedClassNamesBuilder();
+    definedClassNameBuilder.add(sourceClassName).addAll(listInnerClassNames(javaClass));
```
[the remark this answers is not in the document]
suztomo: Simplified:

```
    Set<String> definedClassNames = Sets.newHashSet(listInnerClassNames(javaClass));
    definedClassNames.add(sourceClassName);

    return symbolTableBuilder.setDefinedClassNames(definedClassNames).build();

```

suztomo:
Renamed SymbolTable to SymbolsInFile.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSymbolReference.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.opensource.classpath;
+
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+abstract class MethodSymbolReference {
+  abstract String getTargetClassName();
```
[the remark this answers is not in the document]
suztomo: Added javadoc to SymbolReference interface.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -142,6 +143,97 @@ static ClassDumper create(List<Path> jarFilePaths) {
...
+
+    return symbolTableBuilder.build();
+  }
+
+  static MethodSymbolReference constantToMethodReference(ConstantMethodref constantMethodref,
```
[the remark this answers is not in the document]
suztomo: refactored

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -142,6 +147,97 @@ static ClassDumper create(List<Path> jarFilePaths) {
...
+          break;
+        case Const.CONSTANT_Class:
+          // TODO(suztomo): handle class reference
+          break;
+        default:
+          break;
+      }
+    }
+
+    Set<String> definedClassNames = Sets.newHashSet(listInnerClassNames(javaClass));
```
elharo: Cab you use a construcctor here instead?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -142,6 +146,94 @@ static ClassDumper create(List<Path> jarFilePaths) {
     return methodReferences.build();
   }
 
+  static SymbolsInFile scanClassSymbolTable(JavaClass javaClass) {
+    SymbolsInFile.Builder symbolTableBuilder = SymbolsInFile.builder();
```
elharo: Would SymbolsInFile be better named SymbolTable?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -142,6 +146,94 @@ static ClassDumper create(List<Path> jarFilePaths) {
...
+  }
+
+  static ConstantNameAndType constantNameAndTypeFromConstantCP(
+      ConstantCP constantCP, ConstantPool constantPool) {
+    int nameAndTypeIndex = constantCP.getNameAndTypeIndex();
+    Constant constantAtNameAndTypeIndex = constantPool.getConstant(nameAndTypeIndex);
+    if (!(constantAtNameAndTypeIndex instanceof ConstantNameAndType)) {
+      // This constant_pool entry must be a CONSTANT_NameAndType_info
+      // as specified https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.4.2
+      throw new RuntimeException(
```
elharo: Per https://engdoc.corp.google.com/eng/doc/devguide/java/practices/exceptions.md?cl=head "Treat Throwable, Exception, Error, and RuntimeException as if they were abstract: avoid constructing or throwing a new instance of any of these types directly."

This might be better as a checked exception anyway. I'd need to think about that.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassSymbolReference.java
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.opensource.classpath;
+
+import com.google.auto.value.AutoValue;
+
+/**
+ * A symbolic reference to {@code targetClass} itself referenced from {@code sourceClass}.
```
elharo: delete "itself"

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -32,13 +32,43 @@
...
 abstract class StaticLinkageCheckOption {
   // TODO(suztomo): Add option to specify entry point classes
 
+  /**
+   * Returns a Maven coordinate for a BOM if specified; otherwise null.
+   */
   @Nullable abstract String getBomCoordinate();
+
+  /**
+   * Returns Maven coordinates if specified; otherwise an empty list.
```
elharo: but maven coordinates for what? Why is this field here, especially compared to the BOM coordinates? For example (not sure if this is correct)

returns the Maven coordinates of the artifact being scanned for static linkage errors

or 

returns the Maven coordinates of the artifact that has a.static linkage error

These are two *different* things and from the code alone I can't tell which is intended.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
elharo: throw new UnsupportedOperationException()?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
elharo: You haven't done that though. You've just hidden the exception. It's still there. The caller should be given this information.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReference.java
```diff
@@ -0,0 +1,38 @@
...
+ * names are fully-qualified form known as binary names. For example {@code
+ * io.grpc.MethodDescriptor$MethodType}.
+ *
+ * @see <a href="https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.1">Java
+ *     Language Specification: 13.1. The Form of a Binary</a>
+ */
+interface SymbolReference {
+
+  /**
+   * Returns fully-qualified class name of the source class of the reference.
```
elharo: perhaps "fully-package qualified name of the source class of the reference. For example, com.example.Utility"

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassSymbolReferenceTest.java
elharo: Unit tests aren't only for verifying correctness. They also help explore and design the API in advance of implementation.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -182,6 +187,63 @@ private static void printStaticLinkageErrors(
...
+  static SymbolReferenceSet scanSymbolReferencesInJar(Path jarFilePath) {
+    Preconditions.checkArgument(
+        jarFilePath.isAbsolute(), "The input jar file path is not an absolute path");
+    Preconditions.checkArgument(
+        Files.isReadable(jarFilePath), "The input jar file path is not readable");
+
+    SymbolReferenceSet.Builder symbolTableBuilder = SymbolReferenceSet.builder();
+    try {
+      for (JavaClass javaClass : ClassDumper.topLevelJavaClassesInJar(jarFilePath)) {
+        symbolTableBuilder.merge(ClassDumper.scanSymbolReferencesInClass(javaClass));
```
suztomo: Thx to `merge` in SymbolRefrenceSet, the code got much simpler.

suztomo:
@elharo Would you check the change?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -142,6 +148,118 @@ static ClassDumper create(List<Path> jarFilePaths) {
...
+    Preconditions.checkArgument(
+        Files.isReadable(jarFilePath), "The input jar file path is not readable");
+
+    SymbolReferenceSet.Builder symbolTableBuilder = SymbolReferenceSet.builder();
+    try {
+      for (JavaClass javaClass : topLevelJavaClassesInJar(jarFilePath)) {
+        symbolTableBuilder.merge(scanSymbolReferencesInClass(javaClass));
+      }
+      return symbolTableBuilder.build();
+    } catch (ClassNotFoundException | IOException ex) {
```
elharo: These exceptions should bubble up.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -32,13 +32,46 @@
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
 
+/**
+ * Option for {@link StaticLinkageChecker}. To construct an input class path, the checker requires
+ * either one of the following types of input:
+ *
+ * <ul>
+ *   <li>{@code bomCoordinate}: a Maven coordinate for a BOM
+ *   <li>{@code mavenCoordinates}: list of the coordinates of Maven artifacts
```
elharo: artifacts to check

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -32,13 +32,46 @@
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
 
+/**
+ * Option for {@link StaticLinkageChecker}. To construct an input class path, the checker requires
+ * either one of the following types of input:
```
elharo: delete either since there are three options

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -32,13 +32,46 @@
...
 abstract class StaticLinkageCheckOption {
   // TODO(suztomo): Add option to specify entry point classes
 
-  @Nullable abstract String getBomCoordinate();
+  /**
+   * Returns a Maven coordinate for a BOM if specified; otherwise null. Example value: {@code
+   * com.google.cloud:cloud-oss-bom:pom:1.0.0-SNAPSHOT}
+   */
+  @Nullable
+  abstract String getBomCoordinate();
```
elharo: What happens if more than one of these is set?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Fixed

suztomo:
@elharo Applied changes. Would you check?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
elharo: either one of the following types of input --> exactly one of

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -32,13 +32,46 @@
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
 
+/**
+ * Option for {@link StaticLinkageChecker}. To construct an input class path, the checker requires
+ * either one of the following types of input:
+ *
+ * <ul>
+ *   <li>{@code bomCoordinate}: a Maven coordinate for a BOM
```
elharo: a Maven coordinate --> Maven coordinates

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -47,14 +80,14 @@ static Builder builder() {
 
   @AutoValue.Builder
   abstract static class Builder {
-    abstract Builder setBomCoordinate(@Nullable String value);
-    abstract Builder setMavenCoordinates(List<String> value);
-    abstract Builder setJarFileList(List<Path> value);
+    abstract Builder setBomCoordinate(@Nullable String coordinate);
+    abstract Builder setMavenCoordinates(List<String> coordinates);
```
elharo: It would be nice for this to be nullable too

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -47,14 +80,14 @@ static Builder builder() {
...
+    abstract Builder setJarFileList(List<Path> paths);
```
elharo: should probably be nullable

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
garrettjonesgoogle: Maven docs supports the singular:

https://maven.apache.org/pom.html#Maven_Coordinates

> **groupId, artifactId, version:**
> You will see these elements often. This trinity is used to compute the Maven coordinate of a specific project in time, demarcating it as a dependency of this project.

suztomo:
@elharo Fixed test. Can you check?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
elharo: This needs to be specified in the code itself. Really, though, this needs to throw an exception if someone tries to set more than one. Either that or refactor somehow so that not more than one can be set.

suztomo:
@elharo Removed use of RuntimeException. (except the code to be removed as part of #244)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -127,7 +140,7 @@ static ClassDumper create(List<Path> jarFilePaths) {
       if (!(constantAtNameAndTypeIndex instanceof ConstantNameAndType)) {
         // This constant_pool entry must be a CONSTANT_NameAndType_info
         // as specified https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.4.2
-        throw new RuntimeException(
+        throw new ClassFormatException(
```
suztomo: I read go/java-practices/verification-checks.

@elharo Does "Verify" make more sense to you here?
https://google.github.io/guava/releases/snapshot-jre/api/docs/com/google/common/base/Verify.html

elharo: I think ClassFormatException is more specific and does fit this case.

suztomo: Good. Then I keep ClassFormatException.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
suztomo: Elliotte and I checked the statement yesterday but that's the only document (I could find) that uses singular. Maybe it means

> This trinity is used to compute _(each of)_ the Maven coordinate of a specific project: _groupId, artifactId, and version_".

Maven Aether library uses plural for coordinates.
http://maven.apache.org/resolver/apidocs/org/eclipse/aether/artifact/DefaultArtifact.html#DefaultArtifact(java.lang.String)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
suztomo: As discussed yesterday, not using nullable for now.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
garrettjonesgoogle: Sad times for our pluralization options...

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReference.java
suztomo: Added example in the above.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -184,10 +187,9 @@ static SymbolTable scanClassSymbolTable(JavaClass javaClass) {
     return symbolTableBuilder.build();
   }
 
-  static MethodSymbolReference constantToMethodReference(ConstantMethodref constantMethodref,
-      ConstantPool constantPool, String sourceClassName) {
-    String classNameInMethodReference = constantMethodref.getClass(constantPool);
-    int nameAndTypeIndex = constantMethodref.getNameAndTypeIndex();
+  static ConstantNameAndType constantNameAndTypeFromConstantCP(
```
netdpb: You could just call this "constantNameAndType", since we know it's a `ConstantCP` (whatever that is) from the parameters.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -197,7 +199,14 @@ static MethodSymbolReference constantToMethodReference(ConstantMethodref constan
               + nameAndTypeIndex
               + ". This class file is not compliant with CONSTANT_Methodref_info specification");
```
netdpb: Now we don't know whether it's a method or field or what, so can this message be more general?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassSymbolReference.java
```diff
@@ -18,10 +18,15 @@
...
+/**
+ * A symbolic reference to {@code targetClass} referenced from {@code sourceClass}.
+ */
 @AutoValue
-abstract class ClassSymbolReference {
-  abstract String getTargetClassName();
-  abstract String getSourceClassName();
+abstract class ClassSymbolReference implements SymbolReference {
+  @Override
+  public abstract String getTargetClassName();
```
netdpb: You don't need to redeclare these abstract methods here.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -189,26 +192,28 @@ private static void printStaticLinkageErrors(
 
   /**
    * Finds linkage errors in the input classpath and generates a static linkage check report.
+   *
+   * @return a static linkage check report for the input class path
```
netdpb: This just repeats what's in the summary, so you can leave it out.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -219,15 +224,21 @@ private JarLinkageReport findInvalidReferences(Path jarPath, SymbolTable jarSymb
...
 
   /**
-   * Scans class files in the jar file and creates {@link SymbolTable} populated with class names
+   * Scans class files in the jar file and returns a {@link SymbolTable} populated with class names
    * defined within the jar and symbolic references to other classes outside the jar.
    *
-   * @param jarFilePath absolute path to a jar file.
-   * @return Table containing symbol references and classes defined in the jar file.
+   * @param jarFilePath absolute path to a jar file
+   * @return symbol references and classes defined in the jar file
```
netdpb: This `@return` duplicates information already in the summary, so you can leave it out if you want.

And I see it hasn't been updated along with the summary. That's even more reason to remove it. It's hard to remember to update information everywhere it's repeated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
netdpb: If the `SymbolTable` builder gave you an `ImmutableSet.Builder` for this, you could do

```
symbolTableBuilder.definedClassNamesBuilder()
    .add(sourceClassName)
    .addAll(listInnerClassNames(javaClass));
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FieldSymbolReference.java
```diff
@@ -18,15 +18,21 @@
 
 import com.google.auto.value.AutoValue;
 
-/** A symbolic reference to a field of {@code targetClass} referenced from {@code sourceClass}. */
+/**
+ * A symbolic reference to a field of {@code targetClass} referenced from {@code sourceClass}.
+ * */
```
netdpb: Fix end of Javadoc.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -32,13 +32,43 @@
...
+ *     Linkage Checker: Input</a>
+ */
 @AutoValue
 abstract class StaticLinkageCheckOption {
   // TODO(suztomo): Add option to specify entry point classes
 
+  /**
+   * Returns a Maven coordinate for a BOM if specified; otherwise null.
+   */
   @Nullable abstract String getBomCoordinate();
```
netdpb: Should this be called "getBOMCoordinates"?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -32,13 +32,43 @@
...
+
+  /**
+   * Returns Maven coordinates if specified; otherwise an empty list.
+   */
   abstract ImmutableList<String> getMavenCoordinates();
+
+  /**
+   * Returns list of absolute paths for jar files in the filesystem if specified; otherwise an empty
```
netdpb: You can leave out "list of" since we know it's a list from its return type.

`Returns the absolute paths for jar files...`

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -32,13 +32,43 @@
...
+   * list.
+   */
   abstract ImmutableList<Path> getJarFileList();
+
+  /**
+   * Returns flag to report only reachable linkage errors.
```
netdpb: How about

`Returns {@code true} if only reachable linkage errors should be reported.`

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolsInFile.java
```diff
@@ -20,21 +20,34 @@
 import com.google.common.collect.ImmutableSet;
 
 /**
- * A table of symbolic references and defined classes in a jar file or class file. The
- * symbolic references are from constant pool of class files.
+ * Symbolic references and defined classes in a file (jar file or class file). The
```
netdpb: This could be used to hold references from more than one file, right? There's nothing here that specifies a file. It's just a collection of class names and symbol references. And nothing in this class actually constrains these to come from constant pools. Let's just say this is a collection of symbol references and classes? (I'm still wondering whether those two things belong in the same place, but that can wait.)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolsInFile.java
```diff
@@ -22,7 +22,7 @@
 /**
  * Symbolic references and defined classes in a file (jar file or class file). The
  * symbolic references are from constant pool of class file(s). The defined class names are
- * from the class defined as top-level in a class file and inner classes in it.
+ * from the class defined as top-level in a class file and its inner classes.
```
netdpb: They're really _nested_ classes, not just inner classes, right?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReference.java
```diff
@@ -17,10 +17,19 @@
 package com.google.cloud.tools.opensource.classpath;
 
 /**
- * A reference to a symbol of {@code targetClass} from {@code sourceClass}.
+ * A reference to a symbol of {@code targetClass} from {@code sourceClass}. The values of the class
+ * names are fully-qualified with '.' as separator and '$' for inner class. For example {@code
```
netdpb: nested, not inner.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReference.java
netdpb: Is it the FQCN or the binary name?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReference.java
```diff
@@ -27,12 +27,12 @@
 interface SymbolReference {
 
   /**
-   * Fully-qualified class name of source of the reference
+   * Returns fully-qualified class name of the source class of the reference.
```
netdpb: You need "the" after "Returns". Below too.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -216,71 +213,49 @@ StaticLinkageCheckReport findLinkageErrors() {
...
     JarLinkageReport.Builder reportBuilder = JarLinkageReport.builder().setJarPath(jarPath);
 
     // TODO(suztomo): implement validation for field, method and class references in the table
     return reportBuilder.build();
   }
 
   /**
-   * Scans class files in the jar file and returns a {@link SymbolsInFile} populated with class names
-   * defined within the jar and symbolic references to other classes outside the jar.
+   * Scans class files in the jar file and returns a {@link SymbolsInFile} populated with symbolic
```
netdpb: "symbol references", not "symbolic references".

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -213,43 +213,31 @@ StaticLinkageCheckReport findLinkageErrors() {
...
+   * Scans class files in the jar file and returns a {@link SymbolReferenceSet} populated with
+   * symbolic references.
    *
    * @param jarFilePath absolute path to a jar file
    * @return symbol references and classes defined in the jar file
    */
-  static SymbolsInFile scanSymbolReferences(Path jarFilePath) {
-    Preconditions.checkState(
+  static SymbolReferenceSet scanSymbolReferencesInJar(Path jarFilePath) {
+    Preconditions.checkArgument(
```
netdpb: FYI I find it more readable to statically import any `Preconditions` static method I use. (Same with `Truth.assertThat()`.)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReferenceSet.java
```diff
@@ -20,14 +20,14 @@
 import com.google.common.collect.ImmutableSet;
 
 /**
- * Symbolic references in a file (jar file or class file). The references are from constant pool of
- * class file(s).
+ * A set of symbolic references in a file (jar file or class file). The references are from constant
```
netdpb: I'd say "symbol references" instead of "symbolic references". They're references to symbols, not references that are symbolic.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReferenceSet.java
```diff
@@ -52,6 +53,14 @@ static Builder builder() {
     abstract ImmutableSet.Builder<ClassSymbolReference> classReferencesBuilder();
     abstract ImmutableSet.Builder<MethodSymbolReference> methodReferencesBuilder();
     abstract ImmutableSet.Builder<FieldSymbolReference> fieldReferencesBuilder();
-    abstract SymbolsInFile build();
+
+    abstract SymbolReferenceSet build();
+
+    Builder merge(SymbolReferenceSet other) {
```
netdpb: Maybe "add" instead of "merge"? Up to you.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
netdpb: If you really want to make that happen, you could use `@AutoOneOf` (it's part of the AutoValue package).

But setting more than one flag is a flag validation problem, not programmer error. I think `parseArguments()` below should fail and report an error to the user if they specify an invalid combination of options.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -102,29 +102,28 @@ public static void main(String[] arguments)
...
-    Formatter formatter = new Formatter();
-    formatter.format(
-        "There were %,d unresolved method references from the jar file(s):\n", count);
-    for (FullyQualifiedMethodSignature methodReference : sortedUnresolvedMethodReferences) {
-      formatter.format(
-          "Class: '%s', method: '%s' with descriptor %s\n",
-          methodReference.getClassName(),
-          methodReference.getMethodSignature().getMethodName(),
-          methodReference.getMethodSignature().getDescriptor());
+  private static void printStaticLinkageReport(StaticLinkageCheckReport report) {
```
netdpb: This looks like it could become a method on `StaticLinkageCheckReport` instead. And maybe the stuff in the loop could become a method on `JarLinkageReport`.

suztomo: I prefer to keep these print-related code in one place. It can easily controlling indent in this way.
Is this ok with you?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Removed.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: I'd keep this for-loop because `findInvalidReferences` (currently empty implementation) will have `throws IOException, ClassNotFoundException` as it reads files. Functions with checked exception are not good with lambda so I keep this as for-loop.

(Originally `scanSymbolReferencesInJar` and `findInvalidReferences` were used with stream.)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReference.java
suztomo: It's binary name. Added clarification in comment.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReference.java
suztomo: Added

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReferenceSet.java
suztomo: Updated them.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReferenceSet.java
suztomo: `Merge` makes more sense to me as the parameter is the same class.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Updated to use static import.
`assertThat` would conflict hamcrest's Assert. So I'd keep as they're.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Removed `@return`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
suztomo: Now `parseArguments` throws `IllegalArgumentException` if the combination is invalid.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Renamed to `constantNameAndType`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReferenceSet.java
netdpb: You could also do "addAll", just like Set has "addAll". Totally up to you.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
netdpb: Sure.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -38,7 +38,7 @@
  *
  * <ul>
  *   <li>{@code bomCoordinates}: Maven coordinates for a BOM
- *   <li>{@code mavenCoordinates}: list of the coordinates of Maven artifacts to check
+ *   <li>{@code mavenCoordinatesList}: list of the coordinates of Maven artifacts to check
```
netdpb: This is a small thing, but in general you don't need to name methods or parameters with "list" if they're of a List type. So-called ["Hungarian notation"](https://en.wikipedia.org/wiki/Hungarian_notation) isn't useful in a statically typed language.

This is why I often don't bother with "list of" phrases in documentation either.

It also makes it easier to switch between lists and sets later on, since that doesn't change the meaningful semantics at the documentation level.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -61,16 +61,15 @@
    * Returns list of coordinates of Maven artifacts if specified; otherwise an empty list. Example
    * element: {@code com.google.cloud:google-cloud-bigtable:0.66.0-alpha}
    */
-  abstract ImmutableList<String> getMavenCoordinates();
+  abstract ImmutableList<String> getMavenCoordinatesList();
```
netdpb: Can we better distinguish this from the BOM option? That's Maven coordinates too, just for a BOM artifact instead of a list of JAR artifacts.

The three options for specifying a class path are

1. one BOM artifact
2. a list of non-BOM JAR artifacts
3. a list of paths to jar files

What about

1. `--bom`, `getBomArtifact()`
2. `--jar_artifacts`, `getJarArtifacts()`
3. `--jar_paths`, `getJarPaths()`

If we want to be clever (which we may not), we could combine (2) and (3) into `--jars` and figure out at runtime which are paths and which are Maven coordinates.

```
static_linkage_checker --bom=some.group:some-bom:1.0
static_linkage_checker --jars=some.group:some-artifact:1.0,/path/to/a.jar,/path/to/another.jar,other.group:other-artifact:2.1
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
netdpb: You can make your `builder()` method above prime everything with empty lists:

```
builder() {
  return new AutoValue_StaticLinkageCheckOption.Builder()
      .setMavenCoordinatesList(ImmutableList.of())
      .setJarFileList(ImmutableList.of());
}
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
suztomo: I'd like to keep these three separated as designed in [Static Linkage Checker: Input](https://github.com/GoogleCloudPlatform/cloud-opensource-java/tree/master/dependencies#input).

I updated the properties as "bom", "artifacts", "jarFiles" by removing "coordinates" and "list". Do they look better?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
suztomo: Updated the `builder()`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolReferenceSet.java
suztomo: That sounds better. Now it's `addAll`.

suztomo:
@netdpb Applied code review and responded to comments. PTAL. Do they look good now?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -51,48 +51,53 @@
   // TODO(suztomo): Add option to specify entry point classes
 
   /**
-   * Returns Maven coordinates for a BOM if specified; otherwise null. Example value: {@code
+   * Returns the Maven coordinates for a BOM if specified; otherwise null. Example value: {@code
    * com.google.cloud:cloud-oss-bom:pom:1.0.0-SNAPSHOT}
    */
   @Nullable
-  abstract String getBomCoordinates();
+  abstract String getBom();
```
netdpb: It might be nice (in a later PR!) to have a class representing a Maven coordinate tuple so we're not passing strings around.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -51,48 +51,53 @@
...
+        .setArtifacts(ImmutableList.of())
+        .setJarFiles(ImmutableList.of());
   }
 
   @AutoValue.Builder
   abstract static class Builder {
-    abstract Builder setBomCoordinates(@Nullable String coordinates);
-    abstract Builder setMavenCoordinatesList(List<String> coordinates);
-    abstract Builder setJarFileList(List<Path> paths);
+    abstract Builder setBom(@Nullable String coordinates);
```
netdpb: Do we need the setter to take null? Or should we just not call it if there's no BOM?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -105,14 +110,14 @@ static StaticLinkageCheckOption parseArguments(String[] arguments) throws ParseE
     ImmutableList.Builder<String> mavenCoordinates = ImmutableList.builder();
     try {
       CommandLine commandLine = parser.parse(options, arguments);
-      if ((commandLine.hasOption("b") && commandLine.hasOption("c"))
-          || (commandLine.hasOption("c") && commandLine.hasOption("j"))
+      if ((commandLine.hasOption("b") && commandLine.hasOption("a"))
```
netdpb: How about something like 
```
if (Stream.of("b", "a", "j").filter(commandLine::hasOption).count() > 1) {
```

Better yet, you could build the `StaticLinkageCheckOption` object and then throw if more than one of the properties is not null or empty. Then you're not repeating the option letters again; and it's more meaningful a check.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -105,14 +110,14 @@ static StaticLinkageCheckOption parseArguments(String[] arguments) throws ParseE
...
+          || (commandLine.hasOption("a") && commandLine.hasOption("j"))
           || (commandLine.hasOption("j") && commandLine.hasOption("b"))) {
         throw new IllegalArgumentException(
             "One of BOM, Maven coordinates, or jar files can be specified");
       }
-      if (commandLine.hasOption("c")) {
-        String mavenCoordinatesOption = commandLine.getOptionValue("c");
+      if (commandLine.hasOption("a")) {
+        String mavenCoordinatesOption = commandLine.getOptionValue("a");
         mavenCoordinates.addAll(Arrays.asList(mavenCoordinatesOption.split(",")));
```
netdpb: Consider using [`Splitter`](https://github.com/google/guava/wiki/StringsExplained#splitter), which is safer.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -144,10 +144,10 @@ private static void printStaticLinkageReport(StaticLinkageCheckReport report) {
     ImmutableList.Builder<Path> jarFileBuilder = ImmutableList.builder();
 
     // TODO(suztomo): add logic to convert Maven BOM to list of Maven coordinates as per README.md
-    if (!linkageCheckOption.getJarFileList().isEmpty()) {
-      jarFileBuilder.addAll(linkageCheckOption.getJarFileList());
-    } else if (!linkageCheckOption.getMavenCoordinatesList().isEmpty()) {
-      for (String mavenCoordinates : linkageCheckOption.getMavenCoordinatesList()) {
+    if (!linkageCheckOption.getJarFiles().isEmpty()) {
```
netdpb: You don't have to bother to check for empty. `addAll(empty)` is a no-op, and looping over an empty collection is also a no-op.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOptionTest.java
```diff
@@ -30,27 +30,29 @@ public void parseCommandLineOptions_shortOptions_bom() throws ParseException {
     };
     StaticLinkageCheckOption parsedOption = StaticLinkageCheckOption.parseArguments(arguments);
 
-    Assert.assertEquals("abc.com:dummy:1.2", parsedOption.getBomCoordinates());
+    Assert.assertEquals("abc.com:dummy:1.2", parsedOption.getBom());
+    Assert.assertTrue(parsedOption.getArtifacts().isEmpty());
```
netdpb: `Truth.assertThat(parsedOption.getArtifacts()).isEmpty()` will give you a better error message on failure.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
netdpb: That's okay. But I think you can statically import a method from more than one class, and Java will find the overload that matches the argument.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
suztomo: No it was unnecessary. (I thought it's convention in AutoValue but it's not)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
suztomo: Applied the stream-filter-count. That's much cleaner.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
suztomo: Now it uses Splitter.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Removed unnecessary if-statements.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOptionTest.java
suztomo: Changed to use Truth.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
netdpb: They mean two different things.

1.  `@Nullable Foo getFoo()` on the value type means that the property may be `null`.
2.  `Builder foo(@Nullable foo);` on the builder type means that it's okay to pass `null` to the builder.

(2) doesn't make sense unless (1) is true, but even if `foo` is nullable, you could say that it's an error to _set it to `null`_. Instead, just don't call the method if you want it to be `null`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -110,20 +112,19 @@ static StaticLinkageCheckOption parseArguments(String[] arguments) throws ParseE
...
     try {
       CommandLine commandLine = parser.parse(options, arguments);
-      if ((commandLine.hasOption("b") && commandLine.hasOption("a"))
-          || (commandLine.hasOption("a") && commandLine.hasOption("j"))
-          || (commandLine.hasOption("j") && commandLine.hasOption("b"))) {
+      if (Stream.of('b', 'a', 'j').filter(commandLine::hasOption).count() > 1) {
         throw new IllegalArgumentException(
             "One of BOM, Maven coordinates, or jar files can be specified");
       }
+      Splitter commaSplitter = Splitter.on(",");
```
netdpb: This could even be a constant `COMMA_SPLITTER` if you want.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -110,20 +112,19 @@ static StaticLinkageCheckOption parseArguments(String[] arguments) throws ParseE
...
       if (commandLine.hasOption("a")) {
         String mavenCoordinatesOption = commandLine.getOptionValue("a");
-        mavenCoordinates.addAll(Arrays.asList(mavenCoordinatesOption.split(",")));
+        mavenCoordinates.addAll(commaSplitter.split(mavenCoordinatesOption));
       }
       if (commandLine.hasOption("j")) {
         String jarFiles = commandLine.getOptionValue("j");
         List<Path> jarFilesInArguments =
-            Arrays.stream(jarFiles.split(","))
+            Streams.stream(commaSplitter.split(jarFiles))
```
netdpb: Or `commaSplitter.splitToList(jarFiles).stream()`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Good to know that. I confirmed it works:

```
import static org.junit.Assert.assertThat;
import static org.hamcrest.CoreMatchers.is;
import static com.google.common.truth.Truth.assertThat;

import com.google.common.collect.ImmutableList;
import org.junit.Test;

public class FooTest {
  @Test
  public void testStaticImports() {
    assertThat("This is test", "foo", is("foo"));
    assertThat(ImmutableList.of()).isEmpty();
  }
}
```

PR #252, opened 2018-11-16
fix some warnings, mostly about generics

elharo:
@netdbp also unused code and line length


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -248,6 +248,7 @@ private boolean resolveFullDependencies() {
    * @throws DependencyResolutionException when there is a problem in resolving dependency. This
    *     happens only when graphTraversalOption is FULL_DEPENDENCY or FULL_DEPENDENCY_WITH_PROVIDED.
    */
+  @SuppressWarnings("unchecked")
```
netdpb: What's unchecked? Please add a comment explaining why it's safe.

Is there a smaller scope in which you can apply this?

elharo: It's clone. Done.

PR #257, opened 2018-11-16
Checking to reproduce freemarker error with AbstractListMultimap

suztomo:
No need to approve.


suztomo:
CircleCI succeeds now without explicit cast.

PR #266, opened 2018-11-21
JLBP-1: Rewording 'scrutinize deps' recommendation

### LINKED ISSUE #95: Clarify JLBP-1 Scrutinize all dependency additions
I'm having trouble parsing this:

> Scrutinize all dependency additions. Check the result of mvn dependency:tree (after running mvn install -DskipTests to build the library) to see which transitive dependencies are added by just adding a single dependency to your own library, and if you require all of the transitive dependencies

It's also unclear how exactly this applies to non-Maven projects. I think I get the intent, but let's take another pass at the language used to describe this.

garrettjonesgoogle:
Fixes #95

### DIFF
--- a/library-best-practices/JLBP-1.md
+++ b/library-best-practices/JLBP-1.md
@@ -10,16 +10,18 @@ that library and that library's consumers.
 Some specific notes about minimizing dependencies:
 
 - Use the smallest scope possible. For example, auto-value doesn't
-  need to use compile scope, and can instead use compile-only,
+  need to use `compile` scope, and can instead use `compile-only`,
   since it doesn't need to appear on the classpath of consumers.
-  - All libraries used only for testing should have test scope
+  - All libraries used only for testing should have `test` scope
     (for example junit, mockito, and truth).
 
-- Scrutinize all dependency additions. Check the result of
-  `mvn dependency:tree` (after running `mvn install -DskipTests`
-  to build the library) to see which transitive dependencies are
-  added by just adding a single dependency to your own library,
-  and if you require all of the transitive dependencies.
+- Scrutinize all dependency additions. Whenever you add a new
+  dependency, check the full tree of transitive dependencies that
+  are pulled in as a result. If a large amount of transitive
+  dependencies are pulled in, consider a different direct dependency.
+  - Maven: Run `mvn dependency:tree` (after running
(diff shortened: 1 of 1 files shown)


library-best-practices/JLBP-1.md
```diff
@@ -10,16 +10,18 @@ that library and that library's consumers.
 Some specific notes about minimizing dependencies:
 
 - Use the smallest scope possible. For example, auto-value doesn't
-  need to use compile scope, and can instead use compile-only,
+  need to use `compile` scope, and can instead use `compile-only`,
   since it doesn't need to appear on the classpath of consumers.
-  - All libraries used only for testing should have test scope
+  - All libraries used only for testing should have `test` scope
```
elharo: Delete "All"; it's redundant

library-best-practices/JLBP-1.md
```diff
@@ -10,16 +10,18 @@ that library and that library's consumers.
...
     (for example junit, mockito, and truth).
 
-- Scrutinize all dependency additions. Check the result of
-  `mvn dependency:tree` (after running `mvn install -DskipTests`
-  to build the library) to see which transitive dependencies are
-  added by just adding a single dependency to your own library,
-  and if you require all of the transitive dependencies.
+- Scrutinize all dependency additions. Whenever you add a new
+  dependency, check the full tree of transitive dependencies that
+  are pulled in as a result. If a large amount of transitive
```
elharo: amount --> number

library-best-practices/JLBP-1.md
```diff
@@ -10,16 +10,18 @@ that library and that library's consumers.
...
+  dependencies are pulled in, consider a different direct dependency.
```
elharo: The answer may be no dependency at all; e.g. just reimplementing the small function you need

library-best-practices/JLBP-1.md
garrettjonesgoogle: Will do.

library-best-practices/JLBP-1.md
garrettjonesgoogle: Will do.

library-best-practices/JLBP-1.md
garrettjonesgoogle: Ok I will say something to that effect.

PR #287, opened 2018-12-17
Renaming 'dependencies' directory to 'dependency-tools'

### LINKED ISSUE #234: Move the static linkage checker to a more transparent directory name
The static linkage checker is currently located under https://github.com/GoogleCloudPlatform/cloud-opensource-java/tree/master/dependencies , which reveals very little about the tool. I think `/linkage-checker/` is a good candidate. I'm also thinking we don't want to paint ourselves into a corner with respect to checking static vs reflective errors; should we just call it the "Linkage Checker"? (Or maybe "Jar Linkage Checker" or "Classpath Linkage Checker"?)

suztomo:
Fixes #234 

Quoting from @garrettjonesgoogle 's comment

> In that case, I would suggest dependency-tools. The name dependencies makes it sound like it contains dependencies inside of it, and is very unintuitive.


suztomo:
Discussed with Elliotte. Not changing directory name for now.

PR #290, opened 2018-12-18
Remove some code duplication in StaticLinkageChecker.

netdpb:
Extract common code to look through superclasses.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
@@ -352,7 +352,8 @@ public class StaticLinkageChecker {
       Traverser.forTree(
           javaClass -> {
             try {
-              return ImmutableSet.of(javaClass.getSuperClass());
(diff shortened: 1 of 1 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -353,4 +336,25 @@ private boolean validateMethodReference(MethodSymbolReference methodReference) {
...
+    } catch (ClassNotFoundException ex) {
+      return ImmutableList.of();
+    }
+  }
+
+  private static final Traverser<JavaClass> SUPERCLASSES =
+      Traverser.forTree(
+          javaClass -> {
+            try {
+              return ImmutableSet.of(javaClass.getSuperClass());
```
suztomo: java.lang.Object's super class is null in BCEL API. I think that's why it raises NullPointerException.

netdpb: Yes.

Issue #293, opened 2018-12-19
Add reason and target class location in linkage errors for method references


suztomo:
Add reason and target class location in LinkageErrorMissingMethod

Check modifiers of target symbol and class.

suztomo:
This is implemented by #304

PR #294, opened 2018-12-19
To validate class reference modifier

### LINKED ISSUE #253: StaticLinkageChecker to check accessor to verify source class has valid access to the class symbol
Class symbol may have private access. In this case other class than the defining class should not be able to access the class.

Example private class: com.google.api.core.AbstractApiService$InnerService
AbstractApiService.java
https://github.com/googleapis/api-common-java/blob/master/src/main/java/com/google/api/core/AbstractApiService.java
has 

```
$ javap -cp src/test/resources/testdata/api-common-1.7.0.jar 'com.google.api.core.AbstractApiService$InnerService'
Compiled from "AbstractApiService.java"
class com.google.api.core.AbstractApiService$InnerService extends com.google.common.util.concurrent.AbstractService {
  final com.google.api.core.AbstractApiService this$0;
  protected void doStart();
  protected void doStop();
  com.google.api.core.AbstractApiService$InnerService(com.google.api.core.AbstractApiService, com.google.api.core.AbstractApiService$1);
  static void access$200(com.google.api.core.AbstractApiService$InnerService);
  static void access$300(com.google.api.core.AbstractApiService$InnerService);
  static void access$400(com.google.api.core.AbstractApiService$InnerService, java.lang.Throwable);
}
```

Fields and method references are to be handled in other issues.

suztomo:
To validate class reference modifier.

Fixes #253 

- `StaticLinkageChecker.checkLinkageErrorMissingClassAt` validates modifiers of target class.
- New interface `LinkageErrorWithReason` provides common fields for linkage errors: `reason` and `targetClassLocation`.
  The two field will help us when we diagnose why we get linkage errors
- The reason is represented as enum `LinkageErrorWithReason.Reason`
- LinkageErrorMissingField and LinkageErrorMissingClass implement the interface. Method reference is to follow the same (#293).

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -19,6 +19,7 @@ package com.google.cloud.tools.opensource.classpath;
 import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
@@ -305,9 +306,15 @@ class ClassDumper {
     return parameterTypes;
   }
 
-  /** Returns the jar file URL of a class in the class path. */
+  /**
+   * Returns the jar file URL of a class in the class path. Null if the information is unavailable.
+   */
   URL findClassLocation(String className) throws ClassNotFoundException {
     Class<?> clazz = loadClass(className);
+    if (clazz.getProtectionDomain().getCodeSource() == null) {
+      // javax.activation.SecuritySupport is known to return null here
+      return null;
+    }
     return clazz.getProtectionDomain().getCodeSource().getLocation();
   }
 
@@ -394,5 +401,15 @@ class ClassDumper {
       }
     }
     return javaClasses.build();
+  }
+
+  static boolean classesInSamePackage(String classNameA, String classNameB) {
+    Preconditions.checkArgument(classNameA.contains("."),
+        "Class name A does not have package");
+    Preconditions.checkArgument(classNameB.contains("."),
+        "Class name B does not have package");
+    String packageNameA = classNameA.substring(0, classNameA.lastIndexOf('.'));
+    String packageNameB = classNameB.substring(0, classNameB.lastIndexOf('.'));
+    return packageNameA.equals(packageNameB);
   }
 }
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -403,11 +403,17 @@ class ClassDumper {
     return javaClasses.build();
   }
 
+  /**
+   * Returns true if two class names (binary name JLS 13.1) have the same package.
+   */
   static boolean classesInSamePackage(String classNameA, String classNameB) {
-    Preconditions.checkArgument(classNameA.contains("."),
-        "Class name A does not have package");
-    Preconditions.checkArgument(classNameB.contains("."),
-        "Class name B does not have package");
+    if (classNameA.contains(".") != classNameB.contains(".")) {
+      return false;
+    }
+    if (classNameA.lastIndexOf('.') < 0 && classNameB.lastIndexOf('.') < 0) {
+      return true;
+    }
+    // The conditions above ensure that lastIndexOf returns non-negative number
     String packageNameA = classNameA.substring(0, classNameA.lastIndexOf('.'));
     String packageNameB = classNameB.substring(0, classNameB.lastIndexOf('.'));
     return packageNameA.equals(packageNameB);
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
@@ -67,6 +67,7 @@ public abstract class JarLinkageReport {
     }    
     for (LinkageErrorMissingClass missingClass : getMissingClassErrors()) {
       builder.append(indent + missingClass.getReference());
+      builder.append(" reason:" + missingClass.getReason());
       builder.append("\n");
     }
     for (LinkageErrorMissingMethod missingMethod : getMissingMethodErrors()) {
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
@@ -4,26 +4,41 @@ import java.net.URL;
 import javax.annotation.Nullable;
 
 /**
- * Interface to provide common fields for the different types of static linkage errors.
+ * Interface to provide common fields for different types of static linkage errors.
  */
 interface LinkageErrorWithReason {
 
   /**
-   * Returns the location of the target class in the field reference; null if the target class is
-   * not found in the class path.
+   * Returns the location of the target class in a symbol reference; null if the target class is
+   * not found in the class path or the source location is unavailable.
    */
   @Nullable
   URL getTargetClassLocation();
 
   /**
-   * Returns the reason why a symbol reference is detected as a linkage error.
+   * Returns the reason why a symbol reference is marked as a linkage error.
    */
   Reason getReason();
 
   /**
-   * Reason to distinguish the cause of a static linkage error
+   * Reason to distinguish the cause of a static linkage error against a symbol reference.
    */
   enum Reason {
-    TARGET_CLASS_NOT_FOUND, INVALID_ACCESS_MODIFIER, MISSING_MEMBER
+    /**
+     * The target class of the symbol reference is not found in the class path.
+     */
+    TARGET_CLASS_NOT_FOUND,
+
+    /**
+     * The access modifier (e.g., public or protected) does not allow the source of the symbol
+     * reference to use the target symbol.
+     */
+    INVALID_ACCESS_MODIFIER,
+
+    /**
+     * For a method or field reference, the member is not found in the target class in the class
+     * path.
+     */
+    MISSING_MEMBER
   }
 }
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
@@ -9,8 +9,8 @@ import javax.annotation.Nullable;
 interface LinkageErrorWithReason {
 
   /**
-   * Returns the location of the target class in a symbol reference; null if the target class is
-   * not found in the class path or the source location is unavailable.
+   * Returns the location of the target class in a symbol reference; null if the target class is not
+   * found in the class path or the source location is unavailable.
    */
   @Nullable
   URL getTargetClassLocation();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
@@ -1,0 +1,29 @@
+package com.google.cloud.tools.opensource.classpath;
+
+import java.net.URL;
+import javax.annotation.Nullable;
+
+/**
+ * Interface to provide common fields for the different types of static linkage errors.
+ */
+interface LinkageErrorWithReason {
+
+  /**
+   * Returns the location of the target class in the field reference; null if the target class is
+   * not found in the class path.
+   */
+  @Nullable
+  URL getTargetClassLocation();
+
+  /**
+   * Returns the reason why a symbol reference is detected as a linkage error.
+   */
+  Reason getReason();
+
+  /**
+   * Reason to distinguish the cause of a static linkage error
+   */
+  enum Reason {
+    TARGET_CLASS_NOT_FOUND, INVALID_ACCESS_MODIFIER, MISSING_MEMBER
+  }
+}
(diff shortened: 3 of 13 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -395,4 +402,20 @@ URL findClassLocation(String className) throws ClassNotFoundException {
...
+   * Returns true if two class names (binary name JLS 13.1) have the same package.
+   */
+  static boolean classesInSamePackage(String classNameA, String classNameB) {
+    if (classNameA.contains(".") != classNameB.contains(".")) {
+      return false;
+    }
+    if (classNameA.lastIndexOf('.') < 0 && classNameB.lastIndexOf('.') < 0) {
+      return true;
+    }
+    // The conditions above ensure that lastIndexOf returns non-negative number
```
netdpb: The last index of '.' has to be the same (even if negative) for this to return true, right? Could you replace lines 410 – 415 with

```java
if (classNameA.lastIndexOf('.') != classNameB.lastIndexOf('.')) {
  return false;
}
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Returns the location of the target class in a symbol reference; null if the target class is not
+   * found in the class path or the source location is unavailable.
+   */
+  @Nullable
+  URL getTargetClassLocation();
+
+  /**
+   * Returns the reason why a symbol reference is marked as a linkage error.
```
netdpb: Can you run google-java-format over all changed lines? I think this Javadoc would look different.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * The target class of the symbol reference is not found in the class path.
+     */
+    TARGET_CLASS_NOT_FOUND,
+
+    /**
+     * The access modifier (e.g., public or protected) does not allow the source of the symbol
+     * reference to use the target symbol.
+     */
+    INVALID_ACCESS_MODIFIER,
```
netdpb: I think we want to track illegal access in general. That could be because of the modifiers applied to the symbol itself, or any of its enclosing types.

```java
class Foo {
  public void method() { ... }
}
```

`Foo.method()` is inaccessible outside its package even though it's `public`.

Even if that level of analysis isn't implemented yet, can we rename this enum value to `INACCESSIBLE` or something?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
```diff
@@ -0,0 +1,60 @@
...
+  Reason getReason();
+
+  /**
+   * Reason to distinguish the cause of a static linkage error against a symbol reference.
+   */
+  enum Reason {
+    /**
+     * The target class of the symbol reference is not found in the class path.
+     */
+    TARGET_CLASS_NOT_FOUND,
```
netdpb: The reasons all apply to the target symbol, right? We don't have to specify that it's the _target_ class that's not found. How about just `CLASS_NOT_FOUND`?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
```diff
@@ -0,0 +1,60 @@
...
+     * The access modifier (e.g., public or protected) does not allow the source of the symbol
+     * reference to use the target symbol.
+     */
+    INVALID_ACCESS_MODIFIER,
+
+    /**
+     * For a method or field reference, the member is not found in the target class in the class
+     * path.
+     */
+    MISSING_MEMBER
```
netdpb: Can we call this `SYMBOL_NOT_FOUND`? We use "not found" above for classes, and it's nice to be consistent. Also, constructors (which might not be found) are not technically members.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -303,11 +303,46 @@ JarLinkageReport generateLinkageReport(Path jarPath, SymbolReferenceSet symbolRe
...
+    }
+    // Top-level class can only be declared as public or package private while nested-class can be
+    // declared as private or protected as well.
+    // https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-ClassModifier
+    if (javaClass.isPrivate()) {
+      // Class reference within same file is allowed to access private class. However, such case
+      // is already filtered at errorsFromSymbolReferences.
+      return false;
+    }
+    if (javaClass.isProtected()) {
```
netdpb: You don't really need this case, since it's the same as the default access case.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -303,11 +303,46 @@ JarLinkageReport generateLinkageReport(Path jarPath, SymbolReferenceSet symbolRe
...
+  /**
+   * Returns true if the {@code javaClass} is accessible {@code from sourceClassName} in terms of
+   * the access modifiers in the {@code javaClass}.
+   *
+   * @see <a href="https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-ClassModifier">
+   *   JLS 8.1.1. Class Modifiers</a>
+   */
+  private static boolean isClassAccessibleFrom(JavaClass javaClass, String sourceClassName) {
+    if (javaClass.isPublic()) {
+      // If a class is public, then it is accessible from anywhere
```
netdpb: This analysis is wrong in the case of nested classes.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -305,9 +306,15 @@ private static FieldSymbolReference constantToFieldReference(
     return parameterTypes;
   }
 
-  /** Returns the jar file URL of a class in the class path. */
+  /**
+   * Returns the jar file URL of a class in the class path. Null if the information is unavailable.
+   */
   URL findClassLocation(String className) throws ClassNotFoundException {
     Class<?> clazz = loadClass(className);
+    if (clazz.getProtectionDomain().getCodeSource() == null) {
```
elharo: pull clazz.getProtectionDomain().getCodeSource() out to a local variable since you use it twice

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
```diff
@@ -67,6 +67,7 @@ public String toString() {
     }    
     for (LinkageErrorMissingClass missingClass : getMissingClassErrors()) {
       builder.append(indent + missingClass.getReference());
+      builder.append(" reason:" + missingClass.getReason());
```
elharo: no line break before this?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Fixed with recursively calling this for each enclosing class.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
suztomo: Updated to INACCESSIBLE

suztomo:
@elharo @netdpb Applied code review and responded to comments. PTAL.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Simplified logic

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
suztomo: As discussed, it turned out my setting for google-java-format was only for google3 project!

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Updated the logic with introducing variables for className{A, B}.lastIndexOf('.')

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -403,19 +404,27 @@ URL findClassLocation(String className) throws ClassNotFoundException {
     return javaClasses.build();
   }
 
+  /** Returns true if two class names (binary name JLS 13.1) have the same package. */
+  static boolean classesInSamePackage(String classNameA, String classNameB) {
+    // Because package name cannot have '.' at the beginning, we can use lastDotIndex=0 (that will
```
netdpb: This is fine.

Alternatively, you could just add 1, so that the substring includes the dot after the package if there is one, and is the empty string for the default package.

`int lastDotIndexA = classNameA.lastIndexOf('.') + 1;`

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
```diff
@@ -31,30 +29,24 @@
...
-     * The target class of the symbol reference is not found in the class path.
-     */
-    TARGET_CLASS_NOT_FOUND,
+    /** The target class of the symbol reference is not found in the class path. */
+    CLASS_NOT_FOUND,
 
     /**
-     * The access modifier (e.g., public or protected) does not allow the source of the symbol
-     * reference to use the target symbol.
+     * The access modifier (e.g., public or protected) to the target symbol or its enclosing
```
netdpb: How about:

```java
/**
 * The symbol is inaccessible to the source.
 *
 * <p>If the source is in a different package, the symbol or one of its enclosing types
 * is not public. If the source is in the same package, the symbol or one of its enclosing
 * types is private.
 */
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -319,30 +321,37 @@ JarLinkageReport generateLinkageReport(Path jarPath, SymbolReferenceSet symbolRe
...
+  private boolean isClassAccessibleFrom(JavaClass javaClass, String sourceClassName)
+      throws ClassNotFoundException {
+    String targetClassName = javaClass.getClassName();
+    boolean publicOrSamePackage =
+        javaClass.isPublic() || ClassDumper.classesInSamePackage(targetClassName, sourceClassName);
+    if (javaClass.isNested()) {
+      // Nested class can be declared as private or protected, in addition to public and package
+      // private. Protected is treated same as package private.
+      // https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-ClassModifier
+      if (javaClass.isPrivate() || !publicOrSamePackage) {
```
netdpb: I think splitting `publicOrSamePackage` would make this easier to understand. And the way I think about the logic is maybe simpler?

1. If it's private, return false.
2. If it's public or in the same package as the source, return true iff the enclosing class is accessible.
3. Otherwise (not public and in a different package), return false.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/JarLinkageReportTest.java
```diff
@@ -99,9 +99,9 @@ public void testGetTotalErrorCount() {
   @Test
   public void testToString() {
     Assert.assertEquals("c (3 errors):\n" + 
-        "  ClassSymbolReference{sourceClassName=ClassB, targetClassName=ClassA} reason:TARGET_CLASS_NOT_FOUND\n" +
+        "  ClassSymbolReference{sourceClassName=ClassB, targetClassName=ClassA} reason:CLASS_NOT_FOUND\n" +
```
netdpb: Please split these strings so they're no more than 100 character lines.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: That's also nice.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorWithReason.java
suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Updated. Does this look good now?

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/JarLinkageReportTest.java
suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
netdpb: I think this can be simpler if you don't switch on `javaClass.isNested()` at the top. If you take that away, and treat enclosingClassName() returning null as meaning you're at the top level, then you're left with:

```
if private:
  return false
if public or same package:
  if has enclosing class:
    return isAccessibleFrom(enclosing class, source class)
  else:
    return true
return false
```

suztomo: Updated in that way.

suztomo:
@netdpb @elharo Applied review. PTAL.

PR #298, opened 2018-12-21
Generic Error Class LinkageErrorOnReference

### LINKED ISSUE #295: Consolidate LinkageErrorMissingXXX classes into generic one
Elliotte's suggestion that we may not need three classes: LinkageErrorMissingClass. LinkageErrorMissingMethod, and LinkageErrorMissingField.

I thought about this further. Because the difference among these three classes are only for the types of references held by the error, we can use generics to abstract the difference, while keeping type safety in the report.

# Error class

```
@AutoValue
abstract class LinkageErrorOnReference<T extends SymbolReference> {

  /**
   * Returns a symbol reference that caused this linkage error.
   */
  abstract T getReference();

  /**
   * Returns the location of the target class in a symbol reference; null if the target class is not
   * found in the class path or the source location is unavailable.
   */
  @Nullable
  abstract URL getTargetClassLocation();

  /**
   * Returns the reason why the symbol reference is marked as a linkage error.
   */
  abstract Reason getReason();
  ...
```

# Report class

From

```
@AutoValue
public abstract class JarLinkageReport {
  ...
  abstract ImmutableList<LinkageErrorMissingClass> getMissingClassErrors();
  abstract ImmutableList<LinkageErrorMissingMethod> getMissingMethodErrors();
  abstract ImmutableList<LinkageErrorMissingField> getMissingFieldErrors();

```

to

```
@AutoValue
public abstract class JarLinkageReport {
  ...
  abstract ImmutableList<LinkageErrorOnReference<ClassSymbolReference>> getMissingClassErrors();
  abstract ImmutableList<LinkageErrorOnReference<MethodSymbolReference>> getMissingMethodErrors();
  abstract ImmutableList<LinkageErrorOnReference<FieldSymbolReference>> getMissingFieldErrors();

```

suztomo:
The only difference among `LinkageErrorMissingClass`. `LinkageErrorMissingMethod`, and `LinkageErrorMissingField` are the underlying classes (`ClassSymbolReference`, `MethodSymbolReference` and `FieldSymbolReference`) that caused the linkage error.

This change is to delete the three error classes in favor of one error class with generics (`LinkageErrorOnReference<T extends SymbolReference>`). This reduces the number of classes, still keeping type safety.

Fixes #295


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
```diff
@@ -36,9 +36,11 @@
    */
   abstract ImmutableList<DependencyPath> getDependencyPaths();
 
-  abstract ImmutableList<LinkageErrorMissingClass> getMissingClassErrors();
-  abstract ImmutableList<LinkageErrorMissingMethod> getMissingMethodErrors();
-  abstract ImmutableList<LinkageErrorMissingField> getMissingFieldErrors();
+  abstract ImmutableList<LinkageErrorOnReference<ClassSymbolReference>> getMissingClassErrors();
```
elharo: Maybe this is just LinkageError? Does "OnReference" add any information?

suztomo: No it doesn’t add any information. I like “LinkageError”.

If I remember correctly you didn’t like class named “ XXXError” when it’s not inheriting Error.

garrettjonesgoogle: `LinkageError` seems like a problem because it is the same name as the standard library class: https://docs.oracle.com/javase/7/docs/api/java/lang/LinkageError.html

netdpb: `StaticLinkageError`?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorOnReference.java
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+package com.google.cloud.tools.opensource.classpath;
+
+import com.google.auto.value.AutoValue;
+import java.net.URL;
+import javax.annotation.Nullable;
+
+/**
+ * Static linkage error caused by a symbol reference. The symbol reference is one of the three
```
netdpb: You don't really need the second sentence here. If we add a fourth type, you don't want to have to remember to edit this doc.

PR #304, opened 2018-12-27
To avoid mix-up of class loading

### LINKED ISSUE #258: java.util.zip.ZipException: zip file is empty in tests
Not a major issue given context but it does suggest we're not logging this error correctly:

```
CLASSPATH component /Users/elharo/cloud-opensource-java/dependencies/target/test-classes/testdata/guava-26.0-jre.jar: java.util.zip.ZipException: zip file is empty
CLASSPATH component /Users/elharo/cloud-opensource-java/dependencies/target/test-classes/testdata/guava-26.0-jre.jar: java.util.zip.ZipException: zip file is empty
CLASSPATH component /Users/elharo/cloud-opensource-java/dependencies/target/test-classes/testdata/guava-26.0-jre.jar: java.util.zip.ZipException: zip file is empty
CLASSPATH component /Users/elharo/cloud-opensource-java/dependencies/target/test-classes/testdata/guava-26.0-jre.jar: java.util.zip.ZipException: zip file is empty
```

### LINKED ISSUE #299: Investigate linkage error on bigquery with "com.google.cloud.bigquery.$AutoValue_Labels"
https://storage.googleapis.com/cloud-opensource-java-dashboard/dashboard/target/dashboard/dashboard.html has

```
google-cloud-bigquery-1.56.0.jar (1 errors):
  Linked from: com.google.cloud:google-cloud-bigquery:1.56.0
  ClassSymbolReference{sourceClassName=com.google.cloud.bigquery.AutoValue_Labels, targetClassName=com.google.cloud.bigquery.$AutoValue_Labels}, reason: CLASS_NOT_FOUND, target class location not found
```

`ClassDumper.jarFilesToDefinedClasses` lists top-level classes first and that's because the target class is not found.

$ is valid character in Java identifier:

> The $ character should be used only in mechanically generated source code or, rarely, to access pre-existing names on legacy systems.

https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.8

Guava's ClassPath.getTopLevelClasses is not returning the class as top-level:

```

  @Test
  public void testGuavaTopLevelClass() throws Exception {
    // google-cloud-bigquery-1.56.0.jar contains com.google.cloud.bigquery.$AutoValue_Labels
    URL jarFileUrl =
        new URL(
            "file:///usr/local/google/home/suztomo/.m2/repository/com/google/cloud/google-cloud-bigquery/1.56.0/google-cloud-bigquery-1.56.0.jar");
    URL[] jarFileUrls = new URL[] {jarFileUrl};

    // Setting parent as null because we don't want other classes than this jar file
    URLClassLoader classLoaderFromJar = new URLClassLoader(jarFileUrls, null);

    // Leveraging Google Guava reflection as BCEL doesn't list classes in a jar file
    com.google.common.reflect.ClassPath classPath =

    ImmutableList<String> allClassNames =
        classPath.getAllClasses().stream()
            .map(classInfo -> classInfo.getName())
            .collect(toImmutableList());
    // This does not return class com.google.cloud.bigquery.$AutoValue_Labels
    // Guava only checks the existence of '$'
    // https://github.com/google/guava/blob/master/guava/src/com/google/common/reflect/ClassPath.java#L85
    ImmutableList<String> topLevelClassNames =
        classPath.getTopLevelClasses().stream()
            .map(classInfo -> classInfo.getName())
            .collect(toImmutableList());

    String class$AutoValue_Labels = "com.google.cloud.bigquery.$AutoValue_Labels";
    System.out.println("In all classes? " + allClassNames.contains(class$AutoValue_Labels)); // true
    System.out.println(
        "In top-level classes? " + topLevelClassNames.contains(class$AutoValue_Labels)); // false
  }

  @Test
  public void testLoadClassNameWithDollar() throws Exception {
    Path bigqueryPath =
        Paths.get(
            "/usr/local/google/home/suztomo/.m2/repository/com/google/cloud/google-cloud-bigquery/1.56.0/google-cloud-bigquery-1.56.0.jar");
    // google-cloud-bigquery-1.56.0.jar contains com.google.cloud.bigquery.$AutoValue_Labels
    ClassDumper classDumper = ClassDumper.create(ImmutableList.of(bigqueryPath));
    String class$AutoValue_Labels = "com.google.cloud.bigquery.$AutoValue_Labels";
    JavaClass javaClass = classDumper.loadJavaClass(class$AutoValue_Labels);
    // BCEL checks the attribute field of class file to determine isNested
    // https://github.com/apache/commons-bcel/blob/trunk/src/main/java/org/apache/bcel/classfile/JavaClass.java#L717
    System.out.println("Is this class nested? " + javaClass.isNested()); // false
  }
```

AutoValue has `@Memoized` annotation and it creates a class with `$AutoValue_XXXX` and `AutoValue_XXXX extends $AutoValue_XXXX` in the same package.

Filed a bug in Guava: https://github.com/google/guava/issues/3349

### LINKED ISSUE #302: StaticLinkageChecker: false positive on httpcore
StaticLinkageChecker shows following false positive errors, due to its lacking capability to follow interfaces:

```
httpcore-4.4.6.jar (2 errors):
...
  MethodSymbolReference{sourceClassName=org.apache.http.ssl.SSLContextBuilder$KeyManagerDelegate, targetClassName=javax.net.ssl.X509ExtendedKeyManager, methodName=getClientAliases, descriptor=(Ljava/lang/String;[Ljava/security/Principal;)[Ljava/lang/String;}, reason: SYMBOL_NOT_FOUND, target class location not found
  MethodSymbolReference{sourceClassName=org.apache.http.ssl.SSLContextBuilder$KeyManagerDelegate, targetClassName=javax.net.ssl.X509ExtendedKeyManager, methodName=getServerAliases, descriptor=(Ljava/lang/String;[Ljava/security/Principal;)[Ljava/lang/String;}, reason: SYMBOL_NOT_FOUND, target class location not found
```

httpcore 4.4.6 has `KeyManagerDelegate extends X509ExtendedKeyManager` ([link](https://github.com/apache/httpcomponents-core/blob/ebe5886ae3b164a64aa120990ddb6d8a4edc379d/httpcore/src/main/java/org/apache/http/ssl/SSLContextBuilder.java#L307))

The superclass `X509ExtendedKeyManager` is an abstract class in Java runtime, that implements `X509KeyManager` interface.

```
public abstract class X509ExtendedKeyManager implements X509KeyManager {
  protected X509ExtendedKeyManager() {
  }

  public String chooseEngineClientAlias(String[] var1, Principal[] var2, SSLEngine var3) {
    return null;
  }

  public String chooseEngineServerAlias(String var1, Principal[] var2, SSLEngine var3) {
    return null;
  }
}
```

In order to satisfy the interface, httpcore 4.4.6's KeyManagerDelegate implements the methods `getServerAliases` and `getClientAliases`. They're the ones complained by StaticLinkageChecker.

# Solution

`StaticLinkageChecker.validateMethodReference` to check interface.

```
  /**
   * Returns true if the method reference has a valid referent in the classpath via Java class
   * loader.
   */
  private boolean validateMethodReferenceByClassLoader(MethodSymbolReference methodReference) {
    String className = methodReference.getTargetClassName();
    String methodName = methodReference.getMethodName();
    try {
      Class<?>[] parameterTypes =
          classDumper.methodDescriptorToClass(methodReference.getDescriptor());
      Class<?> clazz = className.startsWith("[") ? Array.class : classDumper.loadClass(className);
      if ("<init>".equals(methodName)) {
        clazz.getConstructor(parameterTypes);
      } else if ("clone".equals(methodName) && clazz == Array.class) {
        // Array's clone method is not returned by getMethod
        // https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getMethod-java.lang.String-java.lang.Class...-
        return true;
      } else {
        try {
          clazz.getMethod(methodName, parameterTypes);
        } catch (NoSuchMethodException e) {
          // when a method is not found in the class itself, check it in interfaces (#302)
          for (AnnotatedType interfaceType : clazz.getAnnotatedInterfaces()) {
            String interfaceTypeName = interfaceType.getType().getTypeName();
            MethodSymbolReference interfaceMethodReference =
                methodReference.toBuilder().setTargetClassName(interfaceTypeName).build();
            boolean interfaceResult = validateMethodReference(interfaceMethodReference);
            if (interfaceResult) {
              return true;
            }
          }
          return false;
        }
      }
      return true;
    } catch (NoSuchMethodException | ClassNotFoundException ex) {
      return false;
    }
  }
```

test case

```

  @Test
  public void testFindMethodReferences_methodsOnInterface()
      throws IOException, URISyntaxException {
    List<Path> paths = ImmutableList.of(absolutePathOfResource("testdata/guava-26.0-jre.jar"));
    StaticLinkageChecker staticLinkageChecker =
        StaticLinkageChecker.create(false, paths, ImmutableSet.copyOf(paths));
    MethodSymbolReference methodOnInterface =
        MethodSymbolReference.builder()
            .setSourceClassName(StaticLinkageCheckReportTest.class.getName())
            .setTargetClassName("javax.net.ssl.X509ExtendedKeyManager")
            .setMethodName("getClientAliases")
            .setDescriptor("(Ljava/lang/String;[Ljava/security/Principal;)[Ljava/lang/String;")
            .build();

    Optional<StaticLinkageError<MethodSymbolReference>> detectedError = staticLinkageChecker
        .checkLinkageErrorMissingMethodAt(methodOnInterface);
    Truth.assertThat(detectedError.isPresent()).isFalse();
  }
```

### LINKED ISSUE #303: StaticLinkageChecker not to mix classes in CLASSPATH
In BCEL's `ClassPath.getInputStream`, `getClass().getClassLoader().getResourceAsStream()` supercedes `getClassFile()`.

```
    /**
     * Return stream for class or resource on CLASSPATH.
     *
     * @param name fully qualified file name, e.g. java/lang/String
     * @param suffix file name ends with suff, e.g. .java
     * @return input stream for file on class path
     */
    public InputStream getInputStream( final String name, final String suffix ) throws IOException {
        InputStream is = null;
        try {
            is = getClass().getClassLoader().getResourceAsStream(name + suffix); // may return null
        } catch (final Exception e) {
            // ignored
        }
        if (is != null) {
            return is;
        }
        return getClassFile(name, suffix).getInputStream();
    }
```

This makes classes mixed from the input class path and the class path of the StaticLinkageChecker.

# Failed attempt

I thought having independent class loader for BCEL class would solve the issue but it hit ClassCastException between two classes with same name but from different class loader.

```
    URL bcelJarUrl = ClassPath.class.getProtectionDomain().getCodeSource().getLocation();
    URLClassLoader bcelClassLoader =
        new URLClassLoader(new URL[] { bcelJarUrl }, extensionClassLoader);

    ClassPath classPath;
    try {
      System.out.println("Loading ClassPath from class loader: " + classLoaderFromJars);
      Class<?> c = bcelClassLoader.loadClass(ClassPath.class.getName());
      System.out.println("Class loaded: " + c);
      Object obj = c.getConstructor(String.class).newInstance(pathAsString);
      System.out.println("Class instantiated: " + obj);
      classPath = (ClassPath) obj;
      System.out.println("Assigned it to classPath");
    } catch (ClassNotFoundException
        | NoSuchMethodException
        | IllegalAccessException
        | InstantiationException
        | InvocationTargetException ex) {
      System.out.println("Class not found" + ex);
      throw new IllegalArgumentException("failed", ex);
    }

```

```
java.lang.ClassCastException: org.apache.bcel.util.ClassPath cannot be cast to org.apache.bcel.util.ClassPath
```

# Solution

Creating our own behavior of `ClassPath.getInputStream` would solve the issue.

```
public class MyClassPath extends ClassPath {

  public MyClassPath(final String classPath) {
    super(classPath);
  }

  @Override
  public InputStream getInputStream( final String name, final String suffix ) throws IOException {
    return getClassFile(name, suffix).getInputStream();
  }
}
```

As a side-effect, built-in classes stop being loaded via BCEL API.

```
expected to be empty
but was: [
  MethodSymbolReference{sourceClassName=org.apache.http.impl.client.DefaultRequestDirector, targetClassName=org.apache.http.NoHttpResponseException, methodName=setStackTrace, descriptor=([Ljava/lang/StackTraceElement;)V}, reason: SYMBOL_NOT_FOUND, target class location not found,
  MethodSymbolReference{sourceClassName=org.apache.http.impl.execchain.RetryExec, targetClassName=org.apache.http.NoHttpResponseException, methodName=setStackTrace, descriptor=([Ljava/lang/StackTraceElement;)V}, reason: SYMBOL_NOT_FOUND, target class location not found]

	at com.google.cloud.tools.opensource.classpath.StaticLinkageCheckerTest.testFindInvalidReferences_selfReferenceFromAbstractClassToInterface(StaticLinkageCheckerTest.java:156)

```

`setStackTrace` is a method in Throwable in Java and should not be reported as linkage errors.

### LINKED ISSUE #305: StaticLinkageChecker to confirm modifiers of fields and target class
StaticLinkageChecker to confirm modifiers of fields and target class

suztomo:
Among other things, this change is to avoid mix-up of class loading between the class path of the subject during static linkage check and the class path of the tool itself.

- `LinkageCheckClassPath` is introduced to properly distinguish the input class path of the subject and the class path of the tool itself.
  There was a bug where Guava 26 was unexpectedly available for the class path of static linkage check.
  Test case `StaticLinkageCheckerTest.testFindClass_guavaClassShouldNotBeAddedAutomatically`
  It turned out that `testdata/guava-26.0-jre.jar` was empty file. It's replaced with guava-23.5-jre.jar (Fixes #258)
- `StaticLinkageChecker.checkLinkageErrorMissingMethodAt` to properly separate the input class path and the class path of the tool itself
  Fixes #303.
  The method takes interfaces into account. (Fixes #302)
  - A genuine static linkage error is found at grpc-core ([link](https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/303#issuecomment-450213507))
  - Removed unnecessary methods: `validateMethodReference`, `validateMethodReferenceByBcelRepository`, `validateMethodReferenceByClassLoader`
- `ClassDumper.findClassLocation` to use `Path` rather than "`URL` via `ClassLoader`"
  This is to avoid unnecessary ClassNotFoundException caused by superclasses
- Not to rely on Guava's getTopLevelClasses
  It's because the method does not return top level classes with dollar ($) in its name.
  Fixes #299 
- Google's Truth extension Truth8 is added to use Java 8 optional subjects in unit tests.
- `StaticLinkageChecker.isMemberAccessibleFrom` to validate whether a field or method is accessible from source class.
  Fixes #305 #293


dependencies/pom.xml
```diff
@@ -81,6 +81,12 @@
       <version>0.42</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.truth.extensions</groupId>
+      <artifactId>truth-java8-extension</artifactId>
```
suztomo: Extension for Java8's optional.
https://google.github.io/truth/extension

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -62,45 +59,46 @@
...
   static ClassDumper create(List<Path> jarFilePaths) throws IOException {
-    // Creates classpath in the same order as inputClasspath for BCEL API
-    String pathAsString =
-        jarFilePaths.stream().map(Path::toString).collect(Collectors.joining(File.pathSeparator));
-    ClassPath classPath = new ClassPath(pathAsString);
-    SyntheticRepository syntheticRepository = SyntheticRepository.getInstance(classPath);
+    URL[] jarFileUrls = new URL[jarFilePaths.size()];
+    for (int i = 0; i < jarFilePaths.size(); ++i) {
+      jarFileUrls[i] = jarFilePaths.get(i).toUri().toURL();
+    }
```
suztomo: for-loop turned out to be simpler than lambda here. (MalformedURLException extends IOException)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -378,48 +417,12 @@ private boolean validateMethodReferenceByClassLoader(MethodSymbolReference metho
...
    * Returns the target class and its superclasses in order (with {@link Object} last). If any can't
    * be found, the list stops with the previous one.
    */
-  private Iterable<JavaClass> getClassAndSuperClasses(String targetClassName) {
-    try {
-      return SUPERCLASSES.breadthFirst(classDumper.loadJavaClass(targetClassName));
-    } catch (ClassNotFoundException ex) {
-      return ImmutableList.of();
-    }
+  private Iterable<JavaClass> getClassAndSuperClasses(JavaClass targetClass) {
```
suztomo: Taking JavaClass as argument makes it easy for callers to distinguish whether the class is in class path or not.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
```diff
@@ -198,4 +210,35 @@ public void testEnclosingClassNames() {
...
+  public void testClassToDefiningJarFile_firstJarFileWins() throws URISyntaxException, IOException {
+    Path firestore65 = absolutePathOfResource("testdata/google-cloud-firestore-0.65.0-beta.jar");
+    Path firestore66 = absolutePathOfResource("testdata/google-cloud-firestore-0.66.0-beta.jar");
+
+    // This class exists in both jar files
+    String grpcClass = "com.google.cloud.firestore.spi.v1beta1.GrpcFirestoreRpc";
+
+    ImmutableMap<String, Path> classToJar65First =
+        ClassDumper.classToDefiningJarFile(ImmutableList.of(firestore65, firestore66));
+    Truth.assertThat((Object) classToJar65First.get(grpcClass)).isEqualTo(firestore65);
```
suztomo: Cast to Object is to avoid ambiguity between `Object` and `Iterable<?>`.

dependencies/pom.xml
elharo: This should probably become part of core truth in 1.0.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -62,46 +63,52 @@
...
-        return jarPath.toUri().toURL();
-      } catch (MalformedURLException ex) {
-        throw new IllegalArgumentException("Jar file " + jarPath + " was not converted to URL",
-            ex);
-      }
-    }).toArray(URL[]::new);
-    URLClassLoader classLoaderFromJars =
-        new URLClassLoader(jarFileUrls, ClassLoader.getSystemClassLoader());
+    URL[] jarFileUrls = new URL[jarFilePaths.size()];
+    for (int i = 0; i < jarFilePaths.size(); ++i) {
```
elharo: i++ because it's more common

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -416,6 +443,16 @@ static boolean classesInSamePackage(String classNameA, String classNameB) {
     return packageNameA.equals(packageNameB);
   }
 
+  /** Returns true if {@code childClass} is a subclass of {@code parentClass}. */
+  static boolean isClassSubClassOf(JavaClass childClass, JavaClass parentClass) {
+    for (JavaClass superClass : getClassAndSuperClasses(childClass)) {
```
elharo: Might be simpler as a while or do while loop that operates while parent is not null.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -427,4 +464,23 @@ static String enclosingClassName(String className) {
     }
     return className.substring(0, lastDollarIndex);
   }
+
+  /**
+   * Returns the target class and its superclasses in order (with {@link Object} last). If any can't
+   * be found, the list stops with the previous one.
+   */
+  static Iterable<JavaClass> getClassAndSuperClasses(JavaClass targetClass) {
```
elharo: getClassHierarchy?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -427,4 +464,23 @@ static String enclosingClassName(String className) {
...
+    return SUPERCLASSES.breadthFirst(targetClass);
```
elharo: breadthFirst sounds strange here. We're just walking up a tree, right?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -260,55 +263,146 @@ JarLinkageReport generateLinkageReport(Path jarPath, SymbolReferenceSet symbolRe
...
+      JavaClass targetJavaClass = classDumper.loadJavaClass(targetClassName);
+      Path classFileLocation = classDumper.findClassLocation(targetClassName);
+      if (!isClassAccessibleFrom(targetJavaClass, reference.getSourceClassName())) {
+        return Optional.of(StaticLinkageError.errorInaccessibleClass(reference, classFileLocation));
+      }
+
+      for (JavaClass javaClass : getClassAndSuperClasses(targetJavaClass)) {
+        for (Field field : javaClass.getFields()) {
+          if (field.getName().equals(fieldName)) {
+            // The field is found. Returning no error.
```
elharo: You could still return a different error here though, right?

suztomo: Right. The comment was outdated. Moved to appropriate line.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: It's correct although the method name may sound strange when Guava's Traverser is used for class hierarchy.

Guava's Traverser (`SUPERCLASSES` below) defines graph's (directed) edges. In this case, it defines the relationship from a class to its parent class. Because a Java class has only one parent class, there's only one (outgoing) edge from a node in the graph defined by `SUPERCLASSES`. So `breadthFirst` just works (so does `depthFirst`).

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Nice. Renamed to `getClassHierarchy`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: I just tried the idea, but as `childClass.getSuperClass()` may throw `ClassNotFoundException` (a checked exception), the logic did not become simpler.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Changed to `i++`.

dependencies/pom.xml
suztomo: I agree.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -62,46 +63,52 @@
...
   ImmutableList<Path> getInputClasspath() {
     return inputClasspath;
   }
 
+  private static SyntheticRepository createSyntheticRepository(List<Path> paths) {
+    ClassPath classPath = new LinkageCheckClassPath(paths);
+    return SyntheticRepository.getInstance(classPath);
+  }
+
   static ClassDumper create(List<Path> jarFilePaths) throws IOException {
```
elharo: jarFilePaths --> jarFiles or jarPaths, jarFilePaths is a tad redundant

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -114,11 +121,21 @@ JavaClass loadJavaClass(String className) throws ClassNotFoundException {
...
+    return extensionClassLoader.loadClass(className);
+  }
+
+  boolean isJavaRuntimeClass(String className) {
+    try {
+      if (className.startsWith("[")) {
+        // Array class
+        return true;
+      }
+      loadClassFromExtensionClassLoader(className);
```
elharo: Perhaps use findClass or findSystemClass instead. I'm not sure about the exact semantics here, but we may not need to actually load the class at this point.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -353,23 +369,38 @@ URL findClassLocation(String className) throws ClassNotFoundException {
...
    */
-  private static ImmutableSetMultimap<Path, String> jarFilesToDefinedClasses(
+  @VisibleForTesting
+  static ImmutableSetMultimap<Path, String> jarFilesToDefinedClasses(
       List<Path> jarFilePaths) throws IOException {
     ImmutableSetMultimap.Builder<Path, String> pathToClasses = ImmutableSetMultimap.builder();
-
     for (Path jarFilePath : jarFilePaths) {
-      for (JavaClass javaClass : topLevelJavaClassesInJar(jarFilePath)) {
+      for (JavaClass javaClass : allJavaClassesInJar(jarFilePath)) {
```
elharo: Not sure but could we simply use invertFrom here? https://google.github.io/guava/releases/25.0-jre/api/docs/com/google/common/collect/Multimaps.html#invertFrom-com.google.common.collect.Multimap-M-

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -353,23 +369,38 @@ URL findClassLocation(String className) throws ClassNotFoundException {
...
 
-  private static ImmutableSet<ClassInfo> listTopLevelClassesFromJar(URL jarFileUrl)
+  /**
+   * Returns mapping from class names to the absolute paths of the first jar file that defines the
+   * classes.
+   *
+   * @param jarFilePaths absolute paths to jar files
+   */
+  @VisibleForTesting
+  static ImmutableMap<String, Path> classToDefiningJarFile(List<Path> jarFilePaths)
```
elharo: again would invertFrom work?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -380,18 +411,14 @@ URL findClassLocation(String className) throws ClassNotFoundException {
...
-    ImmutableSet<ClassInfo> allClassesInJar = classPath.getTopLevelClasses();
-    return allClassesInJar;
+    return classPath.getAllClasses();
   }
 
-  private static ImmutableSet<JavaClass> topLevelJavaClassesInJar(Path jarFilePath)
-      throws IOException {
-    String pathToJar = jarFilePath.toString();
-    SyntheticRepository repository = SyntheticRepository.getInstance(new ClassPath(pathToJar));
+  private static ImmutableSet<JavaClass> allJavaClassesInJar(Path jarFilePath) throws IOException {
```
elharo: jarFilePath --> jarPath
allJavaClassesInJar --> getClassesInJar (We can assume there are no non-Java classes there, and method names are usually verbs.)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -380,18 +411,14 @@ URL findClassLocation(String className) throws ClassNotFoundException {
...
+    SyntheticRepository repository = createSyntheticRepository(ImmutableList.of(jarFilePath));
     ImmutableSet.Builder<JavaClass> javaClasses = ImmutableSet.builder();
     URL jarFileUrl = jarFilePath.toUri().toURL();
```
elharo: jarFileUrl --> jarUrl

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
elharo: Makes me wonder if Traverser is overkill here.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckClassPath.java
```diff
@@ -0,0 +1,73 @@
...
+ *     href="https://commons.apache.org/proper/commons-bcel/apidocs/org/apache/bcel/util/ClassPath.html">BCEL's
+ *     ClassPath</a>
+ */
+public class LinkageCheckClassPath extends ClassPath {
+  private final ClassLoader extensionClassLoader;
+
+  /**
+   * Constructs a class path for static linkage check.
+   *
+   * @param paths List of absolute paths for the elements in the class path
```
elharo: nit: list (no caps)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckClassPath.java
```diff
@@ -0,0 +1,73 @@
...
+    extensionClassLoader = ClassLoader.getSystemClassLoader().getParent();
+  }
+
+  /**
+   * Returns input stream identified by {@code name} and {@code suffix}. This uses either the
+   * extension class loader or the class path specified at the constructor argument, ensuring that
+   * the resource is not loaded from the class path of this tool itself.
+   */
+  @Override
+  public InputStream getInputStream(final String name, final String suffix) throws IOException {
```
elharo: Google style does not use final on arguments except where required.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -260,55 +263,147 @@ JarLinkageReport generateLinkageReport(Path jarPath, SymbolReferenceSet symbolRe
...
+      JavaClass targetJavaClass = classDumper.loadJavaClass(targetClassName);
+      Path classFileLocation = classDumper.findClassLocation(targetClassName);
+      if (!isClassAccessibleFrom(targetJavaClass, reference.getSourceClassName())) {
+        return Optional.of(StaticLinkageError.errorInaccessibleClass(reference, classFileLocation));
+      }
+
+      // Checks the target class, its parent classes, and its interfaces.
+      // Interface check is needed to avoid false positive for a method reference to an abstract
+      // class that implements an interface. For example, Guava's ImmutableList is an abstract class
+      // that implements List interface, but the class does not have get() method. A method
```
elharo: the List interface
a get() method

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -260,55 +263,147 @@ JarLinkageReport generateLinkageReport(Path jarPath, SymbolReferenceSet symbolRe
...
+        logger.warning(
+            "The source class "
+                + sourceClassName
+                + " of a reference was not found in the class path when checking accessibility");
+        return false;
+      }
+    }
+    if (member.isPrivate()) {
+      // Access from within same top-level class is allowed to read private class. However, such
+      // case is already filtered at errorsFromSymbolReferences.
```
elharo: cases are

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: I couldn't find a better solution with invertFrom.

- The input for this function is a list of Paths, not Multimap.
- `ImmutableSetMultimap.flatteningToImmutableSetMultimap` seemed help but as `listAllJavaClassesInJar` throws IOException (checked exception), the lambda would look complex.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Renamed to `findSystemClass`.

I think using the extension class loader is easy way to distinguish whether a class is from JVM or not.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: I couldn't find a method that can handle the ordering (`putIfAbsent` below implements the order), or picking up the first path.

- `invertFrom` is inverting keys and values for multimap, but the input here is a list of Path.
- `jarFileToClasses.inverse()` would give the inverted multimap but the value is a set.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -350,55 +370,65 @@ URL findClassLocation(String className) throws ClassNotFoundException {
...
+    for (Path jarFilePath : jarFilePaths) {
+      for (JavaClass javaClass : listAllJavaClassesInJar(jarFilePath)) {
+        // The first entry wins in the same manner as JVM's class loading.
+        classToJar.putIfAbsent(javaClass.getClassName(), jarFilePath);
+      }
+    }
+    return ImmutableMap.copyOf(classToJar);
+  }
 
+  private static ImmutableSet<ClassInfo> listAllClassInfoFromJar(URL jarUrl)
```
elharo: listClassInfoFromJar or maybe just listClassInfo

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -350,55 +370,65 @@ URL findClassLocation(String className) throws ClassNotFoundException {
...
-    ImmutableSet<ClassInfo> allClassesInJar = classPath.getTopLevelClasses();
-    return allClassesInJar;
+    return classPath.getAllClasses();
   }
 
-  private static ImmutableSet<JavaClass> topLevelJavaClassesInJar(Path jarFilePath)
-      throws IOException {
-    String pathToJar = jarFilePath.toString();
-    SyntheticRepository repository = SyntheticRepository.getInstance(new ClassPath(pathToJar));
+  private static ImmutableSet<JavaClass> listAllJavaClassesInJar(Path jarPath) throws IOException {
```
elharo: listClassesInJar unless there are some non-Java classes we're omitting

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -350,55 +370,65 @@ URL findClassLocation(String className) throws ClassNotFoundException {
   }
 
   /**
-   * @param jarFilePaths absolute paths to jar files
-   * @return map of jar file paths to classes defined in them
+   * Returns mapping from jar files to the names of their defining classes.
```
elharo: the classes they define?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -350,55 +370,65 @@ URL findClassLocation(String className) throws ClassNotFoundException {
...
+   *
+   * @param jarPaths absolute paths to jar files
    */
-  private static ImmutableSetMultimap<Path, String> jarFilesToDefinedClasses(
-      List<Path> jarFilePaths) throws IOException {
+  @VisibleForTesting
+  static ImmutableSetMultimap<Path, String> scanJarFilesToDefinedClasses(
```
elharo: I don't follow this method name. Maybe mapJarToClasses? 

In general try to avoid duplicating the method arguments (here JarFiles/jarPaths) in method names unless it's necessary for clarity or correctness.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -350,55 +370,65 @@ URL findClassLocation(String className) throws ClassNotFoundException {
...
 
-  private static ImmutableSet<ClassInfo> listTopLevelClassesFromJar(URL jarFileUrl)
+  /**
+   * Returns mapping from class names to the absolute paths of the first jar file that defines the
+   * classes.
+   *
+   * @param jarFilePaths absolute paths to jar files
+   */
+  @VisibleForTesting
+  static ImmutableMap<String, Path> scanClassToDefiningJarFile(List<Path> jarFilePaths)
```
elharo: perhaps scan --> map; e.g. mapClassToJar

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckClassPath.java
```diff
@@ -0,0 +1,74 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.bcel.util.ClassPath;
+
+/**
+ * Class path to load class files for {@link StaticLinkageChecker}. When loading a resource for a
```
elharo: Classpath is more commonly written as a single word.

suztomo:
@elharo Applied code review and responded to comments. PTAL.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -62,46 +63,53 @@
   private final ImmutableList<Path> inputClasspath;
   private final SyntheticRepository syntheticRepository;
   private final ClassLoader classLoader;
```
elharo: Do we need both classloaders? If so, why?

suztomo: Good catch. `classLoader` is no longer needed and now removed.
(It used to play a role to construct argument list to lookup methods from Java class. The logic is no longer there)

suztomo:
@elharo Removed unnecessary `classLoader` variable. PTAL.

Issue #309, opened 2019-01-02
StaticLinkageChecker: add Gradle's dependency mediation


suztomo:
As of now StaticLinkageChecker performs dependency mediation in the same manner as Maven does. Add an functionality to check Gradle's way, so that the result of StaticLinkageChecker makes sense to Gradle users too:

> Gradle resolves version conflicts by picking the highest version of a module. 

https://docs.gradle.org/current/userguide/troubleshooting_dependency_resolution.html#sub:version_conflicts

PR #310, opened 2019-01-02
Filter duplicate artifacts of different versions by groupId:artifactId

### LINKED ISSUE #306: StaticLinkageChecker: false positive on anonymous class
StaticLinkageChecker raises an error for junit's use of hamcrest's TypeSafeMatcher

```

junit-4.12.jar (11 errors):
  Linked from: io.grpc:grpc-testing:1.17.1 / junit:junit:4.12
  ...
  MethodSymbolReference{sourceClassName=org.junit.experimental.results.ResultMatchers$1, targetClassName=org.hamcrest.TypeSafeMatcher, methodName=<init>, descriptor=()V}, reason: INACCESSIBLE, target class from /usr/local/google/home/suztomo/.m2/repository/org/hamcrest/hamcrest-core/1.3/hamcrest-core-1.3.jar

```

```
    public static Matcher<PrintableResult> failureCountIs(final int count) {
        return new TypeSafeMatcher<PrintableResult>() {
            public void describeTo(Description description) {
                description.appendText("has " + count + " failures");
            }

            @Override
            public boolean matchesSafely(PrintableResult item) {
                return item.failureCount() == count;
            }
        };
    }
```

### LINKED ISSUE #307: StaticLinkageChecker to tell an artifact is shadowed or not
Maven dependency resolution would not pick up other versions of same artifact. The first version closest to the root is said to _shadow_ other versions.

As of now static linkage report does not tell a jar file is shadowed or not. For example, there are multiple appearance of grpc-netty-shaded and only the first one is concerning.

```
grpc-netty-shaded-1.17.1.jar (2 errors):
...

grpc-netty-shaded-1.16.1.jar (3 errors):
...

```

Further more, if there's a no issue in an artifact, it does not show up in the dashboard, making it difficult to tell whether an entry is concerning or not.

# Solution

## Option 1: Add note on shadowed or not

Static linkage report to tell whether a jar file is shadowed by another version or not.

```
grpc-netty-shaded-1.17.1.jar (2 errors):
...

grpc-netty-shaded-1.16.1.jar (3 errors, shadowed by version 1.17.1):
...

```

## Option 2: Do not run linkage check for shadowed jar files

For each groupId+artifactId combination, static linkage check runs against only one version. `StaticLinkageChecker.artifactsToClasspath` to not to return duplicated versions for one artifact.

`StaticLinkageChecker.artifactsToClasspath` used to filter duplicated jar files for "groupId:artifactId" as key. But the logic is no longer there. To confirm with Elliotte on this.

https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/279/files#diff-cb2eefbf6e8ca674c445c1c7e1b519edL129

suztomo:
Fixes #307 (on the option 2 in the issue)
Fixes #306, by ensuring there should not be duplicated jar files containing `TypeSafeMatcher`

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
@@ -148,12 +148,12 @@ public class StaticLinkageChecker {
   /**
    * Finds jar file paths and dependency paths for Maven artifacts and their transitive
    * dependencies. When there are multiple versions of an artifact, the closest to the root ({@code
-   * artifacts}) is picked up into the key of the return value. This 'pick closest' strategy follows
-   * Maven's dependency mediation.
+   * artifacts}) is picked up. This 'pick closest' strategy follows Maven's dependency mediation.
+   * The values of the returned map for a key (jar file) represent the different Maven dependency
+   * paths from {@code artifacts} to the Maven artifact of the jar file.
    *
    * @param artifacts Maven artifacts to check
-   * @return map absolute paths of jar files to one or more Maven dependency path(s) from ({@code
-   *     artifacts}) to the Maven artifacts of the jar files
+   * @return map absolute paths of jar files to one or more Maven dependency paths
    * @throws RepositoryException when there is a problem in retrieving jar files
    * @see <a
    *     href="https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Transitive_Dependencies">Maven:
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
@@ -32,6 +32,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.ListMultimap;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import java.io.IOException;
 import java.nio.file.Path;
@@ -170,19 +171,27 @@ public class StaticLinkageChecker {
     List<DependencyPath> dependencyPaths = dependencyGraph.list();
 
     // To remove duplicates on (groupId:artifactId) for dependency mediation
-    Set<String> artifactsInPaths = Sets.newHashSet();
+    Map<String, Artifact> keyToFirstArtifact = Maps.newHashMap();
 
     for (DependencyPath dependencyPath : dependencyPaths) {
       Artifact artifact = dependencyPath.getLeaf();
+      String artifactVersion = artifact.getVersion();
       // groupId:artifactId
       String dependencyMediationKey = Artifacts.makeKey(artifact);
-      if (!artifactsInPaths.add(dependencyMediationKey)) {
-        // Not adding this artifact if (<groupId>:<artifactId>) is already in `multimap`.
+      Artifact firstArtifactForKey = keyToFirstArtifact.get(dependencyMediationKey);
+      if (firstArtifactForKey != null
+          && !artifactVersion.equals(firstArtifactForKey.getVersion())) {
+        // Not adding this artifact if different version of the artifact (<groupId>:<artifactId> as
+        // key) is already in `multimap`.
         // As `dependencyPaths` elements are in level order (breadth-first), this first-wins
         // strategy follows Maven's dependency mediation.
         // TODO(#309): add Gradle's dependency mediation
         continue;
       }
+      keyToFirstArtifact.put(dependencyMediationKey, artifact);
+      // When finding key first time, or additional dependency path to the artifact with same
+      // version is encountered, add the dependency path to `multimap`
+
       Path jarAbsolutePath = artifact.getFile().toPath().toAbsolutePath();
       if (!jarAbsolutePath.toString().endsWith(".jar")) {
         continue;
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
@@ -144,7 +144,6 @@ public class StaticLinkageChecker {
     return ImmutableList.copyOf(multimap.keySet());
   }
 
-  // Multimap is a pain, maybe just use LinkedHashMap<Path, List<DependencyPath>>
   /**
    * Finds jar file paths and dependency paths for Maven artifacts and their transitive
    * dependencies. When there are multiple versions of an artifact, the closest to the root ({@code
@@ -161,6 +160,8 @@ public class StaticLinkageChecker {
    */
   public static LinkedListMultimap<Path, DependencyPath> artifactsToPaths(List<Artifact> artifacts)
       throws RepositoryException {
+    // TODO(#315): Consider using LinkedHashMap<Path, List<DependencyPath>> over Multimap
+    // Multimap has many variation with different behaviors.
 
     LinkedListMultimap<Path, DependencyPath> multimap = LinkedListMultimap.create();
     if (artifacts.isEmpty()) {
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
@@ -189,8 +189,8 @@ public class StaticLinkageChecker {
         continue;
       }
       keyToFirstArtifactVersion.put(dependencyMediationKey, artifact.getVersion());
-      // When finding key first time, or additional dependency path to the artifact with same
-      // version is encountered, adds the dependency path to `multimap`,
+      // When finding the key (groupId:artifactId) first time, or additional dependency path to
+      // the artifact of the same version is encountered, adds the dependency path to `multimap`.
 
       Path jarAbsolutePath = artifact.getFile().toPath().toAbsolutePath();
       if (!jarAbsolutePath.toString().endsWith(".jar")) {
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
@@ -33,7 +33,6 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


suztomo:
CircleCI is having an outage:

> Partial System Outage - components affected: CircleCI 2.0, CircleCI, macOS Builds, macOS 2.0 Builds, Workflows, CircleCI UI, Scheduled Jobs
Updated 9 min ago - CircleCI UI unavailable: We are investigating a major outage on our platform that affects multiple components. Our engineers are working to determine the cause and will provide an update within 20 minutes.

suztomo:
Circle CI shows green now.

elharo:
This is tricky. We need to think about whether this approach makes sense in context. Shadowing can change depending on artifact order.

suztomo:
The artifact order from `DependencyGraphBuilder.getStaticLinkageCheckDependencies();` is level order (breadth-first traversal).
I will talk to you later.

suztomo:
For checking static ordering of BOM members, this PR will take care of it.

For discussion on non-static ordering , #311 is created.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -163,8 +169,20 @@ public static void main(String[] arguments)
...
     List<DependencyPath> dependencyPaths = dependencyGraph.list();
 
+    // To remove duplicates on (groupId:artifactId) for dependency mediation
+    Set<String> artifactsInPaths = Sets.newHashSet();
+
     for (DependencyPath dependencyPath : dependencyPaths) {
       Artifact artifact = dependencyPath.getLeaf();
+      // groupId:artifactId
+      String dependencyMediationKey = Artifacts.makeKey(artifact);
+      if (!artifactsInPaths.add(dependencyMediationKey)) {
```
elharo: Do we still need a multimap here if we're only adding each groupId:aritfactId the first time?

suztomo:
@elharo Fixed the logic on handling multimap. PTAL.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -129,31 +132,35 @@ public static void main(String[] arguments)
...
   }
-  
-  
+
   // Multimap is a pain, maybe just use LinkedHashMap<Path, List<DependencyPath>>
   /**
-   * Finds jar file paths for Maven artifacts and their dependencies.
+   * Finds jar file paths and dependency paths for Maven artifacts and their transitive
+   * dependencies. When there are multiple versions of an artifact, the closest to the root
+   * (`artifacts` argument) is picked up in the same manner as Maven's dependency mediation.
```
elharo: This isn't quite a full description of what this method does. In particular, it doesn't say that we include multiple paths to the same version of the artifact, which is sort of a weird thing to do so it's worth a comment.

And on that subject, why do we need multiple dependency paths for the artifact? Why isn't the first one enough? Especially since if the first one is removed from the graph, a different version of the artifact may be picked?

netdpb:
```suggestion
    * ({@code artifacts}) is picked up in the same manner as Maven's dependency mediation.
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -163,8 +170,28 @@ public static void main(String[] arguments)
...
+      if (firstArtifactVersionForKey != null
+          && !artifactVersion.equals(firstArtifactVersionForKey)) {
+        // Not adding this artifact if different version of the artifact (<groupId>:<artifactId> as
+        // key) is already in `multimap`.
+        // As `dependencyPaths` elements are in level order (breadth-first), this first-wins
+        // strategy follows Maven's dependency mediation.
+        // TODO(#309): add Gradle's dependency mediation
+        continue;
+      }
+      keyToFirstArtifactVersion.put(dependencyMediationKey, artifact.getVersion());
```
netdpb: Should all this happen after seeing whether `jarAbsolutePath` ends with ".jar", below?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -163,8 +170,28 @@ public static void main(String[] arguments)
         DependencyGraphBuilder.getStaticLinkageCheckDependencies(artifacts);
     List<DependencyPath> dependencyPaths = dependencyGraph.list();
 
+    // To remove duplicates on (groupId:artifactId) for dependency mediation
+    Map<String, String> keyToFirstArtifactVersion = Maps.newHashMap();
```
netdpb: This all seems like it would be better to be done while creating the dependency graph, not afterwards. Could `getStaticLinkageCheckDependencies()` do this, and just include the path to the correct dependency?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Fixed.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: I prefer this function (the caller of getStaticLinkageCheckDependencies) to pickup correct artifacts.

- `getStaticLinkageCheckDependencies` is responsible to build a graph of dependency.
- The caller of the function is responsible to pick up artifacts (with version) depending on a dependency mediation strategy.
- As of now the strategy is Maven's dependency mediation (pick closest)
- In future we'll add an option to check Gradle's dependency mediation (pick highest version) #309.


Renamed `getStaticLinkageCheckDependencies` to `getStaticLinkageCheckDependencyGraph` to clarify that it returns a graph.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -129,46 +131,73 @@ public static void main(String[] arguments)
...
   }
-  
-  
+
   // Multimap is a pain, maybe just use LinkedHashMap<Path, List<DependencyPath>>
   /**
-   * Finds jar file paths for Maven artifacts and their dependencies.
+   * Finds jar file paths and dependency paths for Maven artifacts and their transitive
+   * dependencies. When there are multiple versions of an artifact, the closest to the root ({@code
+   * artifacts}) is picked up into the key of the return value. This 'pick closest' strategy follows
```
elharo: drop "into the key of the return value". The issue here are the values, not the keys. Can you elaborate why the values are a list instead of a single path? That is, why do we need a multimap?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
elharo: side note, but You probably don't need "StaticLinkageCheck" in either of these method names.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: Dropped unnecessary phrase, and added "The values of the returned map for a key (jar file) are the different Maven dependency paths from ({@code artifacts}) to the Maven artifact of the jar file."

Multimap gives more information. For example:

```
grpc-core-1.17.1.jar (4 errors):
  Linked from: io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-alts:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-auth:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-grpclb:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-netty:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-netty-shaded:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-okhttp:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-protobuf:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-protobuf-lite:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-protobuf-nano:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-stub:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-testing:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-alts:1.17.1 / io.grpc:grpc-auth:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-alts:1.17.1 / io.grpc:grpc-netty-shaded:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-alts:1.17.1 / io.grpc:grpc-protobuf:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-alts:1.17.1 / io.grpc:grpc-stub:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-alts:1.17.1 / io.grpc:grpc-grpclb:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-grpclb:1.17.1 / io.grpc:grpc-protobuf:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-grpclb:1.17.1 / io.grpc:grpc-stub:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-protobuf:1.17.1 / io.grpc:grpc-protobuf-lite:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-testing:1.17.1 / io.grpc:grpc-stub:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-alts:1.17.1 / io.grpc:grpc-protobuf:1.17.1 / io.grpc:grpc-protobuf-lite:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-alts:1.17.1 / io.grpc:grpc-grpclb:1.17.1 / io.grpc:grpc-protobuf:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-alts:1.17.1 / io.grpc:grpc-grpclb:1.17.1 / io.grpc:grpc-stub:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-grpclb:1.17.1 / io.grpc:grpc-protobuf:1.17.1 / io.grpc:grpc-protobuf-lite:1.17.1 / io.grpc:grpc-core:1.17.1
  Linked from: io.grpc:grpc-alts:1.17.1 / io.grpc:grpc-grpclb:1.17.1 / io.grpc:grpc-protobuf:1.17.1 / io.grpc:grpc-protobuf-lite:1.17.1 / io.grpc:grpc-core:1.17.1
  MethodSymbolReference{sourceClassName=io.grpc.internal.ServiceConfigInterceptor$1DelayedHedgingPolicyProvider, targetClassName=com.google.common.base.Verify, methodName=verify, descriptor=(ZLjava/lang/String;Ljava/lang/Object;)V}, reason: SYMBOL_NOT_FOUND, target class from /usr/local/google/home/suztomo/.m2/repository/com/google/guava/guava/20.0/guava-20.0.jar
```
is more informative than

```
grpc-core-1.17.1.jar (4 errors):
  Linked from: io.grpc:grpc-core:1.17.1
MethodSymbolReference{sourceClassName=io.grpc.internal.ServiceConfigInterceptor$1DelayedHedgingPolicyProvider, targetClassName=com.google.common.base.Verify, methodName=verify, descriptor=(ZLjava/lang/String;Ljava/lang/Object;)V}, reason: SYMBOL_NOT_FOUND, target class from /usr/local/google/home/suztomo/.m2/repository/com/google/guava/guava/20.0/guava-20.0.jar
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: DependencyGraphBuilder serves other use cases than StaticLinkageCheck. So I'd keep the function names.

suztomo:
Apart from dependency remediation strategy, a class path can contain different jar files that contains same class. I will work on clarifying the JVM's behavior on what happens in that case.

PR #322, opened 2019-01-07
StaticLinakgeChecker to validate InterfaceMethodref

### LINKED ISSUE #320: StaticLinkageChecker: validate InterfaceMethodref
It already checks CONSTANT_Fieldref_info, CONSTANT_Methodref_info, and CONSTANT_Class_info. 

Let's add support for CONSTANT_InterfaceMethodref_info. Very likely it's treatment is same as methodref from our perspective.

[4.4.2. The CONSTANT_Fieldref_info, CONSTANT_Methodref_info, and CONSTANT_InterfaceMethodref_info Structures](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.4.2)

[5.4.3.4. Interface Method Resolution](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-5.html#jvms-5.4.3.4)

suztomo:
StaticLinakgeChecker already validates method references to classes. This change is to validate method references to interfaces.

Fixes #320

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -182,16 +182,12 @@ class ClassDumper {
           }
           break;
         case Const.CONSTANT_Methodref:
-          ConstantMethodref constantMethodref = (ConstantMethodref) constant;
-          methodReferences.add(
-              constantToMethodReference(constantMethodref, constantPool, sourceClassName, false));
-          break;
+          // fall through
         case Const.CONSTANT_InterfaceMethodref:
-          ConstantInterfaceMethodref constantInterfaceMethodref =
-              (ConstantInterfaceMethodref) constant;
+          // Both ConstantMethodref and ConstantInterfaceMethodref are subclass of ConstantCP
+          ConstantCP constantMethodref = (ConstantCP) constant;
           methodReferences.add(
-              constantToMethodReference(
-                  constantInterfaceMethodref, constantPool, sourceClassName, true));
+              constantToMethodReference(constantMethodref, constantPool, sourceClassName));
           break;
         case Const.CONSTANT_Fieldref:
           ConstantFieldref constantFieldref = (ConstantFieldref) constant;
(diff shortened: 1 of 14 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
```diff
@@ -34,6 +34,9 @@
 
   public abstract ImmutableList<StaticLinkageError<MethodSymbolReference>> getMissingMethodErrors();
 
+  public abstract ImmutableList<StaticLinkageError<MethodSymbolReference>>
+      getMissingInterfaceMethodErrors();
```
netdpb: If `MethodSymbolReference` itself encoded whether it was an interface method, would you still need this separate property?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -183,14 +189,30 @@ JarLinkageReport generateLinkageReport(Path jarPath, SymbolReferenceSet symbolRe
...
   }
 
+  @VisibleForTesting
+  Optional<StaticLinkageError<MethodSymbolReference>> checkLinkageErrorMissingMethodAt(
+      MethodSymbolReference reference) {
+    return checkLinkageErrorMethodAt(reference, false);
+  }
+
+  @VisibleForTesting
+  Optional<StaticLinkageError<MethodSymbolReference>> checkLinkageErrorMissingInterfaceMethodAt(
```
netdpb: As above, if `MethodSymbolReference` encoded whether it was an interface method reference or not, would you need this duplication?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
suztomo: No, it's not needed. Let me implement in that way...

suztomo:
@netdpb Now `isInterfaceMethod` is a property of `MethodSymbolReference`. PTAL.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -257,6 +259,7 @@ private static MethodSymbolReference constantToMethodReference(
         MethodSymbolReference.builder()
             .setSourceClassName(sourceClassName)
             .setMethodName(methodName)
+            .setInterfaceMethod(isInterfaceMethod)
```
netdpb: Could you do an instanceof check on `constantMethodRef` here instead of passing a boolean? Is that better or worse?

Note that if you did this, you could fold the `CONSTANT_Methodref` and `CONSTANT_InterfaceMethodref` cases above.

suztomo: Good idea. Committed change, with '// fall through' in the switch case.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -181,7 +182,10 @@ private static SymbolReferenceSet scanSymbolReferencesInClass(JavaClass javaClas
           }
           break;
         case Const.CONSTANT_Methodref:
-          ConstantMethodref constantMethodref = (ConstantMethodref) constant;
+          // fall through
```
netdpb: This doesn't need `// fall-through`. It's not a fall through from one case to another; it's a multi-label case. Error Prone makes you put the fall-through comment only after at least one statement in a case.

PR #331, opened 2019-01-09
Glossary to clarify class path

### LINKED ISSUE #314: Glossary needs to handle case of multiply defined classes
[glossary.md](https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/library-best-practices/glossary.md) says:

an error when a Java class in a class path references another class (as a class literal, a field access, or a method invocation), and the reference cannot be satisfied with the available classes in the class path.

We seem to be assuming that a class cannot appear in a classpath more than once. This is not true. Indeed not only classes but the same or different versions of an entire jar can and do appear multiple times.

We need to expand the definitions in this document to consider this case. We should be able to answer these questions:

1. If a reference is satisfied by the first appearance of a class in the classpath but not the second, is that a static linkage error or not?

2. If a reference is satisfied by the seconf appearance of a class in the classpath but not the first, is that a static linkage error or not?

suztomo:
Fixes #314 

This change clarifies a class file is _unavailable_ in a class path if it's shadowed by another class file in another path entry in the class path.

# Questions from the issue

> If a reference is satisfied by the first appearance of a class in the classpath but not the second, is that a static linkage error or not?

It's not a static linkage error, because the second class is not in _available classes in the class path_.

> If a reference is satisfied by the second appearance of a class in the classpath but not the first, is that a static linkage error or not?

Yes, it's a static linkage error. The first one, which is not satisfying a reference, is  in _available classes in the class path_.

@elharo What do you think?

### DIFF
--- a/library-best-practices/glossary.md
+++ b/library-best-practices/glossary.md
@@ -3,7 +3,7 @@ Java Dependency Glossary
 
 - **Class path**: an ordered list of jar files, zip files, or directories, each of which
   contains Java class files.
-  A class loader searches a class file by its name through path entries in a class path.
+  A class loader searches a class file by its name through the path entries in a class path.
   When there are two or more path entries (for example, jar files) that contains class files with
   the same name in a class path, the class file in the first path entry in the class path
   is available for use, and the other class files in the rest of the path entries are unavailable
--- a/library-best-practices/glossary.md
+++ b/library-best-practices/glossary.md
@@ -3,11 +3,10 @@ Java Dependency Glossary
 
 - **Class path**: an ordered list of jar files, zip files, or directories, each of which
   contains Java class files.
-  A class loader searches a class file by its name through the path entries in a class path.
-  When there are two or more path entries (for example, jar files) that contains class files with
+  A class loader searches for a class file by its name through the path entries in a class path.
+  When there are two or more path entries (for example, jar files) that contain class files with
   the same name in a class path, the class file in the first path entry in the class path
-  is available for use, and the other class files in the rest of the path entries are unavailable
-  through the class path.
+  is available for use, and the other class files with the same name are unavailable.
 
 ### Types of conflicts and compatibility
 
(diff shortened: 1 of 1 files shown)


library-best-practices/glossary.md
```diff
@@ -1,6 +1,14 @@
 Java Dependency Glossary
 ------------------------
 
+- **Class path**: an ordered list of jar files, zip files, or directories, each of which
+  contains Java class files.
+  A class loader searches a class file by its name through the path entries in a class path.
+  When there are two or more path entries (for example, jar files) that contains class files with
```
garrettjonesgoogle: s/contains/contain/

library-best-practices/glossary.md
```diff
@@ -1,6 +1,14 @@
...
+  the same name in a class path, the class file in the first path entry in the class path
+  is available for use, and the other class files in the rest of the path entries are unavailable
```
garrettjonesgoogle: "and the other class files with the same name"

library-best-practices/glossary.md
```diff
@@ -1,6 +1,14 @@
...
+  through the class path.
```
garrettjonesgoogle: I would say remove "through the class path" because that is implied already.

library-best-practices/glossary.md
```diff
@@ -1,6 +1,14 @@
 Java Dependency Glossary
 ------------------------
 
+- **Class path**: an ordered list of jar files, zip files, or directories, each of which
+  contains Java class files.
+  A class loader searches a class file by its name through the path entries in a class path.
```
elharo: "searches for a class file"

suztomo: Fixed.

library-best-practices/glossary.md
suztomo: Fixed.

library-best-practices/glossary.md
suztomo: Nice. Updated.

library-best-practices/glossary.md
suztomo: Removed.

PR #333, opened 2019-01-09
Example projects to demonstrate runtime errors

### LINKED ISSUE #327: Document gRPC vs google-api-client
Guava inconsistency if google-api-client 1.27.0 precedes grpc 1.17.0 (both latest releases as I type this)

io.grpc:grpc-core:0.17.1 and com.google.api-client:google-api-client:1.27.0  re Guava Verify

see Tomo's example

suztomo:
Fixes #327 .

@elharo Do they (directory location and explanation on the error) look good?


dependencies/example-projects/grpc-vs-google-api-client/README.md
```diff
@@ -0,0 +1,62 @@
+# grpc 1.17.1 v.s. google-api-client 1.27.0
```
netdpb:
```suggestion
# grpc 1.17.1 vs google-api-client 1.27.0
```

Let's be consistent. "vs" is okay, as is "vs." or "versus".

dependencies/example-projects/grpc-vs-google-api-client/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+This project is to demonstrate `NoSuchMethodError` generated from the class path
```
netdpb:
```suggestion
This project demonstrates a `NoSuchMethodError` generated from the class path
```

dependencies/example-projects/grpc-vs-google-api-client/README.md
netdpb: I wonder whether we should name these examples for the dependencies or for the case we're demonstrating.

Should this directory be called "missing-method" or something like that? It could still use these two dependencies, but that way if we decide to change the actual dependencies used to something else we wouldn't have a stale directory name.

dependencies/example-projects/grpc-vs-google-api-client/README.md
```diff
@@ -0,0 +1,62 @@
...
+    at java.lang.Thread.run (Thread.java:748)
+[INFO] ------------------------------------------------------------------------
+[INFO] BUILD FAILURE
+[INFO] ------------------------------------------------------------------------
+[INFO] Total time: 0.489 s
+[INFO] Finished at: 2019-01-09T16:58:10-05:00
+[INFO] Final Memory: 22M/970M
+[INFO] ------------------------------------------------------------------------
+```
+
```
netdpb: Delete this line.

dependencies/example-projects/grpc-vs-google-api-client/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Diagnosis
+
+## google-api-client 1.27.0
+
+This artifact has dependency to `com.google.guava:guava:20.0`.
```
netdpb:
```suggestion
This artifact depends on `com.google.guava:guava:20.0`.
```

dependencies/example-projects/grpc-vs-google-api-client/README.md
```diff
@@ -0,0 +1,62 @@
...
+This version of Guava does not have `Verify.verify` method with the signature below
+([Javadoc](https://google.github.io/guava/releases/20.0/api/docs/com/google/common/base/Verify.html)).
```
netdpb: Let's use reference links for these.

In fact you could make `Verify.verify(boolean, String, Object)` on the line above into the link.

dependencies/example-projects/grpc-vs-google-api-client/README.md
```diff
@@ -0,0 +1,62 @@
...
+[INFO] ------------------------------------------------------------------------
+```
+
+
+# Diagnosis
+
+## google-api-client 1.27.0
+
+This artifact has dependency to `com.google.guava:guava:20.0`.
+This version of Guava does not have `Verify.verify` method with the signature below
```
netdpb:
```suggestion
This version of Guava does not have a `Verify.verify(boolean, String, Object)` method
```

dependencies/example-projects/grpc-vs-google-api-client/README.md
```diff
@@ -0,0 +1,62 @@
...
+([Javadoc](https://google.github.io/guava/releases/20.0/api/docs/com/google/common/base/Verify.html)).
+
+When google-api-client appears first in pom.xml, Maven picks up Guava version 20.0
```
netdpb:
```suggestion
When google-api-client appears first in a breadth-first traversal of the dependencies, Maven puts Guava version 20.0
```

dependencies/example-projects/grpc-vs-google-api-client/README.md
```diff
@@ -0,0 +1,62 @@
...
+into the class path.
+
+## grpc-core 1.17.1 
+
+This artifact has dependency to `com.google.guava:guava:26.0-android`.
```
netdpb:
```suggestion
This artifact depends on `com.google.guava:guava:26.0-android`.
```

dependencies/example-projects/grpc-vs-google-api-client/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+`io.grpc.internal.DnsNameResolver.maybeChooseServiceConfig` uses Guava's
```
netdpb:
```suggestion
`DnsNameResolver.maybeChooseServiceConfig` calls `Verify.verify(boolean, String, Object)`
```

dependencies/example-projects/grpc-vs-google-api-client/README.md
```diff
@@ -0,0 +1,62 @@
...
+`Verify.verify` method with following method signature:
+`void verify(boolean expression, String errorMessageTemplate, Object p1)`
+([Javadoc](https://google.github.io/guava/releases/26.0-android/api/docs/com/google/common/base/Verify.html#verify-boolean-java.lang.String-java.lang.Object-)).
+This method with the signature has been added since Guava version 23.1.
```
netdpb:
```suggestion
This method was added to Guava in version 23.1.
```

suztomo: Fixed.

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: Updated.

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: Updated.

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: Updated.

dependencies/example-projects/grpc-vs-google-api-client/README.md
elharo: I think we should keep the artifacts involved. We may have more than one example.

dependencies/example-projects/grpc-vs-google-api-client/README.md
```diff
@@ -0,0 +1,62 @@
+# grpc 1.17.1 v.s. google-api-client 1.27.0
+
+This project is to demonstrate `NoSuchMethodError` generated from the class path
+for google-api-client and grpc-core.
+
+The package name `io.grpc.internal` is used for convenience, because DnsNameResolver
+is package private.
```
elharo: Any way to make this happen without doing that by calling something higher up in the tree?

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: Updated. Does this look better?

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: Updated.

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: Deleted.

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: Updated to "vs".

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: Updated, and changed active form.

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: I used the artifact-specific name so that we document the conflict between the two artifacts with the versions.

This reflects the title of the issue #327 "Document gRPC vs google-api-client".

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: Searching for the caller...

dependencies/example-projects/grpc-vs-google-api-client/README.md
netdpb: I'm unclear on the point of these examples. Are they meant to be examples of the ways in which static linkage errors could cause runtime errors in user projects? If so, then the specific artifacts involved don't really matter.

If they're meant to be test cases for specific conflicts known to occur, then why aren't they test cases? How does writing them up in markdown help anyone? And if they're meant to be known errors, then they'll get stale as the specific versions shown get older.

Who is the audience for these examples? What do we expect them to learn from them?

dependencies/example-projects/grpc-vs-google-api-client/README.md
netdpb: It doesn't look pushed?

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: I'm stuck at finding an example gRPC service that uses DNS TXT record for its configuration.
https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/DnsNameResolver.java#L503

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: When filing a ticket to another repository (e.g., google-api-client) to fix a diamond dependency, I would add a link from the ticket to this example directory to explain the problem.

A JUnit test case for `StaticLinkageChecker` is not convincing for that purpose.

Audience would be the readers of such tickets.

dependencies/example-projects/grpc-vs-google-api-client/README.md
suztomo: As talked, I've added the explanation on why avoiding `io.grpc.internal` package is difficult.

dependencies/example-projects/grpc-vs-google-api-client/README.md
elharo: I think this is critical to prove we're addressing a real problem. These problems aren't hypothetical. They happen, and this is the simplest, most obvious we've found yet. This isn't a test case though. Its documentation and motivation. 

I don't think we should be using personal Github repos for official Github work.

PR #341, opened 2019-01-15
Upgrade to Guava 26.0-android and google-cloud-java 0.77

elharo:
@garrettjonesgoogle @suztomo 

Mostly this PR upgrades the BOM to the latest available versions which bandages the diamond dependency between the latest gRPC and Guava 20. However, there was one unexpected hitch. 
guava-gwt does not have a 26.0-android (or any android) version. For the moment I've taken it out, but perhaps we should do something else here.


garrettjonesgoogle:
@netdpb why doesn't guava-gwt have a 26.0-android version? We aren't using it for android compatibility, we are using it as a proxy for Java 7 support.

netdpb:
Checking. @cpovirk might have some more info.

boms/cloud-oss-bom/pom.xml
```diff
@@ -50,11 +50,6 @@
         <artifactId>guava</artifactId>
         <version>${guava.version}</version>
       </dependency>
-      <dependency>
```
garrettjonesgoogle: Maybe comment this out instead of deleting it because it's a short-term hack to remove it?

boms/cloud-oss-bom/pom.xml
```diff
@@ -29,8 +29,8 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <guava.version>20.0</guava.version>
-    <google.cloud.java.version>0.74.0-alpha</google.cloud.java.version>
+    <guava.version>26.0-android</guava.version>
```
suztomo: Would you add comment on why "-android" is used?

```
<!-- "-android" is to support Java 7 -->
```

boms/cloud-oss-bom/pom.xml
elharo: done

cpovirk:
I think our hope was that no Guava users needed the combination of Java 7 support and GWT support. Do you, or are you just adding it for completeness? (It looks like @netdpb left it out of [the Android/Java7 Guava BOM](https://github.com/google/guava/blob/838560034dfaa1afdf51a126afe6b8b8e6cce3dd/android/guava-bom/pom.xml).

If it's needed, I'm not aware of any technical obstacles to including it. It's more a question of priorities, especially if we plan to drop Java 7 support in the medium term.

netdpb:
I think the issue for libraries that depend on Guava is that they want to support Java7/Android and GWT users, even if there aren't necessarily users who use Java7/Android _and_ GWT at the same time.

cpovirk:
Ah, hmm. We probably need to unpack the phrase "supporting GWT."

- Most of the GWT stuff in Guava is for the GWT client only (like, to be compiled to JavaScript for the web browser). If a library wants to depend on Guava to use on the server side (including Java 7), apps could still use that library in a GWT server. And if those apps happened to want to use Guava on the client, too, they could add in `guava-gwt`, even though the library doesn't use it.

- We do have _some_ server-side classes (for GWT-RPC). However, I would expect the situation to be the same there: If a library isn't doing GWT-RPC, apps can still add in `guava-gwt` and make it work.

(One possible annoyance is that it might be wise to depend on both `guava` and `guava-gwt` (to increase the chances that Maven gives you, say, `26.0-jre` of both instead of `26.0-jre` for one and `25.0-android` for the other). But this is the same problem we potentially have anywhere people use `guava-gwt`, thanks to Maven's version selection.)

netdpb:
I would be okay with any solution that adds some compatible version of guava-gwt into the -android guava-bom. I think it looks a little weird for guava-bom:$version-android to include guava-gwt:$version-jre, but few people are going to be looking that hard.

cpovirk:
People might conceivably read that as a statement that `guava-gwt:$version-jre` will work on Java 7, but that doesn't seem like a big risk. If we want to make it possible for users to use Guava by using only BOMs, that seems like a good enough reason to add it.

PR #349, opened 2019-01-17
StaticLinkageCheckOption to take option for maven repository

### LINKED ISSUE #338: StaticLinkageChecker to support other repositories than Maven central
When I tried to check [the issue of Google Cloud Pubsub with Spring Cloud GCP][1], StaticLinkageChecker was not able to locate the Maven artifact available in https://repo.spring.io/milestone.

https://github.com/shiraji/demo-pubsub/blob/master/build.gradle

```
Exception in thread "main" org.eclipse.aether.resolution.DependencyResolutionException: The following artifacts could not be resolved: org.springframework.cloud:spring-cloud-gcp-starter-pubsub:jar:1.1.0.RC2, org.springframework.cloud:spring-cloud-gcp-starter-trace:jar:1.1.0.RC2: Could not find artifact org.springframework.cloud:spring-cloud-gcp-starter-pubsub:jar:1.1.0.RC2 in central (http://repo1.maven.org/maven2/)
	at org.eclipse.aether.internal.impl.DefaultRepositorySystem.resolveDependencies(DefaultRepositorySystem.java:355)
	at com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder.resolveCompileTimeDependencies(DependencyGraphBuilder.java:125)
	at com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder.getStaticLinkageCheckDependencyGraph(DependencyGraphBuilder.java:158)
	at com.google.cloud.tools.opensource.classpath.ClassPathBuilder.artifactsToDependencyPaths(ClassPathBuilder.java:86)
	at com.google.cloud.tools.opensource.classpath.ClassPathBuilder.artifactsToClasspath(ClassPathBuilder.java:56)
	at com.google.cloud.tools.opensource.classpath.StaticLinkageCheckOption.generateInputClasspath(StaticLinkageCheckOption.java:115)
	at com.google.cloud.tools.opensource.classpath.StaticLinkageChecker.main(StaticLinkageChecker.java:95)
Caused by: org.eclipse.aether.resolution.ArtifactResolutionException: The following artifacts could not be resolved: org.springframework.cloud:spring-cloud-gcp-starter-pubsub:jar:1.1.0.RC2, org.springframework.cloud:spring-cloud-gcp-starter-trace:jar:1.1.0.RC2: Could not find artifact org.springframework.cloud:spring-cloud-gcp-starter-pubsub:jar:1.1.0.RC2 in central (http://repo1.maven.org/maven2/)
	at org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolve(DefaultArtifactResolver.java:422)
	at org.eclipse.aether.internal.impl.DefaultArtifactResolver.resolveArtifacts(DefaultArtifactResolver.java:224)
	at org.eclipse.aether.internal.impl.DefaultRepositorySystem.resolveDependencies(DefaultRepositorySystem.java:338)
	... 6 more
Caused by: org.eclipse.aether.transfer.ArtifactNotFoundException: Could not find artifact org.springframework.cloud:spring-cloud-gcp-starter-pubsub:jar:1.1.0.RC2 in central (http://repo1.maven.org/maven2/)
	at org.eclipse.aether.connector.basic.ArtifactTransportListener.transferFailed(ArtifactTransportListener.java:48)
	at org.eclipse.aether.connector.basic.BasicRepositoryConnector$TaskRunner.run(BasicRepositoryConnector.java:368)
	at org.eclipse.aether.util.concurrency.RunnableErrorForwarder$1.run(RunnableErrorForwarder.java:75)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

`DependencyGraphBuilder.resolveCompileTimeDependencies` has

```
    collectRequest.addRepository(RepositoryUtility.CENTRAL);
    CollectResult collectResult = system.collectDependencies(session, collectRequest);
    DependencyNode node = collectResult.getRoot();
```

[1]: https://github.com/spring-cloud/spring-cloud-gcp/issues/1373

suztomo:
Fixes #338

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
@@ -122,7 +122,7 @@ public class StaticLinkageCheckOption {
 
   static ImmutableList<Path> generateInputClasspath(CommandLine commandLine)
       throws RepositoryException, ParseException {
-    configureMavenRepositories(commandLine);
+    setRepositories(commandLine);
 
     if (commandLine.hasOption("b")) {
       String bomCoordinates = commandLine.getOptionValue("b");
@@ -149,7 +149,7 @@ public class StaticLinkageCheckOption {
     }
   }
 
-  static void configureMavenRepositories(CommandLine commandLine) throws ParseException {
+  static void setRepositories(CommandLine commandLine) throws ParseException {
     if (!commandLine.hasOption("m")) {
       return;
     }
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
@@ -19,27 +19,23 @@ package com.google.cloud.tools.opensource.classpath;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 
 import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
-import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Streams;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.URL;
+import com.google.common.collect.Lists;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Stream;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.CommandLineParser;
 import org.apache.commons.cli.DefaultParser;
 import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
 import org.eclipse.aether.RepositoryException;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.artifact.DefaultArtifact;
-import org.eclipse.aether.repository.RemoteRepository;
 
 /**
  * Option for {@link StaticLinkageChecker}. To construct an input class path, the checker requires
@@ -56,10 +52,6 @@ import org.eclipse.aether.repository.RemoteRepository;
  *     Linkage Checker: Input</a>
  */
 public class StaticLinkageCheckOption {
-
-  // DefaultTransporterProvider.newTransporter checks the transporters for repository URLs
-  private static final ImmutableSet<String> ALLOWED_REPOSITORY_URL_PROTOCOL =
-      ImmutableSet.of("file", "http", "https");
 
   private static final Options options = configureOptions();
 
@@ -89,26 +81,43 @@ public class StaticLinkageCheckOption {
 
   private static Options configureOptions() {
     Options options = new Options();
+
     options.addOption(
         "b", "bom", true, "BOM to generate a class path, specified by its Maven coordinates");
-    options.addOption(
-        "a",
-        "artifacts",
-        true,
-        "Maven coordinates for Maven artifacts (separated by ',') to generate a class path");
-    options.addOption("j", "jars", true, "Jar files (separated by ',') to generate a class path");
+
+    Option artifactOption = Option.builder("a").longOpt("artifacts").hasArgs()
+        .valueSeparator(',')
+        .desc("Maven coordinates for Maven artifacts (separated by ',') to generate a class path")
+        .build();
+    options.addOption(artifactOption);
+
+    Option jarOption = Option.builder("j").longOpt("jars").hasArgs()
+        .valueSeparator(',')
+        .desc("Jar files (separated by ',') to generate a class path")
+        .build();
+    options.addOption(jarOption);
+
     options.addOption(
         "r",
         "report-only-reachable",
         false,
         "To report only linkage errors reachable from entry point");
-    options.addOption(
-        "m",
-        "maven-repository",
-        true,
-        "Maven repository URL to search for dependencies. "
-            + "If this option is used one or more times, the repositories are added "
-            + "in order before the default Maven Central (http://repo1.maven.org/maven2/).");
+
+    Option repositoryOption = Option.builder("m").longOpt("maven-repositories").hasArgs()
+        .valueSeparator(',')
+        .desc("Maven repository URLs to search for dependencies. "
+            + "The repositories are added to a repository list in order before "
+            + "the default Maven Central (http://repo1.maven.org/maven2/).")
+        .build();
+    options.addOption(repositoryOption);
+
+    Option noMavenCentralOption = Option.builder("nm").longOpt("no-maven-central")
+        .hasArg(false)
+        .desc("Used with '-m'. Not to add Maven Central to the repository list. "
+            + "Useful when the tool is used where Maven central is inaccessible")
+        .build();
+    options.addOption(noMavenCentralOption);
+
     return options;
   }
 
@@ -116,26 +125,22 @@ public class StaticLinkageCheckOption {
       throws RepositoryException, ParseException {
     configureMavenRepositories(commandLine);
 
-    Splitter commaSplitter = Splitter.on(",");
-
     if (commandLine.hasOption("b")) {
       String bomCoordinates = commandLine.getOptionValue("b");
       DefaultArtifact bomArtifact = new DefaultArtifact(bomCoordinates);
       List<Artifact> artifactsInBom = RepositoryUtility.readBom(bomArtifact);
       return ClassPathBuilder.artifactsToClasspath(artifactsInBom);
     } else if (commandLine.hasOption("a")) {
-      String mavenCoordinatesOption = commandLine.getOptionValue("a");
+      String[] mavenCoordinatesOption = commandLine.getOptionValues("a");
       ImmutableList<Artifact> artifacts =
-          commaSplitter
-              .splitToList(mavenCoordinatesOption)
-              .stream()
+          Arrays.stream(mavenCoordinatesOption)
               .map(DefaultArtifact::new)
               .collect(toImmutableList());
       return ClassPathBuilder.artifactsToClasspath(artifacts);
     } else if (commandLine.hasOption("j")) {
-      String jarFiles = commandLine.getOptionValue("j");
+      String[] jarFiles = commandLine.getOptionValues("j");
       ImmutableList<Path> jarFilesInArguments =
-          Streams.stream(commaSplitter.split(jarFiles))
+          Arrays.stream(jarFiles)
               .map(name -> Paths.get(name).toAbsolutePath())
               .collect(toImmutableList());
       return jarFilesInArguments;
@@ -149,26 +154,13 @@ public class StaticLinkageCheckOption {
     if (!commandLine.hasOption("m")) {
       return;
     }
-
-    ImmutableList.Builder<RemoteRepository> repositoryListBulder = ImmutableList.builder();
-    for (String mavenRepositoryUrl : commandLine.getOptionValues("m")) {
-      RemoteRepository repository =
-          new RemoteRepository.Builder(mavenRepositoryUrl, "default", mavenRepositoryUrl).build();
-      if (!ALLOWED_REPOSITORY_URL_PROTOCOL.contains(repository.getProtocol())) {
-        helpFormatter.printHelp("StaticLinkageChecker", options);
-        throw new ParseException(
-            "Invalid URL specified for maven repository: " + mavenRepositoryUrl);
-      }
-      try {
-        // Because the protocol is not an empty string, this URI is absolute.
-        new URI(mavenRepositoryUrl);
-      } catch (URISyntaxException ex) {
-        throw new ParseException("Invalid URL syntax for repository: " + mavenRepositoryUrl);
-      }
-
-      repositoryListBulder.add(repository);
+    try {
+      boolean addMavenCentral = !commandLine.hasOption("nm");
+      RepositoryUtility.configureMavenRepositories(Arrays.asList(commandLine.getOptionValues("m")),
+          addMavenCentral);
+    } catch (IllegalArgumentException ex) {
+      throw new ParseException("Invalid URL specified for Maven repositories: "
+          + ex.getMessage());
     }
-    repositoryListBulder.add(RepositoryUtility.CENTRAL);
-    RepositoryUtility.mavenRepositories = repositoryListBulder.build();
   }
 }
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
@@ -23,6 +23,9 @@ import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Streams;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
@@ -155,6 +158,14 @@ public class StaticLinkageCheckOption {
         helpFormatter.printHelp("StaticLinkageChecker", options);
         throw new ParseException(
             "Invalid URL specified for maven repository: " + mavenRepositoryUrl);
+      }
+      try {
+        URI uri = new URI(mavenRepositoryUrl);
+        if (!uri.isAbsolute()) {
+          throw new ParseException("Repository URL must have an absolute path for file");
+        }
+      } catch (URISyntaxException ex) {
+        throw new ParseException("Invalid URL for repository: " + mavenRepositoryUrl);
       }
 
       repositoryListBulder.add(repository);
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
@@ -21,6 +21,7 @@ import static com.google.common.collect.ImmutableList.toImmutableList;
 import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
 import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Streams;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -35,6 +36,7 @@ import org.apache.commons.cli.ParseException;
 import org.eclipse.aether.RepositoryException;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.artifact.DefaultArtifact;
+import org.eclipse.aether.repository.RemoteRepository;
 
 /**
  * Option for {@link StaticLinkageChecker}. To construct an input class path, the checker requires
@@ -51,13 +53,17 @@ import org.eclipse.aether.artifact.DefaultArtifact;
  *     Linkage Checker: Input</a>
  */
 public class StaticLinkageCheckOption {
-  
+
+  // DefaultTransporterProvider.newTransporter checks the transporters for repository URLs
+  private static final ImmutableSet<String> ALLOWED_REPOSITORY_URL_PROTOCOL =
+      ImmutableSet.of("file", "http", "https");
+
   private static final Options options = configureOptions();
 
   private static final HelpFormatter helpFormatter = new HelpFormatter();
 
   static CommandLine readCommandLine(String[] arguments) throws ParseException {
-    // TODO is this reentrant? Can we reuse it? 
+    // TODO is this reentrant? Can we reuse it?
     // https://issues.apache.org/jira/browse/CLI-291
     CommandLineParser parser = new DefaultParser();
 
@@ -95,9 +101,11 @@ public class StaticLinkageCheckOption {
         "To report only linkage errors reachable from entry point");
     options.addOption(
         "m",
-        "maven-repositories",
+        "maven-repository",
         true,
-        "Maven repositories (separated by ',') to search dependencies");
+        "Maven repository URL to search for dependencies. "
+            + "If this option is used one or more times, the repositories are added "
+            + "in order before the default Maven Central (http://repo1.maven.org/maven2/).");
     return options;
   }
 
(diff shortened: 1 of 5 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -161,7 +165,10 @@ private static File findLocalRepository() {
     RepositorySystemSession session = RepositoryUtility.newSession(system);
 
     ArtifactDescriptorRequest request = new ArtifactDescriptorRequest();
-    request.addRepository(RepositoryUtility.CENTRAL);
+
+    for (RemoteRepository repository : mavenRepositories) {
```
suztomo: I was thinking to use addRepositoryToRequest here too. But `ArtifactDescriptorRequest` and `CollectRequest` don't share a class or interface. AS line 169 is within `RepositoryUtility`, I let this line access `mavenRepositories` directly.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -59,6 +61,8 @@
   public static final RemoteRepository CENTRAL =
       new RemoteRepository.Builder("central", "default", "http://repo1.maven.org/maven2/").build();
 
+  private static final List<RemoteRepository> mavenRepositories = Lists.newArrayList(CENTRAL);
```
netdpb: Let's avoid static mutable state if we can. Can the command line parser just create an `ImmutableList` of repositories and put it here, rather than having a mutable one?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -198,4 +205,14 @@ private static File findLocalRepository() {
     return managedDependencies;
   }
 
+  public static void registerMavenRepository(String mavenRepositoryUrl) {
```
elharo: I'm having trouble following all the code in Github's poor excuse for a code review tool. However I am surprised that this method does not throw any checked exceptions.  What if the string is not a correct URL?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -93,11 +93,18 @@ private static Options configureOptions() {
         "report-only-reachable",
         false,
         "To report only linkage errors reachable from entry point");
+    options.addOption(
+        "m",
+        "maven-repositories",
+        true,
+        "Maven repositories (separated by ',') to search dependencies");
```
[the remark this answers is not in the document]
suztomo: Added description for order.

Format is a URL.

Added description on Maven central is in the list by default.

So far I don't see use case of excluding Maven central but it would be easy to implement when it's needed.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -126,4 +133,14 @@ private static Options configureOptions() {
       throw new ParseException("Missing argument");
     }
   }
+
+  static void configureMavenRepositories(CommandLine commandLine) {
+    if (!commandLine.hasOption("m")) {
+      return;
+    }
+    Splitter commaSplitter = Splitter.on(",");
+    for (String mavenRepositoryUrl : commaSplitter.splitToList(commandLine.getOptionValue("m"))) {
```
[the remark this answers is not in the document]
suztomo: Updated to use `getOptionValues`. It does not split the value but it lets users to specify the option multiple times in a command line.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
suztomo: Updated to
`public static ImmutableList<RemoteRepository> mavenRepositories = ImmutableList.of(CENTRAL);`

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
suztomo: It does not cause an exception when creating `RemoteRepository` instances.

When fetching dependencies, `DefaultTransporterProvider.newTransporter` searches for matching transporters (HttpTransporter and FileTransporter are there) for its protocol.

So I've added validation to check protocol to be `file`, `http` or `https`.

suztomo:
@elharo @netdpb Applied code review. PTAL.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -126,4 +141,25 @@ private static Options configureOptions() {
...
+  static void configureMavenRepositories(CommandLine commandLine) throws ParseException {
+    if (!commandLine.hasOption("m")) {
+      return;
+    }
+
+    ImmutableList.Builder<RemoteRepository> repositoryListBulder = ImmutableList.builder();
+    for (String mavenRepositoryUrl : commandLine.getOptionValues("m")) {
+      RemoteRepository repository =
+          new RemoteRepository.Builder(mavenRepositoryUrl, "default", mavenRepositoryUrl).build();
+      if (!ALLOWED_REPOSITORY_URL_PROTOCOL.contains(repository.getProtocol())) {
```
elharo: there are other ways this could fail. E.g. the input could be "_http_file__https" 

You should probably try to build a java.net.URI object and look for URISyntaxExceptions. You might need to require an absolute URL too.

suztomo: RemoteRepository parses the URL string. For  ‘_http_file__https’, getProtocol method should return an empty string.


I will add the URISyntaxException check.

suztomo: Checking absolute is unnecessary as line 157 already checks the presence of the schema (protocol)

https://docs.oracle.com/javase/8/docs/api/java/net/URI.html#isAbsolute--

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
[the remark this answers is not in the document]
netdpb: Hmm. `Option.Builder.valueSeparator(char)` and `Option.Builder.numberOfArgs(UNLIMITED_VALUES)` look like they let you have an option like "-mfoo,bar,baz"

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -51,13 +56,17 @@
  *     Linkage Checker: Input</a>
  */
 public class StaticLinkageCheckOption {
-  
+
+  // DefaultTransporterProvider.newTransporter checks the transporters for repository URLs
+  private static final ImmutableSet<String> ALLOWED_REPOSITORY_URL_PROTOCOL =
```
netdpb: These are technically URL "schemes".

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -134,13 +145,30 @@ private static Options configureOptions() {
...
       return;
     }
-    Splitter commaSplitter = Splitter.on(",");
-    for (String mavenRepositoryUrl : commaSplitter.splitToList(commandLine.getOptionValue("m"))) {
-      RepositoryUtility.registerMavenRepository(mavenRepositoryUrl);
+
+    ImmutableList.Builder<RemoteRepository> repositoryListBulder = ImmutableList.builder();
+    for (String mavenRepositoryUrl : commandLine.getOptionValues("m")) {
+      RemoteRepository repository =
+          new RemoteRepository.Builder(mavenRepositoryUrl, "default", mavenRepositoryUrl).build();
```
netdpb: The first parameter is supposed to be the ID, and it can be null. Should we pass null?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -134,13 +145,30 @@ private static Options configureOptions() {
     }
   }
 
-  static void configureMavenRepositories(CommandLine commandLine) {
+  static void configureMavenRepositories(CommandLine commandLine) throws ParseException {
```
netdpb: It would be better if this method were on `RepositoryUtility`. Then the non-final field could be private, and you could have a static method that returns it, so that we don't have a public static non-final field.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOptionTest.java
```diff
@@ -81,4 +89,32 @@ public void testConfigureAdditionalMavenRepositories_addingSpringRepository()
     List<Path> paths = ClassPathBuilder.artifactsToClasspath(ImmutableList.of(artifact));
     Truth.assertThat(paths).isNotEmpty();
   }
+
+  @Test
+  public void testConfigureAdditionalMavenRepositories_invalidRepositoryUrl()
+      throws ParseException {
+
+    List<String> invalidUrls =
```
netdpb: I think it would be better to have a private method like `assertMavenRepositoryIsInvalid(String)`, and to have this test call it explicitly:

```
assertMavenRepositoryIsInvalid("foobar");
assertMavenRepositoryIsInvalid("_http_file__https");
// etc.
```

That way the failure would point to the exact line number that failed.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
suztomo: Set to null.

I had thought setting id with URL would give better error message, but `RemoteRepository.toString` outputs its URL and id. So null is just fine.

```
Could not find artifact org.springframework:nosuchartifact:jar:3.1.0.RC2 in  (https://repo.spring.io/milestone)
```

When null is passed to id, then id is set to an empty string.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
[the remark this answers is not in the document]
suztomo: I came up with a use case where Maven Central is inaccessible. So added `no-maven-central` option.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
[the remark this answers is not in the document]
suztomo: Thanks. That's nicer API. Now it uses `Option.Builder`.

`hasArgs` (with "s") takes care of `numberOfArgs(UNLIMITED_VALUES)`

suztomo:
@netdpb @elharo Applied code review. PTAL.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -198,4 +213,42 @@ private static File findLocalRepository() {
     return managedDependencies;
   }
 
+  /**
+   * Configures Maven repositories to search for dependencies.
+   *
+   * @param addMavenCentral if true, add Maven Central to the end of the repository list
+   */
+  public static void configureMavenRepositories(Iterable<String> mavenRepositoryUrls,
```
elharo: I don't think we're likely to be looking at a non-Maven repository, so perhaps just configureRepositories(urls)


Also, this sounds like you're configuring the repositories themselves, which we're not. So maybe `setRepositories`

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -198,4 +213,42 @@ private static File findLocalRepository() {
...
+      boolean addMavenCentral) {
+    ImmutableList.Builder<RemoteRepository> repositoryListBuilder = ImmutableList.builder();
+    for (String mavenRepositoryUrl : mavenRepositoryUrls) {
+      RemoteRepository repository =
+          new RemoteRepository.Builder(null, "default", mavenRepositoryUrl).build();
+      if (!ALLOWED_REPOSITORY_URL_SCHEMES.contains(repository.getProtocol())) {
+        throw new IllegalArgumentException(
```
elharo: need javadoc for this on a public method

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -198,4 +213,42 @@ private static File findLocalRepository() {
...
+            "Protocol: " + repository.getProtocol() + " is not in "
+                + ALLOWED_REPOSITORY_URL_SCHEMES);
+      }
+      try {
```
elharo: Should check this at the top of the for loop before other work is done

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -198,4 +213,42 @@ private static File findLocalRepository() {
     return managedDependencies;
   }
 
+  /**
+   * Configures Maven repositories to search for dependencies.
+   *
+   * @param addMavenCentral if true, add Maven Central to the end of the repository list
```
elharo: since this parameter is always true, we don't need it

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -59,6 +65,12 @@
   public static final RemoteRepository CENTRAL =
       new RemoteRepository.Builder("central", "default", "http://repo1.maven.org/maven2/").build();
 
+  private static ImmutableList<RemoteRepository> mavenRepositories = ImmutableList.of(CENTRAL);
```
elharo: Mutable static state suggests we're growing past the point where this class can plausibly be static. OK, for now but consider whether refactoring might help. Either to use more instance methods or to pass the repsositoryURLs as arguments to a method instead.

suztomo: Created #353

suztomo:
@elharo Applied review and responded to comment. PTAL.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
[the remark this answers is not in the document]
netdpb: Would it be simpler to just have the default for -m be Maven Central, and if any repository is specified then Maven Central must be as well?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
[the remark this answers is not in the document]
netdpb: Ah, good.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -89,53 +80,66 @@ private static void checkInput(CommandLine commandLine) throws ParseException {
...
-        "Maven coordinates for Maven artifacts (separated by ',') to generate a class path");
-    options.addOption("j", "jars", true, "Jar files (separated by ',') to generate a class path");
+
+    Option artifactOption = Option.builder("a").longOpt("artifacts").hasArgs()
+        .valueSeparator(',')
+        .desc("Maven coordinates for Maven artifacts (separated by ',') to generate a class path")
+        .build();
+    options.addOption(artifactOption);
+
+    Option jarOption = Option.builder("j").longOpt("jars").hasArgs()
```
netdpb: Might be worth a little method to create a comma-separated option that takes the short name, the long name, and the description.

`options.addOption(multiValueOption("j", "jars", "Jar files...."));`

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -89,53 +80,66 @@ private static void checkInput(CommandLine commandLine) throws ParseException {
...
+        .valueSeparator(',')
+        .desc("Maven repository URLs to search for dependencies. "
+            + "The repositories are added to a repository list in order before "
+            + "the default Maven Central (http://repo1.maven.org/maven2/).")
+        .build();
+    options.addOption(repositoryOption);
+
+    Option noMavenCentralOption = Option.builder("nm").longOpt("no-maven-central")
+        .hasArg(false)
+        .desc("Used with '-m'. Not to add Maven Central to the repository list. "
```
netdpb: Instead of "Used with '-m'. Not to add Maven Central to the repository list.", let's say "Do not search Maven Central in addition to the repositories specified by -m."

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
```diff
@@ -89,53 +80,66 @@ private static void checkInput(CommandLine commandLine) throws ParseException {
...
+            + "Useful when the tool is used where Maven central is inaccessible")
```
netdpb: "Maven Central" (capital C).

Also, you can delete "the tool is used where": "Useful when Maven Central is inaccessible".

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
netdpb: If you wanted to be fancy, you could check all the URLs and throw an exception that lists all the invalid ones in one message. Not necessary for now, but maybe nicer for the user.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -205,7 +212,44 @@ private static File findLocalRepository() {
...
+      try {
+        // Because the protocol is not an empty string, this URI is absolute.
+        new URI(mavenRepositoryUrl);
+      } catch (URISyntaxException ex) {
+        throw new IllegalArgumentException("Invalid URL syntax: " + mavenRepositoryUrl);
+      }
+
+      RemoteRepository repository =
+          new RemoteRepository.Builder(null, "default", mavenRepositoryUrl).build();
+      if (!ALLOWED_REPOSITORY_URL_SCHEMES.contains(repository.getProtocol())) {
```
netdpb:
```java
checkArgument(
    ALLOWED_REPOSITORY_URL_SCHEMES.contains(repository.getProtocol()),
    "Scheme: '%s' is not in %s", repository.getProtocol(), ALLOWED_REPOSITORY_URL_SCHEMES);
```

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -205,7 +212,44 @@ private static File findLocalRepository() {
...
+        throw new IllegalArgumentException(
+            "Schema: '"
```
netdpb: Scheme, not schema.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -205,7 +212,44 @@ private static File findLocalRepository() {
...
+
+    if (addMavenCentral) {
+      repositoryListBuilder.add(CENTRAL);
+    }
+
+    mavenRepositories = repositoryListBuilder.build();
+  }
+
+  /** Adds {@link #mavenRepositories} to {@code collectRequest}. */
+  public static void addRepositoriesToRequest(CollectRequest collectRequest) {
```
netdpb: Great idea.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOptionTest.java
```diff
@@ -90,22 +112,40 @@ public void testConfigureAdditionalMavenRepositories_addingSpringRepository()
...
-      try {
-        StaticLinkageCheckOption.configureMavenRepositories(commandLine);
-        Assert.fail("URL " + invalidUrl + " should be invalidated");
-      } catch (ParseException ex) {
-        // pass
-      }
+  private static void assertMavenRepositoryIsInvalid(String repositoryUrl)
+      throws ParseException {
+    CommandLine commandLine =
+        StaticLinkageCheckOption.readCommandLine(new String[]{"-m", repositoryUrl});
```
netdpb: If `readCommandLine` took a varargs of String, these tests would be simpler because you wouldn't need all the explicit arrays.

`readCommandLine("-m", repositoryUrl)`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
suztomo: I'd say this Option.Builder API is more explicit on the argument name.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOption.java
suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
suztomo: That's clear. Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
suztomo: Fixed

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOptionTest.java
suztomo: That's nice. Changed the method to take varargs.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
suztomo: That's a nice idea. I think `ClassPathBuilder` can benefit from the idea more than this repository URLs. As of now  `ClassPathBuilder` fails immediately for one unavailable Maven artifact without telling how many failures need to be fixed.

Created #355.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/StaticLinkageCheckOptionTest.java
```diff
@@ -117,7 +117,7 @@ public void testConfigureAdditionalMavenRepositories_notToUseMavenCentral()
       throws ParseException {
     CommandLine commandLine =
         StaticLinkageCheckOption.readCommandLine(
-            new String[] {"-m", "https://repo.spring.io/milestone", "--no-maven-central"});
+            "-m", "https://repo.spring.io/milestone", "--no-maven-central");
```
netdpb: Look throughout the file for other places to do this instead of creating a `String[]` local variable.

suztomo: Removed the local variables.

PR #365, opened 2019-01-24
Added description of Maven Dependency Graph

### LINKED ISSUE #364: Document Dependency Graph that is used to generate a class path
Document Dependency Graph that is used to generate a class path

suztomo:
Fixes #364

The formatted markdown is available in https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/graph_doc/dependencies/README.md#maven_dependency_graph

### DIFF
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -110,13 +110,13 @@ A Maven dependency graph is a graph data structure where
 Given a ordered list of Maven artifacts, a Maven dependency graph is constructed in the
 following manner:
 
-- 1. Start with a graph with nodes of the Maven artifacts in the list.
-     The nodes are called _initial nodes_.
-- 2. Pick up a node in a graph in breadth-first manner.
-- 3. By reading dependencies of the node, add new nodes corresponding to the target Maven artifacts,
-     identified by `groupId:artifactId[:classifier]:version`, if not present.
-- 4. Add edges from the source node to the target nodes of the Maven artifacts.
-- 3. Repeat step 2-4, until all nodes are visited in this breadth-first traversal.
+1. Start with a graph with nodes of the Maven artifacts in the list.
+   The nodes are called _initial nodes_.
+2. Pick up a node in a graph in breadth-first manner.
+3. By reading dependencies of the node, add new nodes corresponding to the target Maven artifacts,
+   identified by `groupId:artifactId[:classifier]:version`, if not present.
+4. Add edges from the source node to the target nodes of the Maven artifacts.
+3. Repeat step 2-4, until all nodes are visited in this breadth-first traversal.
 
 A graph construction may _fail_ when there is a problem in constructing a graph (see below).
 
--- a/dependencies/README.md
+++ b/dependencies/README.md
@@ -92,11 +92,12 @@ The tool allows users to choose the scope of entry point classes:
     are reported.
 
 <a name="maven_dependency_graph"></a>
-## Maven Dependency Graph
+# Maven Dependency Graph
 
 A Maven dependency graph is a graph data structure where
 - Node: a node is a Maven artifact identified by [Maven coordinates][1] such as
-  `com.example.foo:bar:1.5.4`.
+  `com.example.foo:bar:1.5.4`. A node may be marked _unavailable_ when the artifact is not available
+  in Maven repositories.
 - Edge: a directed edge is a dependency from one Maven artifact (_source_ of the edge)
   to another Maven artifact (_target_ of the edge).
 
@@ -105,7 +106,7 @@ A Maven dependency graph is a graph data structure where
 
   Self-loops are not possible. A parallel edge is allowed but is dropped in the model.
 
-### Comparison with Maven Dependency Trees by pom.xml
+## Comparison with Maven Dependency Trees by pom.xml
 
 Compared to a _Maven dependency tree_ retrieved by a pom.xml through
 `RepositorySystem.resolveDependencies`, a Maven Dependency Graph has the following differences:
@@ -117,50 +118,51 @@ Compared to a _Maven dependency tree_ retrieved by a pom.xml through
 - A Maven dependency tree does not contain a cycle,
   whereas a Maven dependency graph can contain cycles.
 
-### Graph Construction
+## Graph Construction
 
 Given an ordered list of Maven artifacts, `DependencyGraphBuilder` constructs a Maven dependency
-graph in the following manner:
+graph in the following steps:
 
 1. Start with a graph with nodes of the Maven artifacts in the list.
    The nodes are called _initial nodes_.
-2. Pick up an unvisited node in a graph in breadth-first manner.
+2. Pick up an unvisited node which is not marked _unavailable_ in a graph in breadth-first manner.
 3. By reading dependencies of the node,
    - add new nodes corresponding to the target Maven artifacts, identified by Maven coordinates,
      if not present.
    - add edges from the source node to the target nodes of the Maven artifacts.
-4. Repeat step 2-3, until all nodes are visited in the breadth-first traversal.
+4. Repeat Step 2-3, until all nodes are visited in the breadth-first traversal.
 
-A graph construction may _fail_ when there is a problem in constructing a graph (see below).
+## Graph Construction Failure
 
-### Edge Cases
+A graph construction may _fail_ when there is a problem.
 
-#### Unavailable Artifact
+### Unavailable Artifact
 
-A Maven artifact may be unavailable through Maven repositories.
+A Maven artifact may be unavailable through Maven repositories. Such nodes are marked unavailable
+at Step 3 of a graph construction.
 
 An unavailability of Maven artifact is called _safe_ when the path from the initial nodes to
-the missing artifact contains `optional` or `scope: provided` dependency. An edge whose source
-artifact is missing but is safe, is skipped in a graph construction (step 4). The resulting graph
-has the unavailable Maven artifact node, which is not connected to other nodes.
+the missing artifact contains `optional` or `scope: provided` dependency. The resulting graph
+has the unavailable Maven artifact node connected to another node.
 
 When there is an unavailable Maven artifact and it is not safe, the graph construction fails.
 
-#### Unsatisfied Version Constraints
+### Unsatisfied Version Constraints
 
 A dependency element in pom.xml may have a [version range specification][4].
 Each of the specifications creates a version constraint.
 
-When there is a version constraint that cannot be satisfied during graph construction,
-the graph construction fails.
+When there is a version constraint that cannot be satisfied, the graph construction fails.
 
-### Class Path Generation through Maven Dependency Graph
+## Class Path Generation through Maven Dependency Graph
 
 A class path (list of jar files of Maven artifacts) can be generated from a Maven dependency graph.
-A class path is built by picking up Maven artifacts by a breadth-first traversal,
-starting from the initial nodes of a Maven dependency graph.
+A class path is built by picking up Maven artifacts by a breadth-first
+traversal. The traversal starts from the initial nodes of a Maven dependency graph.
 
-During the pick-up, duplicate artifacts identified by Maven coordinates are discarded.
+During the pick-up,
+- duplicate artifacts identified by Maven coordinates are discarded.
+- nodes marked as unavailable are skipped.
 
 When there are multiple versions of a Maven artifact identified by Maven coordinates without
 the version part, a version is picked up using one of the following strategies:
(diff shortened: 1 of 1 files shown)


dependencies/README.md
```diff
@@ -87,3 +89,74 @@ The tool allows users to choose the scope of entry point classes:
...
+  A dependency has a boolean attribute `optional` and a string attribute `scope`,
+  among other properties listed in [POM Reference: Dependencies][1].
+
+  Self-loops or parallel edges are not possible.
+
+### Graph Construction
+
+Given a list of Maven artifacts, a Maven dependency graph is defined in the following manner:
+
+- Create a graph with nodes of the Maven artifacts. They are called initial nodes.
```
elharo: I don't quite follow this paragraph. In particular edges are added from what *to what*?

dependencies/README.md
```diff
@@ -87,3 +89,74 @@ The tool allows users to choose the scope of entry point classes:
...
+
+#### Cyclic Dependency
+
+Cyclic dependency may exist in a Maven dependency graph. It does not become a cause of an undefined
+graph.
+
+#### Unavailable Artifact
+
+A Maven artifact may be unavailable through Maven repositories, making it impossible to complete
+a graph construction. Such unavailability is said of _safe_ when the path from the initial nodes to
```
elharo: said of --> called

dependencies/README.md
```diff
@@ -87,3 +89,74 @@ The tool allows users to choose the scope of entry point classes:
...
+the missing artifact contains both optional and `scope: provided` dependency. An edge whose source
```
elharo: It has to be optional and provided? not optional or provided?

dependencies/README.md
```diff
@@ -87,3 +89,74 @@ The tool allows users to choose the scope of entry point classes:
...
+artifact is missing but acceptable is skipped in a graph construction.
+
+When there is an unavailable Maven artifact and it is not safe, the Maven dependency graph is
+undefined.
+
+#### Unsatisfied Version Constraints
+
+A dependency tag in pom.xml may have a [version range specification][2].
```
elharo: tag --> element

dependencies/README.md
```diff
@@ -87,3 +89,74 @@ The tool allows users to choose the scope of entry point classes:
...
+Each of the specifications creates a version constraint.
+When there is a version constraint that cannot be satisfied during graph construction,
+the Maven dependency graph is undefined.
```
elharo: is undefined really right here? It sounds like there is no dependency graph in this case. That is, Maven itself will fail rather than construct something.

dependencies/README.md
```diff
@@ -87,3 +89,74 @@ The tool allows users to choose the scope of entry point classes:
...
+
+### Class Path Generation through Maven Dependency Graph
+
+A class path (list of jar files of Maven artifacts) can be generated from a Maven dependency graph.
+A class path is built by picking up Maven artifacts in breadth-first manner,
+starting from the first node of the initial nodes of a Maven dependency graph.
+
+During the pick-up, duplicated artifacts identified by
```
elharo: duplicated --> duplicate

dependencies/README.md
```diff
@@ -87,3 +89,74 @@ The tool allows users to choose the scope of entry point classes:
...
+`groupId:artifactId[:classifier]:version` are discarded.
+
+When there are multiple versions of a Maven artifact
+identified by `groupId:artifactId[:classifier]` (without version), a version is picked up
+using one of following strategies:
+
+- **Maven remediation strategy**: the version of the Maven artifact closest to the initial nodes
```
elharo: remediation --> dependency mediation 
since "dependency mediation" is the term the maven docs use.

dependencies/README.md
```diff
@@ -87,3 +89,80 @@ The tool allows users to choose the scope of entry point classes:
...
+- Node: a node is a Maven artifact identified as `groupId:artifactId[:classifier]:version`, where
+  `classifier` is optional.
+- Edge: an (directed) edge is a dependency between Maven artifacts. A dependency from a Maven
+  artifact (_source_ of the dependency) to another artifact (_target_ of the dependency) is defined
+  in the `dependencies` element in pom.xml of a Maven artifact.
+  
+  A dependency has a boolean attribute `optional` and a string attribute `scope`,
+  among other properties listed in [POM Reference: Dependencies][1].
+
+  Self-loops or parallel edges are not possible.
```
elharo: Why not? Certainly this could be encoded in the XML. I think there are two cases here:

1. A self-loop is, I think, a compile time error and no graph will be constructed. 
2. A parallel edge is allowed but is dropped in the model.

dependencies/README.md
```diff
@@ -87,3 +89,80 @@ The tool allows users to choose the scope of entry point classes:
...
+
+Given a ordered list of Maven artifacts, a Maven dependency graph is constructed in the
+following manner:
+
+1. Start with a graph with nodes of the Maven artifacts in the list.
+   The nodes are called _initial nodes_.
+2. Pick up an unvisited node in a graph in breadth-first manner.
+3. By reading dependencies of the node, add new nodes corresponding to the target Maven artifacts,
+   identified by `groupId:artifactId[:classifier]:version`, if not present.
+4. Add edges from the source node to the target nodes of the Maven artifacts.
```
elharo: This doesn't sound quite right. You're not constructing the edges until all nodes are found, but then how do you find the nodes?

dependencies/README.md
```diff
@@ -87,3 +89,80 @@ The tool allows users to choose the scope of entry point classes:
...
+  in the `dependencies` element in pom.xml of a Maven artifact.
+  
+  A dependency has a boolean attribute `optional` and a string attribute `scope`,
+  among other properties listed in [POM Reference: Dependencies][1].
+
+  Self-loops or parallel edges are not possible.
+
+### Graph Construction
+
+Given a ordered list of Maven artifacts, a Maven dependency graph is constructed in the
```
elharo: Who is constructing this graph? How does it differ from what's already in the pom.xml's of the artifacts? There's something I'm not quite following here.

dependencies/README.md
```diff
@@ -87,3 +89,80 @@ The tool allows users to choose the scope of entry point classes:
...
+4. Add edges from the source node to the target nodes of the Maven artifacts.
+3. Repeat the step 2-4, until all nodes are visited in the breadth-first traversal.
+
+A graph construction may _fail_ when there is a problem in constructing a graph (see below).
+
+### Edge Cases
+
+#### Cyclic Dependency
+
+Cyclic dependency may exist in a Maven dependency graph. It does not become a cause of a graph
```
elharo: Perhaps just "Maven dependency graphs can contain cycles." 
Then explain what Maven does with such graphs, and what we do with them.
Do we ever construct a graph that contains a cycle or do we break the cycle?

The confusion may be that there are *two* related but different graphs in play here. The one in the pom.xml's themselves and the one we construct.

dependencies/README.md
```diff
@@ -87,3 +89,80 @@ The tool allows users to choose the scope of entry point classes:
...
+construction failure.
+
+#### Unavailable Artifact
+
+A Maven artifact may be unavailable through Maven repositories, making it impossible to complete
```
elharo: "complete a graph construction" --> construct a graph

dependencies/README.md
```diff
@@ -87,3 +89,80 @@ The tool allows users to choose the scope of entry point classes:
...
+a graph construction.
```
elharo: I may have been wrong about this case. We could construct a graph here but could mark the node as unavailable somehow  That is, not require the existence of the pom.xml to infer the existence of the node.

Obviously, this still indicates something wrong with the graph we want to report.

dependencies/README.md
```diff
@@ -87,3 +89,80 @@ The tool allows users to choose the scope of entry point classes:
...
+starting from the initial nodes of a Maven dependency graph.
+
+During the pick-up, duplicate artifacts identified by `groupId:artifactId[:classifier]:version`
+are discarded.
+
+When there are multiple versions of a Maven artifact identified by
+`groupId:artifactId[:classifier]` (without version), a version is picked up using one of following
+strategies:
+
+- **Maven dependency mediation strategy**: the version of the Maven artifact closest to the initial
```
elharo: "closest" --> first version encountered during the breadth first traversal

for precision

dependencies/README.md
suztomo: Updated to 
> `DependencyGraphBuilder` constructs a Maven dependency
graph in the following steps

dependencies/README.md
suztomo: Updated step 3 to clarify the 2 and 3 are tied together.

dependencies/README.md
suztomo: Added a notion of "unavailable" marker.

dependencies/README.md
```diff
@@ -88,3 +90,89 @@ The tool allows users to choose the scope of entry point classes:
...
+
+  A dependency has a boolean attribute `optional` and an enum attribute `scope`,
+  among other properties listed in [POM Reference: Dependencies][2].
+
+  Self-loops are not possible. A parallel edge is allowed but is dropped in the model.
+
+## Comparison with Maven Dependency Trees by pom.xml
+
+Compared to a _Maven dependency tree_ retrieved by a pom.xml through
+`RepositorySystem.resolveDependencies`, a Maven Dependency Graph has the following differences:
```
elharo: "dependency graph" is inconsistently capitalized

dependencies/README.md
```diff
@@ -88,3 +90,89 @@ The tool allows users to choose the scope of entry point classes:
...
+## Graph Construction Failure
+
+A graph construction may _fail_ when there is a problem.
+
+### Unavailable Artifact
+
+A Maven artifact may be unavailable through Maven repositories. Such nodes are marked unavailable
+at Step 3 of a graph construction.
+
+An unavailability of Maven artifact is called _safe_ when the path from the initial nodes to
```
elharo: I'm not sure "safe" gives the right impression here. "Unnecessary"? Optional would be better but it's taken. Maybe just optional or provided?

suztomo: Updated to use 'optional' and 'provided'

dependencies/README.md
```diff
@@ -88,3 +90,90 @@ The tool allows users to choose the scope of entry point classes:
...
+## Comparison with Maven Dependency Trees by pom.xml
+
+Compared to a _Maven dependency tree_ retrieved by a pom.xml through
+`RepositorySystem.resolveDependencies`, a Maven dependency graph has the following differences:
+
+- A Maven dependency tree cannot have multiple versions of a Maven artifact, whereas a Maven
+  dependency graph can have them.
+- A Maven dependency tree does not have transitive `scope: provided` dependencies under
+  a `scope: provided` dependency, whereas a Maven dependency graph can have them.
+- A Maven dependency tree does not contain a cycle,
```
elharo: Really, how?

dependencies/README.md
```diff
@@ -88,3 +90,90 @@ The tool allows users to choose the scope of entry point classes:
...
+  among other properties listed in [POM Reference: Dependencies][2].
+
+  Self-loops are not possible. A parallel edge is allowed but is dropped in the model.
+
+## Comparison with Maven Dependency Trees by pom.xml
+
+Compared to a _Maven dependency tree_ retrieved by a pom.xml through
+`RepositorySystem.resolveDependencies`, a Maven dependency graph has the following differences:
+
+- A Maven dependency tree cannot have multiple versions of a Maven artifact, whereas a Maven
```
elharo: So you're assuming this is a post-mediation tree as opposed to a pre-mediation tree?

dependencies/README.md
suztomo: A Maven dependency tree (post mediation) does not contain a cycle.

For example, a cycle in a dependency graph is dom4j and jaxen:
[dom4j:dom4j:jar:1.5.2](https://search.maven.org/artifact/dom4j/dom4j/1.5.2/jar) depends on [jaxen:jaxen:jar:1.1-beta-4](https://search.maven.org/artifact/jaxen/jaxen/1.1-beta-4/jar), which depends on dom4j:dom4j:jar:1.5.2. This is a cycle in a graph.

However, the Maven dependency tree (post-mediation) does not show a cycle, because such node is omitted.

```
$ mvn dependency:tree -Dverbose
[INFO] com.google.suztomo:cycle:jar:1.0-SNAPSHOT
[INFO] \- dom4j:dom4j:jar:1.5.2:compile
[INFO]    +- jaxme:jaxme-api:jar:0.3:compile
[INFO]    +- (dom4j:dom4j:jar:1.5.2:compile - omitted for cycle)
[INFO]    +- jaxen:jaxen:jar:1.1-beta-4:compile
[INFO]    +- msv:xsdlib:jar:20030807:compile
[INFO]    +- msv:relaxngDatatype:jar:20030807:compile
[INFO]    +- pull-parser:pull-parser:jar:2:compile
[INFO]    +- xpp3:xpp3:jar:1.1.3.3:compile
[INFO]    +- stax:stax-api:jar:1.0:compile
[INFO]    \- xml-apis:xml-apis:jar:1.0.b2:compile
```
(`dom4j:dom4j` appears only once)

suztomo:
Closing for now, in favor of https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/385

PR #370, opened 2019-01-26
To be able to build a class path for spring-cloud-gcp-starter-pubsub

### LINKED ISSUE #367: Investigate the slowness in building a class path for spring-cloud-gcp-starter-pubsub
To reproduce, here is argument for StaticLinkageChecker:

```
-m
https://repo.spring.io/milestone
-a
org.springframework.cloud:spring-cloud-gcp-starter-pubsub:1.1.0.RC2
```

This takes forever (more than 5 minutes, before I give up)

# Investigation

## BFS loop

When checked with debugger, breadth-first iteration in `DependencyGraphBuilder.levelOrder` has more than 20,000 items in the queue. Something seems wrong.

image

## Does this happen with Maven central?

Maven central has 1.1.0.RELEASE. It can cause the problem.

Arguments:
```
-a org.springframework.cloud:spring-cloud-gcp-starter-pubsub:1.1.0.RELEASE
```

suztomo:
Fixes #367

Now StaticLinkageChecker can run against a heavy dependency graph by `org.springframework.cloud:spring-cloud-gcp-starter-pubsub:1.1.0.RELEASE`. This builds a class path of 1466 jar files.

- `DependencyGraphBuilder.levelOrder` not to visit same node twice when called by StaticLinkageChecker.
- `DependencyGraphBuilder.isSafeResolutionException` checks the condition to abort graph building as per https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/365#discussion_r250581145
- `ClassDumper.isUnusedClassSymbolReference` to return true if the target class is not found in source class.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -35,6 +35,7 @@ import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Logger;
 import javax.annotation.Nullable;
 import org.apache.bcel.Const;
 import org.apache.bcel.classfile.Attribute;
@@ -71,6 +72,8 @@ import org.apache.bcel.util.Repository;
  * in them, through the input class path for a static linkage check.
  */
 class ClassDumper {
+
+  private static final Logger logger = Logger.getLogger(ClassDumper.class.getName());
 
   private final ImmutableList<Path> inputClassPath;
   private final Repository classRepository;
@@ -551,9 +554,15 @@ class ClassDumper {
 
       ImmutableSet<Integer> targetConstantPoolIndices =
           constantPoolIndexForClass(sourceJavaClass, targetClassName);
-      Verify.verify(
-          !targetConstantPoolIndices.isEmpty(),
-          "The target class symbol reference is not found in source class");
+      if (targetConstantPoolIndices.isEmpty()) {
+        logger.warning(
+            "The target class "
+                + targetClassName
+                + " symbol reference is not found in source class "
+                + sourceClassName);
+        // Cannot say the target unused
+        return false;
+      }
 
       ConstantPool sourceConstantPool = sourceJavaClass.getConstantPool();
       Constant[] constantPoolEntries = sourceConstantPool.getConstantPool();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -34,7 +34,6 @@ import java.nio.file.Path;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.logging.Logger;
 import javax.annotation.Nullable;
 import org.apache.bcel.Const;
 import org.apache.bcel.classfile.Attribute;
@@ -71,8 +70,6 @@ import org.apache.bcel.util.Repository;
  * in them, through the input class path for a static linkage check.
  */
 class ClassDumper {
-
-  private static final Logger logger = Logger.getLogger(ClassDumper.class.getName());
 
   private final ImmutableList<Path> inputClassPath;
   private final Repository classRepository;
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -19,7 +19,6 @@ package com.google.cloud.tools.opensource.classpath;
 import static com.google.common.base.Preconditions.checkArgument;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Verify;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
@@ -555,13 +554,9 @@ class ClassDumper {
       ImmutableSet<Integer> targetConstantPoolIndices =
           constantPoolIndexForClass(sourceJavaClass, targetClassName);
       if (targetConstantPoolIndices.isEmpty()) {
-        logger.warning(
-            "The target class "
-                + targetClassName
-                + " symbol reference is not found in source class "
-                + sourceClassName);
-        // Cannot say the target unused
-        return false;
+        // This reference is not found in sourceJavaClass any more. This means that there exist
+        // overlapping classes in the class path and that the first one does not use target class.
+        return true;
       }
 
       ConstantPool sourceConstantPool = sourceJavaClass.getConstantPool();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
@@ -93,6 +93,8 @@ public class StaticLinkageChecker {
     CommandLine commandLine = StaticLinkageCheckOption.readCommandLine(arguments);
     ImmutableList<Path> inputClasspath =
         StaticLinkageCheckOption.generateInputClasspath(commandLine);
+    logger.info("Input class path size: " + inputClasspath.size());
+
     // TODO(suztomo): take command-line option to choose entry point classes for reachability
     ImmutableSet<Path> entryPoints = ImmutableSet.of(inputClasspath.get(0));
 
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
@@ -279,14 +279,12 @@ public class DependencyGraphBuilder {
     queue.add(new LevelOrderQueueItem(firstNode, new Stack<>()));
 
     // Records failures rather than existing immediately.
-    List<ExceptionAndPath> resolutionFailures = Lists.newArrayList();
-    // Not to revisit same artifact
+    List<ExceptionAndPath> unsafeResolutionFailures = Lists.newArrayList();
+
     Set<Artifact> visitedArtifacts = Sets.newHashSet(firstNode.getArtifact());
-    int visitedNodeCount = 0;
 
     while (!queue.isEmpty()) {
       LevelOrderQueueItem item = queue.poll();
-      visitedNodeCount++;
       DependencyNode dependencyNode = item.dependencyNode;
       DependencyPath forPath = new DependencyPath();
       Stack<DependencyNode> parentNodes = item.parentNodes;
@@ -323,26 +321,27 @@ public class DependencyGraphBuilder {
               DependencyNode failedDependencyNode = artifactResult.getRequest().getDependencyNode();
               ExceptionAndPath failure =
                   ExceptionAndPath.create(parentNodes, failedDependencyNode, resolutionException);
-              if (isUnacceptableResolutionException(failure)) {
-                resolutionFailures.add(failure);
+              if (!isSafeResolutionException(failure)) {
+                unsafeResolutionFailures.add(failure);
               }
             }
           } catch (DependencyCollectionException collectionException) {
             DependencyNode failedDependencyNode = collectionException.getResult().getRoot();
             ExceptionAndPath failure =
                 ExceptionAndPath.create(parentNodes, failedDependencyNode, collectionException);
-            if (isUnacceptableResolutionException(failure)) {
-              resolutionFailures.add(failure);
+            if (!isSafeResolutionException(failure)) {
+              unsafeResolutionFailures.add(failure);
             }
           }
         }
       }
 
       for (DependencyNode child : dependencyNode.getChildren()) {
-        // A graph including provided scope may become big (more than 20,000).
-        // Optimizing by not visiting a node twice
+        // A graph including provided scope may become very big (more than 20,000 nodes).
         // https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/367
+        // Optimizing the traversal by not visiting a node twice
         if (includeProvidedScope && !visitedArtifacts.add(child.getArtifact())) {
+          // The optimization is not applicable for graphs used for upper-bound suggestions.
           continue;
         }
         @SuppressWarnings("unchecked")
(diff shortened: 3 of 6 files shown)


dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilderTest.java
```diff
@@ -70,7 +70,7 @@ public void testArtifactsToPaths_removingDuplicates() throws RepositoryException
     Path opencensusApiPath = opencensusApiPathFound.get();
     Truth.assertWithMessage("Opencensus API should have multiple dependency paths")
         .that(multimap.get(opencensusApiPath).size())
-        .isGreaterThan(1);
+        .isEqualTo(1);
```
suztomo: This change is because `DependencyGraphBuilder.levelOrder` does not visit same node twice.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilderTest.java
```diff
@@ -184,26 +184,4 @@ public void testArtifactToClasspath_notToGenerateRepositoryException()
     List<Path> paths = ClassPathBuilder.artifactsToClasspath(ImmutableList.of(jamonApiArtifact));
     Truth.assertThat(paths).isNotEmpty();
   }
-
-  @Test
-  public void testArtifactToClasspath_reportAggregatedRepositoryException()
```
suztomo: This test has been removed because now `jamonapi` can build a class path thanks to the change https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/365#discussion_r250581145

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -551,9 +550,11 @@ boolean isUnusedClassSymbolReference(ClassSymbolReference reference) {
 
       ImmutableSet<Integer> targetConstantPoolIndices =
           constantPoolIndexForClass(sourceJavaClass, targetClassName);
-      Verify.verify(
-          !targetConstantPoolIndices.isEmpty(),
-          "The target class symbol reference is not found in source class");
+      if (targetConstantPoolIndices.isEmpty()) {
+        // This reference is not found in sourceJavaClass any more. This means that there exist
```
elharo: Should this case be flagged as an error? Or perhaps a warning?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -318,46 +321,54 @@ private static void levelOrder(
...
           }
         }
       }
+
       for (DependencyNode child : dependencyNode.getChildren()) {
+        // A graph including provided scope may become very big (more than 20,000 nodes).
+        // https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/367
+        // Optimizing the traversal by not visiting a node twice
+        if (includeProvidedScope && !visitedArtifacts.add(child.getArtifact())) {
+          // The optimization is not applicable for graphs used for upper-bound suggestions.
```
elharo: Why? does a node here not include version?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -318,46 +321,54 @@ private static void levelOrder(
...
   }
 
   /**
-   * Returns true if {@code exceptionAndPath.getPath} does not contain {@code optional} dependency
-   * and the path does not contain {@code scope:provided} dependency.
+   * Returns true if {@code exceptionAndPath.getPath} contains an {@code optional} dependency or a
+   * {@code scope:provided} dependency.
    */
-  private static boolean isUnacceptableResolutionException(ExceptionAndPath exceptionAndPath) {
+  private static boolean isSafeResolutionException(ExceptionAndPath exceptionAndPath) {
```
elharo: I wonder if there's an accepted name for this concept?

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilderTest.java
```diff
@@ -70,7 +70,7 @@ public void testArtifactsToPaths_removingDuplicates() throws RepositoryException
     Path opencensusApiPath = opencensusApiPathFound.get();
     Truth.assertWithMessage("Opencensus API should have multiple dependency paths")
         .that(multimap.get(opencensusApiPath).size())
```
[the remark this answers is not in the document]
suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
elharo: I was thinking something we would surface to the client via an exception or return value.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
elharo: I still l don't follow this. Why would we ever want to traverse a path from the same artifact a second time?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
elharo: Perhaps just `isOptionalOrProvided`

library-best-practices/JLBP-10.md
```diff
@@ -30,7 +30,7 @@ accepting breaking changes for various reasons:
 
 As soon as you introduce breaking changes, they might decide to stick with the
 last working version indefinitely. As an example of a popular library that took
-a long time to upgrade, Hadoop introduced Guava 11.0.2 in May of 2012, and
+a long time to upgrade, Hadoop had introduced Guava 11.0.2 in May of 2012, and
```
elharo: don't need "had" here

suztomo: Sorry, this file is not to be in the PR.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: The tool doesn't have level of severity. For now, I add
- warning message and
- return `false` so that the reference is shown as an error.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: It's because a tree starting with each artifact should have complete upper-bound suggestion.

For example, when Artifact A and B show upper-bound suggestions for artifact X, following suggestions are expected:

```
(expected)
Artifact A: version 1
  upper-bound suggestion: artifact X: version 2
Artifact B: version 1
  upper-bound suggestion: artifact X: version 2
```

But the optimization skips artifact X in artifact B, resulting in following suggestions, because artifact C node in artifact B was skipped.

```
(unexpected, missing suggestion for Artifact B because of the optimization)
Artifact A: version 1
  upper-bound suggestion: artifact X: version 2
Artifact B: version 1
  upper-bound suggestion: (none, skipped!)
```

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Renamed.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
elharo: Sorry, I'm still not seeing this. Perhaps we should discuss offline with a whiteboard. If the artifact is the same as what we already saw, we should not follow it ever. If the artifact is different than what we saw, we should always follow it. How is anything missed?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -553,7 +556,14 @@ boolean isUnusedClassSymbolReference(ClassSymbolReference reference) {
       if (targetConstantPoolIndices.isEmpty()) {
         // This reference is not found in sourceJavaClass any more. This means that there exist
         // overlapping classes in the class path and that the first one does not use target class.
-        return true;
+        logger.warning(
+            "The source "
```
netdpb: Maybe use `String.format()`.

Also, how about this more specific message:

```java
String.format(
    "Some version of %1$s has a reference to %2$s, which would be a linkage error, but a different version of %1$s appears earlier in the class path and does not have that reference.",
    sourceClassName, targetClassName);
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -553,7 +556,14 @@ boolean isUnusedClassSymbolReference(ClassSymbolReference reference) {
...
+                + sourceClassName
+                + " does not have target "
+                + targetClassName
+                + ", because of overlapping source classes on the class path");
+        // By returning false ("the reference used"), the reference is shown as error
+        return false;
```
netdpb: I'm not sure this is right. Given this class path, this potential problem will never actually cause a runtime error, right?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
```diff
@@ -93,7 +93,14 @@ public static void main(String[] arguments)
     CommandLine commandLine = StaticLinkageCheckOption.readCommandLine(arguments);
     ImmutableList<Path> inputClasspath =
         StaticLinkageCheckOption.generateInputClasspath(commandLine);
-    logger.info("Input class path size: " + inputClasspath.size());
+
+    StringBuilder classPathOutputBuilder =
```
netdpb: Oh, sorry. I didn't mean to make you do anything nontrivial. I thought you could just do

`logger.info("Input class path: " + inputClassPath);`

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -356,19 +356,30 @@ private static void levelOrder(
...
-   * Returns true if {@code exceptionAndPath.getPath} contains an {@code optional} dependency or a
+   * Records the path and {@code repositoryException} to {@code unsafeResolutionFailures} if the the
+   * exception at {@code failedDependencyNode} is not safe.
+   */
+  private static void recordUnsafeFailuresIfNotSafe(
+      Stack<DependencyNode> parentNodes,
+      DependencyNode failedDependencyNode,
+      RepositoryException repositoryException,
+      List<ExceptionAndPath> unsafeResolutionFailures) {
+    ExceptionAndPath failure =
```
netdpb: Would this be simpler?

```java
if (isOptionalOrProvided(parentNodes)) {
  unsafeResolutionFailures.add(
      ExceptionAndPath.create(parentNodes, collectionException.getResult().getRoot(), collectionException));
}
```

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
```diff
@@ -341,6 +341,6 @@ public void testIsUnusedClassSymbolReference_classSymbolReferenceNotFound()
             .setSubclass(false)
             .setTargetClassName("dummy.NoSuchClass")
             .build();
-    Truth.assertThat(classDumper.isUnusedClassSymbolReference(referenceToUnusedClass)).isTrue();
+    Truth.assertThat(classDumper.isUnusedClassSymbolReference(referenceToUnusedClass)).isFalse();
```
netdpb: I thought you made it into an error now?

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilderTest.java
```diff
@@ -69,8 +69,9 @@ public void testArtifactsToPaths_removingDuplicates() throws RepositoryException
     Truth8.assertThat(opencensusApiPathFound).isPresent();
     Path opencensusApiPath = opencensusApiPathFound.get();
     Truth.assertWithMessage("Opencensus API should have multiple dependency paths")
```
netdpb: Is this message wrong now?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -272,11 +272,16 @@ private static void levelOrder(
       throws AggregatedRepositoryException {
 
     boolean resolveFullDependency = graphTraversalOption.resolveFullDependencies();
+    boolean includeProvidedScope =
```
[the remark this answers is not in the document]
suztomo: You're right; this variable is used by line 311 (within a try clause) and line 341 (outside the try clause).

Having said that, defining variable before while (at the beginning of a function) tells that the value does not change over the iteration.

Discussed with @elharo , the cause of the complexity in this function is that this function started handling two topology, a tree for upper-bound and a (more generic) graph for building a class path. I think I'll revise this while loop to use Guava's Graph.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -318,46 +321,54 @@ private static void levelOrder(
...
+              unsafeResolutionFailures.add(failure);
             }
           }
         }
       }
+
       for (DependencyNode child : dependencyNode.getChildren()) {
+        // A graph including provided scope may become very big (more than 20,000 nodes).
+        // https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/367
+        // Optimizing the traversal by not visiting a node twice
```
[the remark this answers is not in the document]
suztomo: The change does not skip other versions of the same artifact, because the equality of `AbstractArtifact`  includes `version`.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: `isOptionalOrProvided` also checks `failedDependencyNode`. The concatenation is done at `ExceptionAndPath create`.

If concatenation is done outside ExceptionAndPath, then that would be

```
if (isOptionalOrProvided(ImmutableList.<DependencyNode>builder()
                .addAll(parentDependencyNodes)
                .add(dependencyNode)
                .build())) {
  unsafeResolutionFailures.add(
      ExceptionAndPath.create(parentNodes, collectionException.getResult().getRoot(), collectionException));
}

```

So I think prev version is good.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilderTest.java
suztomo: Right. Fixed.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
suztomo: returning false for "isUnused?" meaning this is used. So the reference is reported as an error.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -315,49 +318,68 @@ private static void levelOrder(
               if (artifactResult.getExceptions().isEmpty()) {
                 continue;
               }
-              DependencyNode failedDependencyNode = artifactResult.getRequest().getDependencyNode();
-              ExceptionAndPath failure =
-                  ExceptionAndPath.create(parentNodes, failedDependencyNode, resolutionException);
-              if (isUnacceptableResolutionException(failure)) {
-                resolutionFailures.add(failure);
-              }
+              recordUnsafeFailuresIfNotSafe(
```
elharo: Are there unsafe failures that are safe?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/StaticLinkageChecker.java
suztomo: The input class path may have more than hundreds of elements. So I think multi-line look better.

PR #382, opened 2019-01-29
clarify definition of shading

elharo:
@garrettjonesgoogle

### DIFF
--- a/library-best-practices/JLBP-18.md
+++ b/library-best-practices/JLBP-18.md
@@ -1,7 +1,7 @@
 [JLBP-18] Only shade dependencies as a last resort
 --------------------------------------------------
 
-Shading is a process where a dependency is relocated to different Java package
+Shading is a process where a dependency is relocated to a different Java package
 and copied into the same jar as the code that relies on that dependency.
 The main purpose of shading is to avoid conflicts
 between the versions of dependencies used by a library and the versions used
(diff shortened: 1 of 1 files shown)


library-best-practices/JLBP-18.md
```diff
@@ -1,9 +1,9 @@
 [JLBP-18] Only shade dependencies as a last resort
 --------------------------------------------------
 
-Shading is a process where selected classes from a library and its
-dependencies are relocated to different Java packages and then copied into the
-final jar of the library. The main purpose of shading is to avoid conflicts
+Shading is a process where a dependency is relocated to different Java package
```
garrettjonesgoogle: "to a different"

library-best-practices/JLBP-18.md
```diff
@@ -1,9 +1,9 @@
...
+and copied into the same jar as the code that relies on that dependency.
```
garrettjonesgoogle: There's something that smells off about this - it isn't clear in this sentence which library does the shading (it's always the one that has the dependency on something else).

library-best-practices/JLBP-18.md
elharo: done

library-best-practices/JLBP-18.md
elharo: Neither does the shading. A separate tool does the shading. The dependency is shaded, and it is bundled into the jar with the code that depends on it.

elharo: True enough. However it's not necessarily another library that shades. It could be and often is a non-library application. This feels like another example where we're having trouble because we don't have a clear, well-understood term for the consumer of a dependency.

PR #387, opened 2019-01-30
Jlbp3 special case

### LINKED ISSUE #371: JLBP-3's special case: not bumping major version upon minimum Java version change
@garrettjonesgoogle What do you think about below. (Elliotte and I talked about this Friday. We would like to hear your opinion)

JLBP-3 says

> Special case: Merely bumping up the minimum required Java version (and not making any breaking surface changes) does not necessarily mean that a library should bump its major version, because new Java versions break very little surface from prior versions.
https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/5bed105c7c642bf7efd3b354d4a11bc63b0a5df4/library-best-practices/JLBP-3.md

I think, however, "changing minimum required Java version" matches "when you make incompatible API changes" best, among the three cases explained in Semantic Versioning:

> Given a version number MAJOR.MINOR.PATCH, increment the:
> 1. MAJOR version when you make incompatible API changes,
> 2. MINOR version when you add functionality in a backwards-compatible manner, and
> 3. PATCH version when you make backwards-compatible bug fixes.
https://semver.org/

For example, when a library changes minimum required Java version from Java 8 to Java 11, then other Java8 projects that use the library need to update their build configuration to Java 11; otherwise their Java runtime becomes _incompatible_ with the library.

Therefore, I propose that JLBP-3 have "Special case: When bumping up the minimum required Java version, the library should bump its major version, because users of the library need to update their build configuration to use compatible Java version." This is in line with [JLBP-9's second paragraph](https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/library-best-practices/JLBP-9.md) 

Note that [JEP 238: multi-release JAR](https://openjdk.java.net/jeps/238) allows libraries to leverage new Java 11 functionalities without bumping up minimum required Java version to Java 8 or earlier. Thus, multi-release JAR is out of the scope of this discussion.

suztomo:
Fixes #371

@garrettjonesgoogle Updated the sentence incorporating your input. PTAL.

- "does not necessarily mean that" -> "need not" -> "do not have to"
- bump -> increment (more precise specific word)
- It's human (maintainers) who increase major version. Using active voice. https://developers.google.com/style/anthropomorphism
- Focus on "end-of-life Java version" rather than "new Java version breaks very little surface"

### DIFF
--- a/library-best-practices/JLBP-3.md
+++ b/library-best-practices/JLBP-3.md
@@ -36,7 +36,7 @@ Definition of semantic versioning (SemVer): https://semver.org
     features) that is exposed through a library's public API
   - Adding a new class
   - Adding a new method
-- Special case: maintainers need not increase a library's major version when
-  a new release only drops the support of an end-of-life Java version that is
-  not widely used by the consumers of the library and the release does not make
-  any breaking surface change.
+- Special case: Maintainers need not increase a library's major version when
+  a new release does not make any breaking surface change and the release only
+  drops the support of an end-of-life Java version that is not widely used by
+  the consumers of the library.
--- a/library-best-practices/JLBP-3.md
+++ b/library-best-practices/JLBP-3.md
@@ -36,7 +36,7 @@ Definition of semantic versioning (SemVer): https://semver.org
     features) that is exposed through a library's public API
   - Adding a new class
   - Adding a new method
-- Special case: Maintainers need not increase a library's major version when
+- Special case: Maintainers need not increment a library's major version when
   a new release does not make any breaking surface change and the release only
   drops the support of an end-of-life Java version that is not widely used by
   the consumers of the library.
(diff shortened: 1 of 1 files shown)


library-best-practices/JLBP-3.md
```diff
@@ -36,7 +36,7 @@ Definition of semantic versioning (SemVer): https://semver.org
     features) that is exposed through a library's public API
   - Adding a new class
   - Adding a new method
-- Special case: Merely bumping up the minimum required Java version (and not
-  making any breaking surface changes) does not necessarily mean that a
-  library should bump its major version, because new Java versions break
-  very little surface from prior versions.
+- Special case: Maintainers need not increment a library's major version when
```
elharo: need not --> do not have to

This sentence is long. Can it broken up?

suztomo: Removed "a new release does not make any breaking surface change" in favor of "a release is only to drop..."

PR #412, opened 2019-02-05
Clarify relative to terms defined in Java API docs and JLS 

elharo:
@garrettjonesgoogle @suztomo 

1. errors are not exceptions
2. throws clause, not declaration,
3. signature is redundant

### DIFF
--- a/library-best-practices/glossary.md
+++ b/library-best-practices/glossary.md
@@ -19,23 +19,26 @@ Java Dependency Glossary
 ### Types of conflicts and compatibility
 
 <a name="linkage-error"></a>
-- **Linkage error**: A "class has some dependency on another class; however, the
-  latter class has incompatibly changed after the compilation of the former class."<sup>[1](#myfootnote1)</sup>
-  The reference can be through a class literal, a field access, or a method invocation.
-  Linkage errors detected at runtime manifest as a subclass of `LinkageError` such as
-  `NoSuchMethodError`, `NoClassDefFoundError`, `NoSuchFieldError`, or similar errors.
+- **Linkage error**: an abnormal condition of a classpath in which a
+  "class has some dependency on another class; however, the
+  latter class has incompatibly changed after the compilation of the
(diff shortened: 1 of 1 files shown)


library-best-practices/glossary.md
```diff
@@ -37,9 +38,9 @@ Java Dependency Glossary
 <a name="linkage-conflict"></a>
 <a name="static-linkage-conflict"></a>
 <a name="dynamic-linkage-conflict"></a>
-- **Linkage conflict**: a linkage error when the signature, return type,
-  modifiers, or throws declaration of a non-private method, field, or class
-  in a dependency has changed (or removed) in an incompatible way between
+- **Linkage conflict**: a linkage error when the name, return type,
+  modifiers, arguments, or throws clause of a non-private method, field, or class
```
suztomo: I found a change to `throws` clause does not break compatibility. Shall we remove this part?

> Changes to the throws clause of methods or constructors do not break compatibility with pre-existing binaries; these clauses are checked only at compile time.
https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.4.21

library-best-practices/glossary.md
```diff
@@ -19,50 +19,26 @@ Java Dependency Glossary
 ### Types of conflicts and compatibility
 
 <a name="linkage-error"></a>
-<a name="static-linkage-error"></a>
-<a name="dynamic-linkage-error"></a>
-- **Linkage error**: an error when a Java class references
-  another class and the reference cannot be satisfied with the available classes in the class path.
+- **Linkage error**: A "class has some dependency on another class; however, the
```
suztomo: I felt the double quotation typo.
image

Also would you define this as a noun, not a quoted sentence? For example "an error when ..."

suztomo: How about 

> An unsatisfactory dependency where "class has some dependency on another ....

?

elharo: fixed

PR #425, opened 2019-02-08
Dashboard to use ul tags for source classes

suztomo:
As part of #395 

- From pre tags to ul tags for list of source classes.
  Formatting has moved from `JarLinkageReport` to .ftl files.
- Removed duplicate count for inner classes
- Fixed freaky build failure ([CircleCI](https://circleci.com/gh/GoogleCloudPlatform/cloud-opensource-java/tree/source_class_ul)) by applying surefire-plugin config at root project.
- New file `macros.ftl` to write Freemarker macro. Shared between dashboard.ftl and component.ftl.

Now dashboard looks like this:

image

PR #429, opened 2019-02-11
Utility function toList(Nodes)

### LINKED ISSUE #426: Utility to create ImmutableList from xom's Nodes
Followup for David's idea:

>I know `Nodes` doesn't implement `Iterable<Node>`, but could we (in a separate PR) have some static method that creates an `ImmutableList<Node>` from a `Nodes`, so that we could `assertThat(nodes).comparingElementsUsing(NODE_VALUES).containsExacltly(something)` or do other useful things?
https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/425/files#r255252366

suztomo:
Fixes #426 . Followup for David's suggestion for a utility function to convert `Nodes` to `ImmutableList<Node>`:

>I know `Nodes` doesn't implement `Iterable<Node>`, but could we (in a separate PR) have some static method that creates an `ImmutableList<Node>` from a `Nodes`, so that we could `assertThat(nodes).comparingElementsUsing(NODE_VALUES).containsExacltly(something)` or do other useful things?
https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/425/files#r255252366

### DIFF
--- a/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
+++ b/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
@@ -16,6 +16,8 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URLDecoder;
@@ -48,11 +50,27 @@ import org.junit.Test;
 
 import com.google.cloud.tools.opensource.dependencies.Artifacts;
 import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
+import com.google.common.collect.ImmutableList;
 import com.google.common.io.MoreFiles;
 import com.google.common.io.RecursiveDeleteOption;
+import com.google.common.truth.Correspondence;
 import com.google.common.truth.Truth;
 
 public class DashboardTest {
+
+  private static final Correspondence<Node, String> NODE_VALUES =
+      new Correspondence<Node, String>() {
+        @Override
+        public boolean compare(Node node, String expected) {
+          String nodeValue = node.getValue().trim();
+          return nodeValue.equals(expected);
+        }
+
+        @Override
+        public String toString() {
+          return "has value equal to";
+        }
+      };
 
   private static Path outputDirectory;
   private Builder builder = new Builder();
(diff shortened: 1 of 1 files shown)


dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -116,7 +134,9 @@ public void testDashboard()
       Nodes href = document.query("//tr/td[@class='artifact-name']/a/@href");
       for (int i = 0; i < href.size(); i++) {
```
suztomo: I thought this loop could use `toList` function to eliminate the index variable `i`, but it turned out `artifacts.get(i)` uses the index variable.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -48,12 +50,28 @@
...
+      new Correspondence<Node, String>() {
+        @Override
+        public boolean compare(Node node, String expected) {
+          String nodeValue = node.getValue().trim();
+          return nodeValue.equals(expected);
+        }
+
+        @Override
+        public String toString() {
+          return "has value equal to";
```
elharo: This is strange. Why do we need this method? and returning a constant?

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -303,4 +314,11 @@ public int compare(String s1, String s2) {
 
   }
 
+  private static ImmutableList<Node> toList(Nodes nodes) {
```
elharo: This will no longer be necessary if I can ever convince Sonatype to let me upload XOM 1.3.0 to Maven Central (but I've failed twice so it may not be happening anytime soon)

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
suztomo: This is to help Correspondence's error message:

> suitable to fill the gap in a failure message of the form "<some actual element> is an element that ... <some expected element>"
https://google.github.io/truth/api/0.40/com/google/common/truth/Correspondence.html#toString--

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
suztomo: Looking forward.

Issue #395, opened 2019-01-31
better display large chains of errors on dashboard


elharo:
That is no need to show 2613 lines of output here; maybe combine cases where same class is missing.

```
appengine-api-1.0-sdk-1.9.64.jar (2613 errors):
  ClassSymbolReference{sourceClassName=com.google.appengine.api.appidentity.AppIdentityServicePb$SigningService$1, targetClassName=com.google.net.rpc3.client.RpcStubDescriptor, subclass=true}, reason: CLASS_NOT_FOUND, target class location not found
  ClassSymbolReference{sourceClassName=com.google.appengine.api.appidentity.AppIdentityServicePb$SigningService$BaseBlockingServerImpl, targetClassName=com.google.net.rpc3.RpcException, subclass=false}, reason: CLASS_NOT_FOUND, target class location not found
  ClassSymbolReference{sourceClassName=com.google.appengine.api.appidentity.AppIdentityServicePb$SigningService$BaseBlockingServerImpl, targetClassName=com.google.net.rpc3.impl.RpcUtil, subclass=false}, reason: CLASS_NOT_FOUND, target class location not found
  ClassSymbolReference{sourceClassName=com.google.appengine.api.appidentity.AppIdentityServicePb$SigningService$BaseServerImpl, targetClassName=com.google.net.rpc3.impl.RpcUtil, subclass=false}, reason: CLASS_NOT_FOUND, target class location not found
  ClassSymbolReference{sourceClassName=com.google.appengine.api.appidentity.AppIdentityServicePb$SigningService$BlockingServerInterface, targetClassName=com.google.net.rpc3.RpcException, subclass=false}, reason: CLASS_NOT_FOUND, target class location not found
  ClassSymbolReference{sourceClassName=com.google.appengine.api.appidentity.AppIdentityServicePb$SigningService$ClientInterface, targetClassName=com.google.net.rpc3.RpcException, subclass=false}, reason: CLASS_NOT_FOUND, target class location not found
 ```

suztomo:
Now they are grouped to 101 lines.

![image](https://user-images.githubusercontent.com/28604/52098004-2efcf880-259b-11e9-976d-edaba2b3af43.png)

The scrolling got much better. Do you want to shorten them further?

elharo:
possibly, but we can do this in stages

suztomo:
Started thinking about the enhancement.

# Problems

After grouping them by classes, now the width is very big.
[...]

PR #431, opened 2019-02-12
Show simplified message when dependency paths can be grouped

### LINKED ISSUE #430: Dashboard to find the same suffix in DependencyPaths
For #395 's second problem of having too many`DependencyPath`s, the dashboard can find the common longest suffix among `DependencyPath`s.

image

- All of them have "com.google.http-client:google-http-client". Versions are 1.19.0, 1.23.0, 1.20.0, 1.27.0 etc.

suztomo:
Fixes #430. (as part of #395 )

image

@elharo As discussed last week, here is a simplified message for the same cause to avoid listing 900 dependency paths to log4j.

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -32,7 +32,6 @@ import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -72,7 +71,6 @@ import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.ListMultimap;
-import com.google.common.collect.Maps;
 import com.google.common.collect.Multimaps;
 import com.google.common.collect.Sets;
 
@@ -411,14 +409,18 @@ public class DashboardMain {
         .count();
   }
 
-  private static final int SUMMARIZING_DEPENDENCY_PATH_SIZE_THRESHOLD = 5;
+  private static final int MINIMUM_NUMBER_DEPENDENCY_PATHS = 5;
 
   /**
    * Returns mapping from jar files to summaries of the root problem in their {@link
    * DependencyPath}s. The summary explains common patterns ({@code groupId:artifactId}) in the path
-   * elements. The returned map does not have a key for a jar file when the length of its {@link
-   * DependencyPath} list is smaller than {@link #SUMMARIZING_DEPENDENCY_PATH_SIZE_THRESHOLD} or a
-   * common pattern is not found among the elements in the paths.
+   * elements. The returned map does not have a key for a jar file when it has fewer than {@link
+   * #MINIMUM_NUMBER_DEPENDENCY_PATHS} dependency paths or a common pattern is not found among the
+   * elements in the paths.
+   *
+   * <p>Example summary: "Artifacts 'com.google.http-client:google-http-client &gt;
+   * commons-logging:commons-logging &gt; log4j:log4j' exist in all 994 dependency paths. Example
+   * path: com.google.cloud:google-cloud-core:1.59.0 ..."
    *
    * <p>Using this summary in the BOM dashboard avoids repetitive items in the {@link
    * DependencyPath} list that share the same root problem caused by widely-used libraries, for
@@ -432,35 +434,50 @@ public class DashboardMain {
     ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
 
     for (Path jar : jarToDependencyPaths.keySet()) {
-      Set<String> versionlessCoordinatesIntersection = null; // null for 1st iteration
       List<DependencyPath> dependencyPaths = jarToDependencyPaths.get(jar);
 
-      for (DependencyPath dependencyPath : dependencyPaths) {
-        // List of versionless coordinates ("groupId:artifactId")
-        ImmutableList<String> versionlessCoordinatesInPath =
-            dependencyPath.getPath().stream().map(Artifacts::makeKey).collect(toImmutableList());
-        if (versionlessCoordinatesIntersection == null) {
-          // For 1st iteration, initializes the intersection set with LinkedHashSet, which keeps
-          // the order of the common artifacts.
-          versionlessCoordinatesIntersection = Sets.newLinkedHashSet(versionlessCoordinatesInPath);
-        } else {
-          // intersection of elements in DependencyPaths
-          versionlessCoordinatesIntersection.retainAll(versionlessCoordinatesInPath);
-        }
-      }
+      Set<String> commonVersionlessArtifacts = commonVersionlessArtifacts(dependencyPaths);
 
       // When dependencyPaths is not empty, versionlessCoordinates is not null
-      if (dependencyPaths.size() > SUMMARIZING_DEPENDENCY_PATH_SIZE_THRESHOLD
-          && versionlessCoordinatesIntersection.size() > 1) { // The last paths elements are always same
-        StringBuilder messageBuilder = new StringBuilder();
-        messageBuilder.append("Artifacts '");
-        messageBuilder.append(Joiner.on(" > ").join(versionlessCoordinatesIntersection));
-        messageBuilder.append("' exist in all " + dependencyPaths.size() + " dependency paths. ");
-        messageBuilder.append("Example path: ");
-        messageBuilder.append(dependencyPaths.get(0));
-        builder.put(jar.toString(), messageBuilder.toString());
+      if (dependencyPaths.size() > MINIMUM_NUMBER_DEPENDENCY_PATHS
+          && commonVersionlessArtifacts.size() > 1) { // The last paths elements are always same
+        builder.put(
+            jar.toString(),
+            summaryMessage(
+                dependencyPaths.size(), commonVersionlessArtifacts, dependencyPaths.get(0)));
       }
     }
     return builder.build();
+  }
+
+  private static Set<String> commonVersionlessArtifacts(List<DependencyPath> dependencyPaths) {
+    Set<String> versionlessCoordinatesIntersection = null; // null for 1st iteration
+    for (DependencyPath dependencyPath : dependencyPaths) {
+      // List of versionless coordinates ("groupId:artifactId")
+      ImmutableList<String> versionlessCoordinatesInPath =
+          dependencyPath.getPath().stream().map(Artifacts::makeKey).collect(toImmutableList());
+      if (versionlessCoordinatesIntersection == null) {
+        // For 1st iteration, initializes the intersection set with LinkedHashSet, which keeps
+        // the order of the common artifacts.
+        versionlessCoordinatesIntersection = Sets.newLinkedHashSet(versionlessCoordinatesInPath);
+      } else {
+        // intersection of elements in DependencyPaths
+        versionlessCoordinatesIntersection.retainAll(versionlessCoordinatesInPath);
+      }
+    }
+    return versionlessCoordinatesIntersection;
+  }
+
+  private static String summaryMessage(
+      int dependencyPathCount,
+      Set<String> versionlessCoordinatesIntersection,
+      DependencyPath examplePath) {
+    StringBuilder messageBuilder = new StringBuilder();
+    messageBuilder.append("Artifacts '");
+    messageBuilder.append(Joiner.on(" > ").join(versionlessCoordinatesIntersection));
+    messageBuilder.append("' exist in all " + dependencyPathCount + " dependency paths. ");
+    messageBuilder.append("Example path: ");
+    messageBuilder.append(examplePath);
+    return messageBuilder.toString();
   }
 }
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 
 import java.io.File;
@@ -31,12 +32,12 @@ import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.TreeMap;
-import java.util.stream.Collectors;
 
 import freemarker.template.Configuration;
 import freemarker.template.DefaultObjectWrapper;
@@ -431,19 +432,20 @@ public class DashboardMain {
     ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
 
     for (Path jar : jarToDependencyPaths.keySet()) {
-      Set<String> versionlessCoordinates = null; // null for 1st iteration
+      // To keep order of the common artifact
+      LinkedHashSet<String> versionlessCoordinates = null; // null for 1st iteration
       boolean firstIteration = true;
       ImmutableList<DependencyPath> dependencyPaths = ImmutableList
           .copyOf(jarToDependencyPaths.get(jar));
       for (DependencyPath dependencyPath : dependencyPaths) {
         // Set of versionless coordinates ("groupId:artifactId")
-        Set<String> versionlessCoordinatesInPath =
-            dependencyPath.getPath().stream().map(Artifacts::makeKey).collect(Collectors.toSet());
+        ImmutableList<String> versionlessCoordinatesInPath =
+            dependencyPath.getPath().stream().map(Artifacts::makeKey).collect(toImmutableList());
         if (firstIteration) {
-          versionlessCoordinates = Sets.newHashSet(versionlessCoordinatesInPath);
+          versionlessCoordinates = Sets.newLinkedHashSet(versionlessCoordinatesInPath);
           firstIteration = false;
         } else {
-          // intersection of elements in among DependencyPaths
+          // intersection of elements in DependencyPaths
           versionlessCoordinates.retainAll(versionlessCoordinatesInPath);
         }
       }
@@ -453,7 +455,7 @@ public class DashboardMain {
         StringBuilder messageBuilder = new StringBuilder();
         messageBuilder.append("There are " + dependencyPaths.size() + " paths to the artifact. ");
         messageBuilder.append("All of them have the same artifacts: ");
-        messageBuilder.append(Joiner.on(", ").join(versionlessCoordinates));
+        messageBuilder.append(Joiner.on(" > ").join(versionlessCoordinates));
         messageBuilder.append(". Example path: ");
         messageBuilder.append(dependencyPaths.get(0));
         builder.put(jar.toString(), messageBuilder.toString());
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -29,6 +29,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -36,6 +37,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.stream.Collectors;
 
 import freemarker.template.Configuration;
 import freemarker.template.DefaultObjectWrapper;
@@ -61,6 +63,7 @@ import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
 import com.google.cloud.tools.opensource.dependencies.Update;
 import com.google.cloud.tools.opensource.dependencies.VersionComparator;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMultimap;
@@ -69,8 +72,11 @@ import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.ListMultimap;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
+import com.google.common.collect.Sets;
 
 public class DashboardMain {
   public static final String TEST_NAME_STATIC_LINKAGE_CHECK = "Static Linkage Errors";
@@ -303,6 +309,7 @@ public class DashboardMain {
       templateData.put("dependencyRootNode", Iterables.getFirst(dependencyTree.values(), null));
       templateData.put("jarLinkageReports", staticLinkageCheckReports);
       templateData.put("jarToDependencyPaths", jarToDependencyPathsForArtifact);
+      templateData.put("jarToSimplifiedPathMessage", Maps.newHashMap());
       templateData.put("totalLinkageErrorCount", totalLinkageErrorCount);
       report.process(templateData, out);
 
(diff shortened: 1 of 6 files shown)


dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +414,44 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
         .filter(row -> row.getResult(columnName) == null || row.getFailureCount(columnName) > 0)
         .count();
   }
+
+  /**
+   * Returns mapping from jar file to simplified description of its {@link DependencyPath}s. A
```
elharo: I think we're doing a little more than that. We're identifying the true cause of the problem.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +414,44 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+   * simplified description of {@link DependencyPath}s explains a common pattern ({@code
+   * groupId:artifactId}) in the path elements, helping to avoid listing more than 5 repetitive
+   * items in BOM dashboard.
+   */
+  private static ImmutableMap<String, String> simplifyDependencyPathMessage(
+      Multimap<Path, DependencyPath> jarToDependencyPaths) {
```
elharo: since we care about order (I think) we should probably declare this as ImmutableMultimap

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +414,44 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+    // Freemarker is not good at handling non-string key
+    // https://freemarker.apache.org/docs/app_faq.html#faq_nonstring_keys
+    ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+
+    for (Path jar : jarToDependencyPaths.keySet()) {
+      List<Set<String>> commonVersionLessCoordinates = Lists.newArrayList();
```
elharo: versionlessCoordinates or maybe just artifactIds or some such

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +414,44 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+      Collection<DependencyPath> dependencyPaths = jarToDependencyPaths.get(jar);
+      for (DependencyPath dependencyPath : dependencyPaths) {
+        // Set of "groupId:artifactId"
```
elharo: I wonder if Maven has a name for this?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +414,44 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+        Set<String> versionLessCoordinates =
+            dependencyPath.getPath().stream().map(Artifacts::makeKey).collect(Collectors.toSet());
+        commonVersionLessCoordinates.add(versionLessCoordinates);
+      }
+      Set<String> intersection = Sets.newHashSet(Iterables.getFirst(commonVersionLessCoordinates,
+          Sets.newHashSet()));
+      for (Set<String> versionLessCoordinates : commonVersionLessCoordinates) {
+        intersection.retainAll(versionLessCoordinates);
```
elharo: This looks wrong. Only elements in the last pass through the loop will be retained.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +414,44 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+      }
+
+      if (dependencyPaths.size() > 5 && intersection.size() > 1) {
+        StringBuilder messageBuilder = new StringBuilder();
```
elharo: I need to think about rewriting the message to emphasize the true problem rather than the first artifact it's found in

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -158,6 +158,18 @@ public void testDashboard()
       // grpc-testing-1.17.1, shown as first item in linkage errors, has these errors
       Assert.assertTrue(linkage.getValue().contains("3 linkage errors in 3 classes"));
 
+      ImmutableList<Node> dependencyPaths = toList(
+          document.query("//p[@class='static-linkage-check-dependency-paths']"));
+      Node log4jDependencyPathMessage = dependencyPaths.get(dependencyPaths.size() - 1);
+      // Not to show all 900 dependency paths to log4j in the dashboard
+      Truth.assertThat(log4jDependencyPathMessage.getValue()).contains(
+          "and other 993 dependency paths to the jar file. All of them have the same artifacts:");
```
elharo: I don't think we need "All of them have the same artifacts:"

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Simplified the logic; no first pass or last pass any more.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Updated message as below. Let me know if you come up with something better.

> There are 994 paths to the artifact. All of them have the same artifacts: log4j:log4j, commons-logging:commons-logging, com.google.http-client:google-http-client. Example path: com.google.cloud:google-cloud-core:1.59.0 / com.google.http-client:google-http-client:1.27.0 / com.google.android:android:1.5_r4 / commons-logging:commons-logging:1.1.1 / log4j:log4j:1.2.12

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
suztomo: Simplified it.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Enriched the javadoc.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Updated.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +412,53 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+    for (Path jar : jarToDependencyPaths.keySet()) {
+      Set<String> versionlessCoordinates = null; // null for 1st iteration
+      boolean firstIteration = true;
+      ImmutableList<DependencyPath> dependencyPaths = ImmutableList
+          .copyOf(jarToDependencyPaths.get(jar));
+      for (DependencyPath dependencyPath : dependencyPaths) {
+        // Set of versionless coordinates ("groupId:artifactId")
+        Set<String> versionlessCoordinatesInPath =
+            dependencyPath.getPath().stream().map(Artifacts::makeKey).collect(Collectors.toSet());
+        if (firstIteration) {
```
elharo: can you just create an empty set before starting the loop instead?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +412,53 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+          versionlessCoordinates = Sets.newHashSet(versionlessCoordinatesInPath);
+          firstIteration = false;
+        } else {
+          // intersection of elements in among DependencyPaths
```
elharo: in among --> in

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
elharo: You're losing the order here. I think we want something like 

com.google.http-client:google-http-client > commons-logging:commons-logging > log4j:log4j 

Also, don't we want versions here? Or maybe not. 

And what happened to android in this example? So I think something like

log4j:log4j:1.2.12 is referenced from com.google.http-client:google-http-client 
via the path
com.google.http-client:google-http-client:1.27.0 > com.google.android:android:1.5_r4 > commons-logging:commons-logging:1.1.1 > log4j:log4j:1.2.12

suztomo: Updated to use `LinkedHashSet` to keep order of common elements.

# Version
I omit version because the paths for this log4j case are more diverse than I thought when we discussed. There are multiple versions of `com.google.http-client:google-http-client`:

image

# Android

There is a path that does not have android (the screenshot above)

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
elharo: FYI, per javadoc: "Instead of using the Multimap interface directly, prefer the subinterfaces ListMultimap and SetMultimap. These take their names from the fact that the collections they return from get behave like (and, of course, implement) List and Set, respectively."

https://google.github.io/guava/releases/23.3-jre/api/docs/com/google/common/collect/Multimap.html

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +413,54 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+          versionlessCoordinates = Sets.newLinkedHashSet(versionlessCoordinatesInPath);
+          firstIteration = false;
+        } else {
+          // intersection of elements in DependencyPaths
+          versionlessCoordinates.retainAll(versionlessCoordinatesInPath);
+        }
+      }
+
+      // When dependencyPaths is not empty, versionlessCoordinates is not null
+      if (dependencyPaths.size() > 5 && versionlessCoordinates.size() > 1) {
```
elharo: why 5?

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -158,6 +158,18 @@ public void testDashboard()
       // grpc-testing-1.17.1, shown as first item in linkage errors, has these errors
       Assert.assertTrue(linkage.getValue().contains("3 linkage errors in 3 classes"));
 
+      ImmutableList<Node> dependencyPaths = toList(
+          document.query("//p[@class='static-linkage-check-dependency-paths']"));
+      Node log4jDependencyPathMessage = dependencyPaths.get(dependencyPaths.size() - 1);
+      // Not to show all 900 dependency paths to log4j in the dashboard
```
elharo: I don't understand this comment here. Perhaps recast as a failure message.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +413,54 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+   * the jar files. The summary explains common patterns ({@code groupId:artifactId}) in the path
+   * elements. The returned map does not have a key for a jar file when the length of its {@link
+   * DependencyPath} list is small or a common pattern is not found among the items.
+   *
+   * <p>Using this summary in the BOM dashboard avoids repetitive items in the {@link
+   * DependencyPath} list that share the same root problem caused by widely-used libraries, for
+   * example, {@code commons-logging:commons-logging}, {@code
+   * com.google.http-client:google-http-client} and {@code log4j:log4j}.
+   */
+  private static ImmutableMap<String, String> dependencyPathRootProblemMessage(
```
elharo: This method name doesn't seem to match the return type.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +413,54 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
         .filter(row -> row.getResult(columnName) == null || row.getFailureCount(columnName) > 0)
         .count();
   }
+
+  /**
+   * Returns mapping from jar files to a summary of the root problem in {@link DependencyPath}s of
```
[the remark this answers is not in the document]
suztomo: Updated to "Returns mapping from jar files to summaries of the root problem in their {@link  DependencyPath}s."

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -414,19 +413,22 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+   * common pattern is not found among the items.
    *
    * <p>Using this summary in the BOM dashboard avoids repetitive items in the {@link
    * DependencyPath} list that share the same root problem caused by widely-used libraries, for
    * example, {@code commons-logging:commons-logging}, {@code
    * com.google.http-client:google-http-client} and {@code log4j:log4j}.
    */
-  private static ImmutableMap<String, String> dependencyPathRootProblemMessage(
-      Multimap<Path, DependencyPath> jarToDependencyPaths) {
+  private static ImmutableMap<String, String> jarToDependencyPathSummary(
```
elharo: findRootCauses?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -365,6 +370,8 @@ static void generateDashboard(
       templateData.put("latestArtifacts", latestArtifacts);
       templateData.put("jarLinkageReports", classpathCheckReport.getJarLinkageReports());
       templateData.put("jarToDependencyPaths", jarToDependencyPaths);
+      templateData.put(
+          "jarToSimplifiedPathMessage", jarToDependencyPathSummary(jarToDependencyPaths));
```
elharo: parameter name no longer matches Java code

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +412,57 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+        }
+      }
+
+      // When dependencyPaths is not empty, versionlessCoordinates is not null
+      if (dependencyPaths.size() > SUMMARIZING_DEPENDENCY_PATH_SIZE_THRESHOLD
+          && versionlessCoordinates.size() > 1) { // The last elements are always same among paths
+        StringBuilder messageBuilder = new StringBuilder();
+        messageBuilder.append("There are " + dependencyPaths.size() + " paths to the artifact. ");
+        messageBuilder.append("All of them have the same artifacts: ");
+        messageBuilder.append(Joiner.on(" > ").join(versionlessCoordinates));
```
elharo: This is the information we care about. It should be first.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -158,6 +158,18 @@ public void testDashboard()
       // grpc-testing-1.17.1, shown as first item in linkage errors, has these errors
       Assert.assertTrue(linkage.getValue().contains("3 linkage errors in 3 classes"));
 
+      ImmutableList<Node> dependencyPaths = toList(
+          document.query("//p[@class='static-linkage-check-dependency-paths']"));
+      Node log4jDependencyPathMessage = dependencyPaths.get(dependencyPaths.size() - 1);
+      // log4j dependency paths, which would have 994 lines, should be shown as summarized
```
elharo: delete "shown as"

elharo: There are 994 paths to log4j. These should be summarized.

suztomo: Updated.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Updated. For example it now has 

> Artifacts 'com.google.cloud:google-cloud-core-http > com.google.http-client:google-http-client-appengine > com.google.appengine:appengine-api-1.0-sdk' exist in all 9 dependency paths. Example path: com.google.cloud:google-cloud-core-http:1.59.0 / com.google.http-client:google-http-client-appengine:1.27.0 / com.google.appengine:appengine-api-1.0-sdk:1.9.64

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Updated to `findRootCauses`

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Updated to `dependencyPathRootCauses`

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
elharo: I find I have to read and really think about this algorithm every time I read it. It's correct so far so far as I can see but not self-evident from the code. If there's nothing simpler, please add some comments explaining how it works.

elharo: Also, instead of using a boolean firstIteration variable could we instead check that versionlessCoordinates is null or better yet empty?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +410,57 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
         .filter(row -> row.getResult(columnName) == null || row.getFailureCount(columnName) > 0)
         .count();
   }
+
+  private static final int SUMMARIZING_DEPENDENCY_PATH_SIZE_THRESHOLD = 5;
+
+  /**
+   * Returns mapping from jar files to summaries of the root problem in their {@link
+   * DependencyPath}s. The summary explains common patterns ({@code groupId:artifactId}) in the path
+   * elements. The returned map does not have a key for a jar file when the length of its {@link
```
elharo: "the length of its {@link
   * DependencyPath} list is smaller than {@link #SUMMARIZING_DEPENDENCY_PATH_SIZE_THRESHOLD}" --> "it has fewer than SUMMARIZING_DEPENDENCY_PATH_SIZE_THRESHOLD dependency paths"

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +410,57 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
         .filter(row -> row.getResult(columnName) == null || row.getFailureCount(columnName) > 0)
         .count();
   }
+
+  private static final int SUMMARIZING_DEPENDENCY_PATH_SIZE_THRESHOLD = 5;
```
elharo: This constant name is still unclear. Perhaps MINIMUM_NUMBER_DEPENDENCY_PATHS?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +410,57 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+      ListMultimap<Path, DependencyPath> jarToDependencyPaths) {
+    // Freemarker is not good at handling non-string key. Path object in .ftl is automatically
+    // converted to String. https://freemarker.apache.org/docs/app_faq.html#faq_nonstring_keys
+    ImmutableMap.Builder<String, String> builder = ImmutableMap.builder();
+
+    for (Path jar : jarToDependencyPaths.keySet()) {
+      Set<String> versionlessCoordinatesIntersection = null; // null for 1st iteration
+      List<DependencyPath> dependencyPaths = jarToDependencyPaths.get(jar);
+
+      for (DependencyPath dependencyPath : dependencyPaths) {
```
elharo: I think this would be clearer if you moved this for loop and line 435 into a helper method that returns the common suffix.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +410,57 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+          versionlessCoordinatesIntersection = Sets.newLinkedHashSet(versionlessCoordinatesInPath);
+        } else {
+          // intersection of elements in DependencyPaths
+          versionlessCoordinatesIntersection.retainAll(versionlessCoordinatesInPath);
+        }
+      }
+
+      // When dependencyPaths is not empty, versionlessCoordinates is not null
+      if (dependencyPaths.size() > SUMMARIZING_DEPENDENCY_PATH_SIZE_THRESHOLD
+          && versionlessCoordinatesIntersection.size() > 1) { // The last paths elements are always same
```
elharo: I think the body of this if block could also be pulled into a helper method that returns a string.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Updated

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Updated

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Changed it to `summaryMessage`

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Great idea. Updated with `commonVersionlessArtifacts` function.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +408,76 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+        builder.put(
+            jar.toString(),
+            summaryMessage(
+                dependencyPaths.size(), commonVersionlessArtifacts, dependencyPaths.get(0)));
+      }
+    }
+    return builder.build();
+  }
+
+  private static Set<String> commonVersionlessArtifacts(List<DependencyPath> dependencyPaths) {
```
elharo: this should return LinkedHashSet or even a list since we care about the order of these elements.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +408,76 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+    Set<String> versionlessCoordinatesIntersection = null; // null for 1st iteration
```
elharo: I think you can do something like

ImmutableList<String> v =  dependencyPaths.get(0).getPath().stream().map(Artifacts::makeKey).collect(toImmutableList());
Set<String> versionlessCoordinatesIntersection = Sets.newLinkedHashSet(v);

Then you don't need the special casing for the first iteration of the for loop since you'll just be intersecting with itself. 

You could pull dependencyPath.getPath().stream().map(Artifacts::makeKey).collect(toImmutableList()) out to a helper function to make this clearer and avoid duplication.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -405,4 +408,76 @@ public static long countFailures(List<ArtifactResults> table, String columnName)
...
+        // intersection of elements in DependencyPaths
+        versionlessCoordinatesIntersection.retainAll(versionlessCoordinatesInPath);
+      }
+    }
+    return versionlessCoordinatesIntersection;
+  }
+
+  private static String summaryMessage(
+      int dependencyPathCount,
+      Set<String> versionlessCoordinatesIntersection,
```
elharo: versionlessCoordinatesIntersection --> coordinates

PR #446, opened 2019-02-15
Split testDashboard into multiple cases

### LINKED ISSUE #439: Split DashboardTest.testDashboard into different test cases
`DashboardTest.testDashboard` is too long and is obscure to what it is asserting. Split it into different test cases.

suztomo:
Fixes #439.

- `DashboardTest.testDashboard` is split into multiple test cases for the sections
- New function `assertDocument` abstracts repetitive steps of opening file, asserting its existence, and parsing html document via xom.
- Removed unnecessary `throws` clause

### DIFF
--- a/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
+++ b/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
@@ -28,7 +28,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
-import java.util.function.Consumer;
 
 import nu.xom.Builder;
 import nu.xom.Document;
@@ -88,10 +87,7 @@ public class DashboardTest {
       Assert.fail("Could not generate dashboard");
     }
 
-    Path dashboardHtml = outputDirectory.resolve("dashboard.html");
-    try (InputStream source = Files.newInputStream(dashboardHtml)) {
-      dashboard = builder.build(source);
-    }
+    dashboard = parseOutputFile("dashboard.html");
   }
 
   @AfterClass
@@ -110,7 +106,7 @@ public class DashboardTest {
     Assert.assertTrue(Files.isRegularFile(dashboardCss));
   }
 
-  private void assertDocument(String fileName, Consumer<Document> assertion)
+  private static Document parseOutputFile(String fileName)
       throws IOException, ParsingException {
     Path html = outputDirectory.resolve(fileName);
     Assert.assertTrue("Could not find a regular file for " + fileName,
@@ -119,7 +115,7 @@ public class DashboardTest {
 
     try (InputStream source = Files.newInputStream(html)) {
       Document document = builder.build(source);
-      assertion.accept(document);
+      return document;
     }
   }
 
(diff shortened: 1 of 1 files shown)


dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -81,8 +78,7 @@ private static String trimAndCollapseWhiteSpace(String value) {
   private Builder builder = new Builder();
 
   @BeforeClass
-  public static void setUp()
-      throws IOException, TemplateException, RepositoryException, ClassNotFoundException {
```
suztomo: These exceptions were all gray (unnecessary) in IntelliJ

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -157,64 +162,6 @@ public void testDashboard()
           Assert.fail(message);
         }
       }
-
-      // TODO(#439): these should all be separate tests for the different components
```
suztomo: The following queries and assertions are moved to individual test cases.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -102,16 +98,25 @@ public static void cleanUp() {
...
-  public void testCss()
-      throws IOException, TemplateException, ParsingException, ArtifactDescriptorException {
+  public void testCss() {
     Path dashboardCss = outputDirectory.resolve("dashboard.css");
     Assert.assertTrue(Files.exists(dashboardCss));
     Assert.assertTrue(Files.isRegularFile(dashboardCss));
   }
 
+  private void assertDocument(String fileName, Consumer<Document> assertion)
+      throws IOException, ParsingException {
```
suztomo: Helper function to abstract repetitive steps for most of test cases.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -227,112 +174,196 @@ private static void assertValidCellValue(Element cellElement) {
...
-      Truth.assertThat(trimAndCollapseWhiteSpace(reports.get(0).getValue()))
-          .isEqualTo("2 target classes causing linkage errors referenced from 2 source classes.");
-      Nodes causes = document.query("//p[@class='jar-linkage-report-cause']");
-      Truth.assertWithMessage("grpc-alts should show linkage errors for CommunicatorServer")
-          .that(toList(causes))
-          .comparingElementsUsing(NODE_VALUES)
-          .contains(
-              "com.sun.jdmk.comm.CommunicatorServer is not found,"
-                  + " referenced from 1 source class ▶"); // '▶' is in the toggle button
+  private static class SortWithoutVersion implements Comparator<String> {
```
suztomo: Moved this to closer to the usage.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -102,16 +98,25 @@ public static void cleanUp() {
...
+    try (InputStream source = Files.newInputStream(html)) {
+      Document document = builder.build(source);
+      assertion.accept(document);
+    }
+  }
+
   @Test
-  public void testDashboard()
-      throws IOException, TemplateException, ParsingException, ArtifactDescriptorException {
+  public void testDashboard() throws IOException, ParsingException, ArtifactDescriptorException {
```
suztomo: The reason why `testDashboard` is not using `assertDocument` is because the assertion part has function calls with checked exception and not suitable for a lambda expression for 2nd argument of `assertDocument`

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -227,112 +174,196 @@ private static void assertValidCellValue(Element cellElement) {
...
-    try (InputStream source = Files.newInputStream(zeroLinkageErrorHtml)) {
-      Document document = builder.build(source);
-      Nodes staticLinkageTotalMessage = document.query("//p[@id='static-linkage-errors-total']");
-      Truth.assertThat(staticLinkageTotalMessage.size()).isEqualTo(1);
-      Truth.assertThat(staticLinkageTotalMessage.get(0).getValue())
-          .contains("0 static linkage error(s)");
-    }
-  }
-
-  private static class SortWithoutVersion implements Comparator<String> {
```
suztomo: This is moved to closer to the usage.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -102,16 +98,25 @@ public static void cleanUp() {
...
     Assert.assertTrue(Files.isRegularFile(dashboardCss));
   }
 
+  private void assertDocument(String fileName, Consumer<Document> assertion)
+      throws IOException, ParsingException {
+    Path html = outputDirectory.resolve(fileName);
+    Assert.assertTrue(Files.isReadable(html));
+
+    try (InputStream source = Files.newInputStream(html)) {
+      Document document = builder.build(source);
```
elharo: are we always asserting the same document? I think so, and if so we should simply store a Document field in setUp

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
elharo: I don't think you need the lambda or this method  if you set up the document once in setUp

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
elharo: I still suspect Consumer and lambdas are overkill. This method can return the document and the specific tests can be run in the test methods.

suztomo: Updated. I was having wrong assumption that `Document` needs to be within a try clause.

PR #451, opened 2019-02-21
rename StaticLinkageError

elharo:
@suztomo Rename StaticLinkageError since:

1. This class is not a java.lang.Error, much less a java.lang.LinkageError
2. We're no longer calling this a "static" linkage error.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClasspathChecker.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClasspathChecker.java
@@ -160,11 +160,11 @@ public class ClasspathChecker {
   }
 
   private static <R extends SymbolReference>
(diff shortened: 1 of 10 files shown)


suztomo:
I agree with renaming from `StaticLinkageError`. However, `SymbolNotFound` is just one case of `LinkageError`.

Shall we pick another name?

- SymbolReferenceError
- SymbolLinkageError
(or something else?)

suztomo:
Not coming up with a good one yet. The difficulty comes from:
- we defined ["linkage error" in glossary](https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/library-best-practices/glossary.md#types-of-conflicts-and-compatibility)
- we're trying to name a Java class that represents it.
- but the Java class should have "error" in its name.

elharo:
SymbolNotResolvable? That could be because the symbol is not found or because it's inaccessible.

suztomo:
SymbolNotResolvable sounds good.

PR #463, opened 2019-02-24
Set version 0.1-snapshot

### LINKED ISSUE #456: Set version to 0.1-snapshot
Currently the version is set 1.0-snapshot. Set version to 0.1-snapshot

suztomo:
Fixes #456.

As we are not releasing a public library, it is not using a semantic versioning of the form "X.Y.Z".


suztomo:
Changes to boms/ are references to parent pom.xml version. Maybe they don't need parent pom. I'll try to remove them.

boms/pom.xml
```diff
@@ -1,17 +1,15 @@
...
   
-  <parent>
-    <groupId>com.google.cloud.tools.opensource</groupId>
-    <artifactId>cloud-tools-opensource-parent</artifactId>
-    <version>1.0-SNAPSHOT</version>
-  </parent>
-
+  <groupId>com.google.cloud.tools.opensource</groupId>
   <artifactId>cloud-tools-opensource-boms</artifactId>
+  <version>1.0-SNAPSHOT</version>
```
suztomo: Files under `boms/` has independent versions from `cloud-tools-opensource-parent`.

PR #464, opened 2019-02-24
Uncomment FreemarkerTest cleanup code

### LINKED ISSUE #448: Why doesn't FreemarkerTest delete? in cleanUp?
Either remove the method or uncomment it.

```
public class FreemarkerTest {

  private static Path outputDirectory;

  @AfterClass
  public static void cleanUp() throws IOException {
    // MoreFiles.deleteRecursively(outputDirectory, RecursiveDeleteOption.ALLOW_INSECURE);
  }
```

suztomo:
Fixes #448 . It was accidentally commented out when I was checking the output html.

- uncomment the directory cleanup
- removed ALLOW_INSECURE option when the tests do cleanup. As per [Javadoc](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/io/RecursiveDeleteOption.html#ALLOW_INSECURE), ALLOW_INSECURE is for symbolic links on (insecure) file systems. 
This test does not use symbolic links.

### DIFF
--- a/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/FreemarkerTest.java
+++ b/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/FreemarkerTest.java
@@ -35,6 +35,8 @@ import com.google.cloud.tools.opensource.dependencies.DependencyPath;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.ListMultimap;
+import com.google.common.io.MoreFiles;
+import com.google.common.io.RecursiveDeleteOption;
 import com.google.common.truth.Truth;
 
 import freemarker.template.Configuration;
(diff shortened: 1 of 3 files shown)


dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/FreemarkerTest.java
```diff
@@ -61,7 +63,7 @@ public static void setUp() throws IOException {
 
   @AfterClass
   public static void cleanUp() throws IOException {
-    // MoreFiles.deleteRecursively(outputDirectory, RecursiveDeleteOption.ALLOW_INSECURE);
+    MoreFiles.deleteRecursively(outputDirectory);
```
suztomo: As per [Javadoc](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/io/RecursiveDeleteOption.html#ALLOW_INSECURE), ALLOW_INSECURE is for symbolic links on (insecure) file systems. 
This test does not use symbolic links.

PR #470, opened 2019-02-25
To include transitive dependencies in mock test

### LINKED ISSUE #469: Investigate Guava 27 InternalFutureFailureAccess not found
Enforcer-rule test shows an error for guava:
Investigate Guava 27 InternalFutureFailureAccess not found

image

Source: `com.google.common.util.concurrent.AbstractFuture`
Target: `com.google.common.util.concurrent.internal.InternalFutureFailureAccess` and `com.google.common.util.concurrent.internal.InternalFutures`

The InternalFutures and InternalFutureAccess classes are now in `com.google.guava:failureaccess:1.0.1` [artifact](https://search.maven.org/artifact/com.google.guava/failureaccess/1.0.1/bundle).

https://github.com/google/guava/issues/3302

Now Guava has `<type>bundle</type>`. I think that's where LinkageChecker is missing dependency.

image

# mvn dependency:tree

This shows dependencies correctly, including `com.google.guava:failureaccess:jar:1.0.1:compile`

```
[INFO] com.google.cloud.tools.opensource:no-such-method-error-example:jar:1.0-SNAPSHOT
[INFO] \- com.google.guava:guava:jar:27.0.1-android:compile
[INFO]    +- com.google.guava:failureaccess:jar:1.0.1:compile
[INFO]    +- com.google.guava:listenablefuture:jar:9999.0-empty-to-avoid-conflict-with-guava:compile
[INFO]    +- com.google.code.findbugs:jsr305:jar:3.0.2:compile
[INFO]    +- org.checkerframework:checker-compat-qual:jar:2.5.2:compile
[INFO]    +- com.google.errorprone:error_prone_annotations:jar:2.2.0:compile
[INFO]    +- com.google.j2objc:j2objc-annotations:jar:1.1:compile
[INFO]    \- org.codehaus.mojo:animal-sniffer-annotations:jar:1.17:compile
```

# Answer: mock code in test was not enough

When it's running in JUnit, it returns 1-element dummy dependencies.

image

When it's running in real enforcer-plugin, it has 8 dependencies:

image

suztomo:
Fixes #469.

Previously `LinkageCheckerRuleTest.createResolvedDependency` was not returning transitive dependencies. This change is to populate transitive dependencies in the mock objects.

This bug was hidden in guava-26 as guava-26 does not require transitive dependencies. Since Guava 27, guava requires transitive dependency to `failureaccess` artifact.

### DIFF
--- a/enforcer-rules/src/test/java/com/google/cloud/tools/opensource/enforcer/LinkageCheckerRuleTest.java
+++ b/enforcer-rules/src/test/java/com/google/cloud/tools/opensource/enforcer/LinkageCheckerRuleTest.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.opensource.enforcer;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
@@ -43,7 +44,10 @@ import org.eclipse.aether.artifact.DefaultArtifact;
 import org.eclipse.aether.collection.CollectRequest;
 import org.eclipse.aether.graph.Dependency;
 import org.eclipse.aether.graph.DependencyNode;
+import org.eclipse.aether.resolution.ArtifactRequest;
+import org.eclipse.aether.resolution.ArtifactResult;
 import org.eclipse.aether.resolution.DependencyRequest;
+import org.eclipse.aether.resolution.DependencyResult;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


enforcer-rules/src/test/java/com/google/cloud/tools/opensource/enforcer/LinkageCheckerRuleTest.java
```diff
@@ -100,21 +105,25 @@ private Dependency createResolvedDependency(String coordinates) throws Repositor
 
     DependencyRequest dependencyRequest = new DependencyRequest();
     dependencyRequest.setRoot(dependencyNode);
-    repositorySystem.resolveDependencies(repositorySystemSession, dependencyRequest);
-
-    return dependencyNode.getDependency();
```
suztomo: This was returning only one, without transitive dependencies.

PR #471, opened 2019-02-25
LinkageCheckerRule not to use readBom which uses LocalRepository

### LINKED ISSUE #457: EnforcerRule test for non empty BOM
Add EnforcerRule test for non empty BOM. Both good one and bad one.

```
  @Test
  public void testExecute_shouldFailBadBom() throws RepositoryException {
    rule.setDependencySection(DependencySection.DEPENDENCY_MANAGEMENT);
    setupMockDependencyManagementSection(
        "com.google.api-client:google-api-client:1.27.0", "io.grpc:grpc-core:1.17.1");

    try {
      rule.execute(mockRuleHelper);
      Assert.fail("Enforcer rule should detect conflict between google-api-client and grpc-core");
    } catch (EnforcerRuleException ex) {
      // pass
    }
```

suztomo:
Fixes #457 . With the following change, now we have non-trivial test cases for LinkageCheckerRule checking a BOM.

Before this commit, LinkageCheckerRule was using `RepositoryUtility.readBom` to get the content of BOM project, which in turn asks a local repository. This required that the BOM is installed in local repository. This was unnecessary and was preventing from writing simple unit tests.

With this commit, LinkageCheckerRule reads BOM content through `MavenProject.getDependencyManagement()` without asking a repository. `RepositoryUtility.readBom` has special logic to skip artifacts, which now available as a public method `shouldSkipBomMember`.

### DIFF
--- a/enforcer-rules/src/test/java/com/google/cloud/tools/opensource/enforcer/LinkageCheckerRuleTest.java
+++ b/enforcer-rules/src/test/java/com/google/cloud/tools/opensource/enforcer/LinkageCheckerRuleTest.java
@@ -24,10 +24,12 @@ import static org.mockito.Mockito.when;
 import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
 import com.google.cloud.tools.opensource.enforcer.LinkageCheckerRule.DependencySection;
 import com.google.common.collect.ImmutableList;
+import java.util.Arrays;
 import org.apache.maven.artifact.handler.DefaultArtifactHandler;
 import org.apache.maven.enforcer.rule.api.EnforcerRuleException;
 import org.apache.maven.enforcer.rule.api.EnforcerRuleHelper;
 import org.apache.maven.execution.MavenSession;
+import org.apache.maven.model.DependencyManagement;
 import org.apache.maven.plugin.logging.Log;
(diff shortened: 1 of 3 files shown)


enforcer-rules/src/test/java/com/google/cloud/tools/opensource/enforcer/LinkageCheckerRuleTest.java
```diff
@@ -142,21 +145,76 @@ public void testExecute_shouldFailForBadProject() throws RepositoryException {
...
+            .map(DefaultArtifact::new)
+            .map(artifact -> new Dependency(artifact, "compile"))
+            .map(LinkageCheckerRuleTest::toDependency)
+            .collect(toImmutableList());
+    when(mockDependencyManagement.getDependencies()).thenReturn(bomMembers);
+
+    when(mockMavenSession.getRepositorySession()).thenReturn(repositorySystemSession);
+  }
+
+  private static org.apache.maven.model.Dependency toDependency(Dependency resolvedDependency) {
```
suztomo: I was looking for a method to convert `org.eclipse.aether.graph.Dependency` to `org.apache.maven.model.Dependency`, but could not find one.

PR #488, opened 2019-03-04
LinkageChecker to provide diagnosis on why an error happens

suztomo:
LinkageChecker to provide diagnosis on why an error happens.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilder.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilder.java
@@ -43,10 +43,6 @@ import org.eclipse.aether.artifact.Artifact;
  */
 public class ClassPathBuilder {
 
-  public static ClassPathBuilder create() {
-    return new ClassPathBuilder();
-  }
-
   private ClassPathBuilder() {}
 
   /**
@@ -57,7 +53,7 @@ public class ClassPathBuilder {
    * @return list of absolute paths to jar files
    * @throws RepositoryException when there is a problem retrieving jar files
    */
-  public ImmutableList<Path> artifactsToClasspath(List<Artifact> artifacts)
+  public static ImmutableList<Path> artifactsToClasspath(List<Artifact> artifacts)
       throws RepositoryException {
 
     // LinkedListMultimap keeps the key order as they were first added to the multimap
@@ -66,7 +62,7 @@ public class ClassPathBuilder {
     return ImmutableList.copyOf(pathToArtifact.keySet());
   }
 
-  ImmutableMap<Path, Artifact> getPathToArtifact(List<Artifact> artifacts)
+  static ImmutableMap<Path, Artifact> getPathToArtifact(List<Artifact> artifacts)
       throws RepositoryException {
     //    this.artifacts = artifacts;
 
@@ -101,7 +97,7 @@ public class ClassPathBuilder {
    * @return an ordered map of absolute paths of jar files to one or more Maven dependency paths
    * @throws RepositoryException when there is a problem retrieving jar files
    */
-  public LinkedListMultimap<Path, DependencyPath> artifactsToDependencyPaths(
+  public static LinkedListMultimap<Path, DependencyPath> artifactsToDependencyPaths(
       List<Artifact> artifacts) throws RepositoryException {
 
     LinkedListMultimap<Path, DependencyPath> multimap = LinkedListMultimap.create();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilder.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilder.java
@@ -75,8 +75,6 @@ public class ClassPathBuilder {
       DependencyPath firstValue = Iterables.getFirst(multimap.get(key), null);
       pathToArtifact.put(key, firstValue.getLeaf());
     }
-    multimap.forEach((key, value) -> {});
-
     return pathToArtifact.build();
   }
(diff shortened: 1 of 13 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilder.java
```diff
@@ -41,6 +43,12 @@
  */
 public class ClassPathBuilder {
 
+  public static ClassPathBuilder create() {
```
suztomo: (question to myself) Do I need to change this?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilder.java
```diff
@@ -75,8 +75,6 @@ private ClassPathBuilder() {}
       DependencyPath firstValue = Iterables.getFirst(multimap.get(key), null);
       pathToArtifact.put(key, firstValue.getLeaf());
     }
-    multimap.forEach((key, value) -> {});
```
elharo: Why did this go away?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -47,9 +49,21 @@
...
   private final ClassReferenceGraph classReferenceGraph;
+  private final ImmutableMap<Path, Artifact> pathToArtifact;
 
   public static LinkageChecker create(List<Path> jarPaths, Iterable<Path> entryPoints)
       throws IOException {
+    return create(jarPaths, entryPoints, null);
+  }
+
+  public static LinkageChecker create(
+      ImmutableMap<Path, Artifact> pathToArtifact, Iterable<Path> entryPoints) throws IOException {
```
elharo: I notice we're using a *lot* of maps for these sorts of things. This suggests we might be missing an abstraction somewhere.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -47,9 +49,21 @@
...
+    return create(ImmutableList.copyOf(pathToArtifact.keySet()), entryPoints, pathToArtifact);
+  }
+
+  public static LinkageChecker create(
+      List<Path> jarPaths, Iterable<Path> entryPoints, ImmutableMap<Path, Artifact> pathToArtifact)
```
elharo: e.g. instead of passing lists of paths and maps of paths to artifacts perhaps we need a type that includes both a path and an artifact. Not to fix in this PR, but let's think about that.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerArguments.java
```diff
@@ -62,7 +60,7 @@
   private final CommandLine commandLine;
   private ImmutableList<Path> cachedInputClasspath;
   private ImmutableList<Artifact> cachedArtifacts;
-  private ImmutableMap<Path, Artifact> pathToArtifact;
+  private ImmutableMap<Path, Artifact> cachedPathToArtifact;
```
elharo: Here's another one. We have two lists and a map to track what is conceptually one collection.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorDiagnosis.java
```diff
@@ -91,4 +106,76 @@ public final String toString() {
...
+                .build();
+        return diagnosis;
+      }
+    }
+
+    LinkageErrorDiagnosis diagnosis =
+        LinkageErrorDiagnosis.builder()
+            .setLinkageErrorCause(cause)
+            .setSourceArtifact(sourceArtifact)
+            .setArtifactWithResolvableSymbol(null)
```
elharo: why do you need to set these to null? Would that be the default?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorDiagnosis.java
```diff
@@ -91,4 +106,76 @@ public final String toString() {
...
+
+    for (LinkageErrorCause cause : causeToSourceClasses.keySet()) {
+      switch (cause.getReason()) {
+        case CLASS_NOT_FOUND:
+          LinkageErrorDiagnosis diagnosis =
+              diagnoseMissingClass(cause, sourceArtifact, selectedArtifacts, pathToArtifactForJar);
+          causeToDiagnosis.put(cause, diagnosis);
+          break;
+        default:
+          // TODO: Implement other cases (SYMBOL_NOT_FOUND, INACCESSIBLE_CLASS, etc.)
```
elharo: probably worth filling in the switch statement now even if the methods are no-ops for the time being.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
suztomo: Right. I was not sure about `pathToArtifact` either. Artifact has `getFile()` and File has `toPath()`. Let me think about better data structure.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilder.java
suztomo: This statement does not anything. (I think I forgot to remove while inspecting the content)

PR #496, opened 2019-03-07
Not to load shadowed class file in a class path

### LINKED ISSUE #495: LinkageChecker unable to check  io.grpc:grpc-alts:jar:1.18.0 com.google.cloud:google-cloud-nio:jar:0.81.0-alpha
There seems a bug:

argument
```
-a io.grpc:grpc-alts:jar:1.18.0,com.google.cloud:google-cloud-nio:jar:0.81.0-alpha
```

```
Exception in thread "main" com.google.common.base.VerifyException: The target class symbol reference org.apache.http.client.config.RequestConfig$Builder is not found in source class com.google.api.client.http.apache.ApacheHttpRequest
	at com.google.common.base.Verify.verify(Verify.java:124)
	at com.google.cloud.tools.opensource.classpath.ClassDumper.isUnusedClassSymbolReference(ClassDumper.java:585)
	at com.google.cloud.tools.opensource.classpath.LinkageChecker.checkLinkageErrorMissingClassAt(LinkageChecker.java:340)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
	at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
	at com.google.cloud.tools.opensource.classpath.LinkageChecker.errorsFromSymbolReferences(LinkageChecker.java:141)
	at com.google.cloud.tools.opensource.classpath.LinkageChecker.generateLinkageReport(LinkageChecker.java:110)
	at com.google.cloud.tools.opensource.classpath.LinkageChecker.lambda$findLinkageErrors$0(LinkageChecker.java:86)
	at com.google.common.collect.RegularImmutableMap.forEach(RegularImmutableMap.java:186)
	at com.google.cloud.tools.opensource.classpath.LinkageChecker.findLinkageErrors(LinkageChecker.java:84)
	at com.google.cloud.tools.opensource.classpath.LinkageCheckerMain.main(LinkageCheckerMain.java:51)

```

The class com.google.api.client.http.apache.ApacheHttpRequest exists in multiple jar files in the class path:

```
/usr/local/google/home/suztomo/.m2/repository/com/google/http-client/google-http-client/1.19.0/google-http-client-1.19.0.jar contains com.google.api.client.http.apache.ApacheHttpRequest
/usr/local/google/home/suztomo/.m2/repository/com/google/http-client/google-http-client-apache/2.0.0/google-http-client-apache-2.0.0.jar contains com.google.api.client.http.apache.ApacheHttpRequest
```

suztomo:
Fixes #495 , a bug where LinkageChecker was not able to handle duplicated class files in a class path correctly. [A test case](https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/496/files#diff-40dc800f7109446f3727fe9ff7f3a937R822) is also added.

# Detail of the issue

In this case of #495, `com.google.api.client.http.apache.ApacheHttpRequest` in `google-http-client-apache-2.0.0.jar` was checked for its reference. LinkageChecker failed because the class is shadowed by the one in `google-http-client-1.19.0.jar`. Because the former class file is shadowed, it's never used in the class path. No need to check its references.

This fix is not to load a class file shadowed by another class file. This is achieved by using one shared class path (ClassDumper's instance variable `classRepository`) to read jar files for `JavaClass`es.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -588,7 +588,7 @@ class ClassDumper {
           constantPoolIndexForClass(sourceJavaClass, targetClassName);
       Verify.verify(
           !targetConstantPoolIndices.isEmpty(),
-          "When checking a class reference from %s to %d, the reference to the target class is no"
+          "When checking a class reference from %s to %s, the reference to the target class is no"
               + " longer found in the source class's constant pool.", // This should not happen
           sourceJavaClass.getClassName(),
           targetClassName);
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -388,22 +388,17 @@ class ClassDumper {
     return ImmutableMap.copyOf(classToJar);
   }
 
-  private static ImmutableSet<ClassInfo> listClassInfo(URL jarUrl)
-      throws IOException {
+  static ImmutableSet<String> listClassNamesInJar(Path jar) throws IOException {
+    URL jarUrl = jar.toUri().toURL();
     // Setting parent as null because we don't want other classes than this jar file
-    URLClassLoader classLoaderFromJar = new URLClassLoader(new URL[]{jarUrl}, null);
+    URLClassLoader classLoaderFromJar = new URLClassLoader(new URL[] {jarUrl}, null);
 
     // Leveraging Google Guava reflection as BCEL doesn't list classes in a jar file
     com.google.common.reflect.ClassPath classPath =
         com.google.common.reflect.ClassPath.from(classLoaderFromJar);
 
-    return classPath.getAllClasses();
-  }
-
-  static ImmutableSet<String> listClassNamesInJar(Path jar) throws IOException {
-    URL jarUrl = jar.toUri().toURL();
-    return listClassInfo(jarUrl).stream()
-        .map(classInfo -> classInfo.getName())
+    return classPath.getAllClasses().stream()
+        .map(ClassInfo::getName)
         .collect(toImmutableSet());
   }
 
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -427,10 +427,9 @@ class ClassDumper {
    * #classRepository}. The class path to instantiate them is from {@link #classRepository} of this
    * {@link ClassDumper} instance.
    *
-   * <p>The set of class names between the return values of {@link #listClassesInJar(Path)} and
-   * {@link #listClassesInJarFromClassPath(Path)} are the same. However, the implementation of the
-   * returned instances between them may be different if a class appears multiple times in jar files
-   * in the class path.
+   * <p>The set of class names between the return values of {@link #listClassesInJar(Path)} and this
+   * method are the same. However, the implementation of the returned instances between them may be
+   * different if a class appears multiple times in jar files in the class path.
    */
   private ImmutableSet<JavaClass> listClassesInJarFromClassPath(Path jar) throws IOException {
     return listClassesInJar(jar, classRepository);
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -604,10 +604,9 @@ class ClassDumper {
           constantPoolIndexForClass(sourceJavaClass, targetClassName);
       Verify.verify(
           !targetConstantPoolIndices.isEmpty(),
-          "The target class "
-              + targetClassName
-              + " is not found in source class "
-              + sourceJavaClass.getClassName());
+          "The target class %s is not found in source class %s",
+          targetClassName,
+          sourceJavaClass.getClassName());
 
       ConstantPool sourceConstantPool = sourceJavaClass.getConstantPool();
       Constant[] constantPoolEntries = sourceConstantPool.getConstantPool();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -17,6 +17,7 @@
 package com.google.cloud.tools.opensource.classpath;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Verify;
@@ -86,10 +87,15 @@ class ClassDumper {
   static ClassDumper create(List<Path> jarPaths) throws IOException {
     ClassLoader systemClassLoader = ClassLoader.getSystemClassLoader();
     ClassLoader extensionClassLoader = systemClassLoader.getParent();
+    jarPaths.forEach(
+        jar -> {
+          checkArgument(
+              Files.isRegularFile(jar), "The jar file " + jar + " is not a regular file.");
+          checkArgument(Files.isReadable(jar), "The jar file " + jar + " is not readable.");
+        });
 
     return new ClassDumper(
         jarPaths,
-        createClassRepository(jarPaths),
         extensionClassLoader,
         mapJarToClasses(jarPaths),
         mapClassToJar(jarPaths));
@@ -97,12 +103,11 @@ class ClassDumper {
 
   private ClassDumper(
       List<Path> inputClassPath,
-      Repository classRepository,
       ClassLoader extensionClassLoader,
       SetMultimap<Path, String> jarToClasses,
       ImmutableMap<String, Path> classToFirstJar) {
     this.inputClassPath = ImmutableList.copyOf(inputClassPath);
-    this.classRepository = classRepository;
+    this.classRepository = createClassRepository(inputClassPath);
     this.extensionClassLoader = extensionClassLoader;
     this.jarFileToClasses = ImmutableSetMultimap.copyOf(jarToClasses);
     this.classToFirstJarFile = classToFirstJar;
@@ -156,7 +161,7 @@ class ClassDumper {
     checkArgument(Files.isReadable(jar), "The input jar file path is not readable");
 
     SymbolReferenceSet.Builder symbolTableBuilder = SymbolReferenceSet.builder();
-    for (JavaClass javaClass : listClassesInJarFromClassPath(jar)) {
+    for (JavaClass javaClass : listClassesInJar(jar)) {
       if (!isCompatibleClassFileVersion(javaClass)) {
         continue;
       }
@@ -352,15 +357,14 @@ class ClassDumper {
   /**
    * Returns mapping from jar files to the names of the classes they define.
    *
-   * @param jarPaths absolute paths to jar files
+   * @param jars absolute paths to jar files
    */
   @VisibleForTesting
-  static ImmutableSetMultimap<Path, String> mapJarToClasses(
-      List<Path> jarPaths) throws IOException {
+  static ImmutableSetMultimap<Path, String> mapJarToClasses(List<Path> jars) throws IOException {
     ImmutableSetMultimap.Builder<Path, String> pathToClasses = ImmutableSetMultimap.builder();
-    for (Path jarPath : jarPaths) {
-      for (JavaClass javaClass : listClassesInJar(jarPath)) {
-        pathToClasses.put(jarPath, javaClass.getClassName());
+    for (Path jar : jars) {
+      for (String className : listClassNamesInJar(jar)) {
+        pathToClasses.put(jar, className);
       }
     }
     return pathToClasses.build();
@@ -376,9 +380,9 @@ class ClassDumper {
   static ImmutableMap<String, Path> mapClassToJar(List<Path> jars) throws IOException {
     Map<String, Path> classToJar = new HashMap<>();
     for (Path jar : jars) {
-      for (JavaClass javaClass : listClassesInJar(jar)) {
+      for (String className : listClassNamesInJar(jar)) {
         // The first entry wins in the same manner as JVM's class loading.
-        classToJar.putIfAbsent(javaClass.getClassName(), jar);
+        classToJar.putIfAbsent(className, jar);
       }
     }
     return ImmutableMap.copyOf(classToJar);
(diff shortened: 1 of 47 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -151,13 +151,12 @@ boolean isSystemClass(String className) {
...
    */
-  static SymbolReferenceSet scanSymbolReferencesInJar(Path jarFilePath)
-      throws IOException {
+  SymbolReferenceSet scanSymbolReferencesInJar(Path jarFilePath) throws IOException {
     checkArgument(jarFilePath.isAbsolute(), "The input jar file path is not an absolute path");
     checkArgument(Files.isReadable(jarFilePath), "The input jar file path is not readable");
 
     SymbolReferenceSet.Builder symbolTableBuilder = SymbolReferenceSet.builder();
-    for (JavaClass javaClass : listClassesInJar(jarFilePath)) {
+    for (JavaClass javaClass : listClassesInJarFromClassPath(jarFilePath)) {
```
suztomo: Because `listClassesInJarFromClassPath` uses instance variable `classRepository`, these methods are now non-static method.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -151,13 +151,12 @@ boolean isSystemClass(String className) {
    *
    * @param jarFilePath absolute path to a jar file
    */
-  static SymbolReferenceSet scanSymbolReferencesInJar(Path jarFilePath)
-      throws IOException {
+  SymbolReferenceSet scanSymbolReferencesInJar(Path jarFilePath) throws IOException {
```
elharo: argument could simply be jar or maybe jarPath. jarFilePath is redundant.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -415,6 +415,19 @@ Path findClassLocation(String className) {
     return javaClasses.build();
   }
 
+  /** Returns a set of {@link JavaClass}es in {@code jarPath}. */
+  static ImmutableSet<JavaClass> listClassesInJar(Path jarPath) throws IOException {
+    return listClassesInJar(jarPath, createClassRepository(ImmutableList.of(jarPath)));
+  }
+
+  /**
+   * Returns a set of {@link JavaClass}es in {@code jarPath} via the class path in {@link
```
elharo: From this comment I don't understand the difference between this method and the previous one. Can you elaborate and clarify?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -584,7 +597,10 @@ boolean isUnusedClassSymbolReference(ClassSymbolReference reference) {
           constantPoolIndexForClass(sourceJavaClass, targetClassName);
       Verify.verify(
           !targetConstantPoolIndices.isEmpty(),
-          "The target class symbol reference is not found in source class");
+          "The target class symbol reference "
```
elharo: delete "symbol reference" here. It's understood.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Added comparison.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Updated

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
netdpb: I agree. Should at least the one that takes a `Repository` become an instance method on `Repository`? Then you don't really need the next one at all.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -584,7 +604,10 @@ boolean isUnusedClassSymbolReference(ClassSymbolReference reference) {
           constantPoolIndexForClass(sourceJavaClass, targetClassName);
       Verify.verify(
           !targetConstantPoolIndices.isEmpty(),
-          "The target class symbol reference is not found in source class");
+          "The target class "
+              + targetClassName
```
netdpb: `Verify.verify` is varargs and uses a simple format.

```java
verify(
    !targetConstantPoolIndices.isEmpty(),
    "The target class %s is not found in source class %s",
    targetClassName,
    sourceJavaClass.getClassName());
```

The above is easier to read and can be more performant (which doesn't matter in this case) because string formatting doesn't happen unless verification fails.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
```diff
@@ -345,7 +347,9 @@ public void testIsUnusedClassSymbolReference_classSymbolReferenceNotFound()
     } catch (VerifyException ex) {
       // pass
       Truth.assertThat(ex.getMessage())
```
netdpb: `Truth.assertThat(ex).hasMessageThat().isEqualTo(...);`

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
elharo: I disagree that format strings are easier to read than simple string concatenation. I like my words to come in the order I read them. Missing and out of order format string arguments are a common source of bugs. 

And to performance, I say [Knuth](http://wiki.c2.com/?PrematureOptimization)!

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: I agree that string concatenation is just fine compared to adding `String.format()`.

For this case, I would use the varargs is for `verify()`; I like calling functions the way the implementation intends to.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Right. That design works well.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: The private one that takes a Repository is to have the common logic between the other two. Because one of them is `static` (no need to be as an instance method), the private one is static.

suztomo: As discussed, the two static methods were not needed, as the caller only needed class names. Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -371,45 +374,47 @@ Path findClassLocation(String className) {
...
+        .collect(toImmutableSet());
   }
 
-  static ImmutableSet<JavaClass> listClassesInJar(Path jarPath) throws IOException {
-    Repository repository = createClassRepository(ImmutableList.of(jarPath));
+  /**
+   * Returns a set of {@link JavaClass}es which has entries in the {@code jar} through {@link
+   * #classRepository}.
+   */
+  private ImmutableSet<JavaClass> listClassesInJar(Path jar) throws IOException {
```
suztomo: Now other 2 static methods are removed.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
```diff
@@ -100,11 +100,11 @@ public void testListInnerClasses() throws IOException {
   }
 
   @Test
-  public void testCreationInvalidInput() {
+  public void testCreationInvalidInput() throws IOException {
     try {
       ClassDumper.create(ImmutableList.of(Paths.get("")));
       Assert.fail("Empty path should generate IOException");
-    } catch (IOException ex) {
+    } catch (IllegalArgumentException ex) {
```
suztomo: Previously it was throwing IOException when ClassRepository was trying to read file with empty string. Now it has checkArgument validation against such case.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -371,45 +373,47 @@ Path findClassLocation(String className) {
...
-    return classPath.getAllClasses();
+    return classPath.getAllClasses().stream()
+        .map(ClassInfo::getName)
+        .collect(toImmutableSet());
   }
 
-  static ImmutableSet<JavaClass> listClassesInJar(Path jarPath) throws IOException {
-    Repository repository = createClassRepository(ImmutableList.of(jarPath));
+  /**
+   * Returns a set of {@link JavaClass}es which has entries in the {@code jar} through {@link
```
elharo: has --> have

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -584,7 +588,9 @@ boolean isUnusedClassSymbolReference(ClassSymbolReference reference) {
           constantPoolIndexForClass(sourceJavaClass, targetClassName);
       Verify.verify(
           !targetConstantPoolIndices.isEmpty(),
-          "The target class symbol reference is not found in source class");
+          "The target class %s is not found in source class %s",
```
elharo: This message doesn't sound right unless target class is an inner class.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Fixed

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -584,7 +588,9 @@ boolean isUnusedClassSymbolReference(ClassSymbolReference reference) {
           constantPoolIndexForClass(sourceJavaClass, targetClassName);
       Verify.verify(
           !targetConstantPoolIndices.isEmpty(),
-          "The target class symbol reference is not found in source class");
+          "The reference to the target class %s is not found in the source class %s",
```
elharo: Is this accurate? The source class is missing a reference? I'd expect something like

Source class A references target class B, but B could not be found. 

This is more like 

Target class B could be found but source class A does not have a reference to target class B.

suztomo: Clarified the message. This `verify` is asserting "This should not happen".

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerTest.java
```diff
@@ -817,4 +817,23 @@ public void testFindLinkageErrors_doesNotCatchNoClassDefFoundError()
...
+    List<Path> paths =
+        ClassPathBuilder.artifactsToClasspath(
+            ImmutableList.of(
+                new DefaultArtifact("io.grpc:grpc-alts:jar:1.18.0"),
+                new DefaultArtifact("com.google.cloud:google-cloud-nio:jar:0.81.0-alpha")));
+
+    LinkageChecker linkageChecker = LinkageChecker.create(paths, paths);
+
+    // This should not raise an exception
+    linkageChecker.findLinkageErrors();
```
elharo: What's the return value here? Should this report an error, though not throw an exception?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -86,23 +87,26 @@ private static Repository createClassRepository(List<Path> paths) {
   static ClassDumper create(List<Path> jarPaths) throws IOException {
     ClassLoader systemClassLoader = ClassLoader.getSystemClassLoader();
     ClassLoader extensionClassLoader = systemClassLoader.getParent();
+    jarPaths.forEach(
+        jar -> {
```
elharo: Looks like this PR needs to be updated to current master.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerTest.java
suztomo: Updated the line to clarify the return value. The return value is a valid `LinkageCheckReport` instance.
(Duplicate class files in class path do not mean a linkage error.)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Merged.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -371,45 +373,47 @@ Path findClassLocation(String className) {
...
         com.google.common.reflect.ClassPath.from(classLoaderFromJar);
 
-    return classPath.getAllClasses();
+    return classPath.getAllClasses().stream()
+        .map(ClassInfo::getName)
+        .collect(toImmutableSet());
   }
 
-  static ImmutableSet<JavaClass> listClassesInJar(Path jarPath) throws IOException {
-    Repository repository = createClassRepository(ImmutableList.of(jarPath));
```
suztomo: This (creating a `Repository` with a new class path that has only this jar file) was the cause of the bug. When listing `JavaClass`, it was not using the class path passed to `ClassDumper` constructor.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -86,23 +87,26 @@ private static Repository createClassRepository(List<Path> paths) {
   static ClassDumper create(List<Path> jarPaths) throws IOException {
     ClassLoader systemClassLoader = ClassLoader.getSystemClassLoader();
     ClassLoader extensionClassLoader = systemClassLoader.getParent();
+    jarPaths.forEach(
+        jar -> {
+          checkArgument(Files.isRegularFile(jar), "The jar file is not a regular file: %s", jar);
```
suztomo: Previously, when an invalid input was given, it was caught at this function. But it's no longer the case.
https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/496/files#diff-6e13c859770f686d992e68ec76956e75R103
Without `checkArgument`, `ClassDumper.create()` would succeed even with invalid input.

suztomo:
@elharo PTAL.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -376,65 +379,46 @@ Path findClassLocation(String className) {
   static ImmutableMap<String, Path> mapClassToJar(List<Path> jars) throws IOException {
```
netdpb: You could implement this method based on `mapJarToClasses` and avoid some code duplication:

```java
return ImmutableMap.copyOf(
    Maps.transformValues(
        Multimaps.asMap(mapJarToClasses(jars).inverse()), jars -> Iterables.get(jars, 0)));
```

But I have a question: Why is it important to know the _first_ jar that contains a given class, and not the rest? How is this method used?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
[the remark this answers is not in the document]
suztomo: Updated. That's nice to report invalid inputs at once.

PR #485, opened 2019-03-01
Read BOM via Path

suztomo:
Fixes #79.

To discuss with Elliotte about the design when he comes back.

- RepositoryUtility.readBom now takes `Path` to a pom.xml
  This design does not require a local repository to have the BOM installed.
- DashboardMain.main to take a path to BOM as the argument.
  Kokoro's periodic build script now calls it with the argument

This `readBom` will help to run DashboardTest using a fixed pom.xml (#466).

Dependencies to locally-installed cloud-oss-bom has been removed. Fixes #521 .

PR #517, opened 2019-03-18
added test that artifact exist in maven central

### LINKED ISSUE #72: Test artifacts exist
The BOM needs a test that all the artifacts referenced are in Maven Central.

Just realized I incorrectly hyphenated one. Found it in another test but there should be a direct test for this. The libraries.json tests in appengine-plugins-core have code like this.

vgrankina:
fix #72

### DIFF
--- a/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/BomTest.java
+++ b/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/BomTest.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2018 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.opensource.dashboard;
+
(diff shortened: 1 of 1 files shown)


dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/BomTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
+
+public class BomTest {
+
+  @Test
+  public void testArtifactsExist() throws ArtifactDescriptorException, IOException {
+    DefaultArtifact bom = new DefaultArtifact("com.google.cloud:cloud-oss-bom:pom:1.0.0-SNAPSHOT");
```
elharo: Thanks. I'm going to see if we can get #485 in first, since that will allow this test to read from the current code instead of the maven repo.

suztomo: 485 has been merged to master. The change removed a necessity of installing cloud-oss-bom locally.

@vgrankina Here is example code to load a BOM with relative path: https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/485/files#diff-c64c027455014b27d3fae482725a3938R134

vgrankina: test is updated

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/BomTest.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright 2018 Google LLC.
```
elharo: 2019

vgrankina: done

PR #527, opened 2019-03-19
[not for merge] Pair-wise comparison league table

suztomo:
This is the one that takes 9 hours to generate a table.


suztomo:
The base branch is too old.

PR #564, opened 2019-04-15
rename BOM artifact ID to libraries-bom

elharo:
Putting this out here for discussion. 
Secondary purpose: see how much work it is to rename this thing.


elharo:
Right

PR #566, opened 2019-04-15
Format error strings for enforcer rule

### LINKED ISSUE #560: More human friendly error formatting in enforcer rule output
When the maven plugin finds linkage errors they are output  like this: 

MethodSymbolReference{sourceClassName=net.bytebuddy.agent.VirtualMachine$ForHotSpot$OnUnix, targetClassName=org.newsclub.net.unix.AFUNIXSocket, methodName=setSoTimeout, interfaceMethod=false, descriptor=(I)V}, reason: CLASS_NOT_FOUND, target class location not found, isReachable: true
  

This is a simple dump of `toString`. We can do much better. Maybe something like 

Method org.newsclub.net.unix.AFUNIXSocket.setSoTimeout(int) not found. Invoked by net.bytebuddy.agent.VirtualMachine.ForHotSpot.OnUnix.connect()

Do **not** simply change the `toString` method. We need something we can iterate on and improve to provide detailed, actionable warnings. For instance we might want a stack trace that shows exactly how you get to the relevant method from the entry point. 

Or perhaps what we need is something more like we currently present in the dashboard:

```
4 target classes causing linkage errors referenced from 4 source classes.

org.apache.tools.ant.Task is not found, referenced from 1 source class
    org.mockito.cglib.transform.AbstractProcessTask

org.apache.tools.ant.types.FileSet is not found, referenced from 1 source class
    org.mockito.cglib.transform.AbstractProcessTask

org.apache.tools.ant.DirectoryScanner is not found, referenced from 1 source class
    org.mockito.cglib.transform.AbstractProcessTask

org.apache.tools.ant.BuildException is not found, referenced from 1 source class
    org.mockito.cglib.transform.AbstractProcessTask

The following paths to the jar file from BOM are found in the dependency tree.
    io.grpc:grpc-testing:1.19.0 / org.mockito:mockito-core:1.9.5
```

elharo:
@netdpb 
as opposed to simply using the `toString` from AutoValue. 
fixes #560

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FieldSymbolReference.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FieldSymbolReference.java
@@ -23,10 +23,18 @@ import com.google.auto.value.AutoValue;
  */
 @AutoValue
 abstract class FieldSymbolReference implements SymbolReference {
+  
   /**
-   * Returns the field name of the reference.
+   * Returns the name of the referenced field.
    */
   abstract String getFieldName();
+  
+  @Override
+  public String getDisplayString() {
+    return this.getTargetClassName()
+        + " is not found, referenced from "
+        + this.getSourceClassName();
+  }
 
   static Builder builder() {
     return new AutoValue_FieldSymbolReference.Builder();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FieldSymbolReference.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FieldSymbolReference.java
@@ -31,9 +31,8 @@ abstract class FieldSymbolReference implements SymbolReference {
   
   @Override
   public String getErrorString() {
-    return this.getTargetClassName() + "." + getFieldName()
-        + " is not found, referenced from "
-        + this.getSourceClassName();
+    return getTargetClassName() + "." + getFieldName() + " is not found, referenced from "
+        + getSourceClassName();
   }
 
   static Builder builder() {
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FieldSymbolReference.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FieldSymbolReference.java
@@ -30,8 +30,8 @@ abstract class FieldSymbolReference implements SymbolReference {
   abstract String getFieldName();
   
   @Override
-  public String getDisplayString() {
-    return this.getTargetClassName()
+  public String getErrorString() {
+    return this.getTargetClassName() + "." + getFieldName()
         + " is not found, referenced from "
         + this.getSourceClassName();
   }
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
@@ -149,7 +149,7 @@ public abstract class JarLinkageReport {
     return builder.build();
   }
 
-  public String getErrorString() {
+  String getErrorString() {
     String indent = "  ";
     StringBuilder builder = new StringBuilder();
     int totalErrors = getCauseToSourceClassesSize();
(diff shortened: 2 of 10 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FieldSymbolReference.java
```diff
@@ -23,10 +23,18 @@
...
+  
   /**
-   * Returns the field name of the reference.
+   * Returns the name of the referenced field.
    */
   abstract String getFieldName();
+  
+  @Override
+  public String getErrorString() {
+    return this.getTargetClassName() + "." + getFieldName()
```
netdpb: Don't need `this`es.

netdpb: Consider `String.format()` for this one (and the others, frankly). It'd be easier to read.

netdpb: Actually, it looks like these all might follow the same format: "<reference> is not found, referenced from <source>". Suggests a base class implementation.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
```diff
@@ -148,4 +148,25 @@ public JarLinkageReport reachableErrors() {
...
     return builder.build();
   }
+
+  public String getErrorString() {
+    String indent = "  ";
+    StringBuilder builder = new StringBuilder();
+    int totalErrors = getCauseToSourceClassesSize();
+
+    builder.append(getJarPath().getFileName() + " (" + totalErrors + " errors):\n");
+    for (SymbolNotResolvable<ClassSymbolReference> missingClass : getMissingClassErrors()) {
```
netdpb: Only if you want, this can be condensed:

```java
String.format("%s (%s errors):\n%s\n",
    getJarPath().getFileName(),
    totalErrors,
    Stream.of(getMissingClassErrors(), getMissingMethodErrors(), getMissingFieldErrors())
        .flatMap(Collection::stream)
        .map(error -> indent + error.getReference().getErrorString() + "\n")
        .collect(joining("\n"));
```

Or you could use that stream with a `forEachOrdered(builder::append)` if you want.

Or you could use Guava's `Iterables.concat(getMissingClassErrors(), getMissingMethodErrors(), getMissingFieldErrors())` and a single for loop.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSymbolReference.java
```diff
@@ -55,4 +55,13 @@ static Builder builder() {
     abstract Builder setSourceClassName(String className);
     abstract MethodSymbolReference build();
   }
+
+  @Override
+  public String getErrorString() {
+    return this.getTargetClassName()
```
netdpb:
```suggestion
    return getTargetClassName()
```

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/MethodSymbolReferenceTest.java
```diff
@@ -20,20 +20,29 @@
 import org.junit.Test;
 
 public class MethodSymbolReferenceTest {
+  
+  private MethodSymbolReference methodSymbolReference =
```
netdpb:
```suggestion
  private static final MethodSymbolReference REFERENCE =
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSymbolReference.java
elharo: done

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/MethodSymbolReferenceTest.java
elharo: done

PR #570, opened 2019-04-17
Replaced deprecated Correspondence constructors

IRuizM:
Replaced deprecated Correspondence constructor with Correspondence.from.

fix #565

PR #572, opened 2019-04-17
Add fmt-maven-plugin so users can run fmt:format

loosebazooka:
- Does NOT automatically enforce formatting rules.
- Will apply directly on source google-java-format.
- Should help users with issues like https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/570/files/5d454478a7f79ca469226f93576951687789497b#r276329495

### DIFF
--- a/dashboard/pom.xml
+++ b/dashboard/pom.xml
@@ -74,6 +74,10 @@
           <mainClass>com.google.cloud.tools.opensource.dashboard.DashboardMain</mainClass>
         </configuration>
       </plugin>
+      <plugin>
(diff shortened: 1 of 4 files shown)


loosebazooka:
The google-java-format library is used and this plugin is just a wrapper: https://github.com/coveo/fmt-maven-plugin#fmt-maven-plugin

elharo:
The more I think about this, the more I'm thinking that putting formatting in a build tool is wrong. Formatting is for the editor, not the build tool. Maven should be building code, not changing it.

PR #573, opened 2019-04-18
cause --> target

elharo:
@netdpb This begins removing the confusing terminology "cause" from the code base. I think we can also get rid of the `LinkageErrorCause` class completely since it's jsut a wrapper around a `SymbolNotResolvable` used as a key in the map. However that started to turn into a bigger change, so I thought I should send this out here before the changes get too big to follow.

### DIFF
--- a/dashboard/src/main/resources/templates/macros.ftl
+++ b/dashboard/src/main/resources/templates/macros.ftl
@@ -15,7 +15,7 @@
       causing linkage errors referenced from
       ${pluralize(sourceClassCount, "source class", "source classes")}.
     </p>
-    <#list jarLinkageReport.getCauses() as cause >
+    <#list jarLinkageReport.getUnresolvableTargets() as cause >
       <#assign sourceClasses = jarLinkageReport.getSourceClasses(cause) />
       <p class="jar-linkage-report-cause">${cause?html}, referenced from ${
         pluralize(sourceClasses?size, "source class", "source classes")?html}
--- a/dashboard/src/main/resources/templates/macros.ftl
+++ b/dashboard/src/main/resources/templates/macros.ftl
@@ -15,9 +15,9 @@
       causing linkage errors referenced from
       ${pluralize(sourceClassCount, "source class", "source classes")}.
     </p>
-    <#list jarLinkageReport.getUnresolvableTargets() as cause >
-      <#assign sourceClasses = jarLinkageReport.getSourceClasses(cause) />
-      <p class="jar-linkage-report-cause">${cause?html}, referenced from ${
+    <#list jarLinkageReport.getUnresolvableTargets() as unresolvableTarget >
+      <#assign sourceClasses = jarLinkageReport.getSourceClasses(unresolvableTarget) />
+      <p class="jar-linkage-report-cause">${unresolvableTarget?html}, referenced from ${
         pluralize(sourceClasses?size, "source class", "source classes")?html}
         <button onclick="toggleSourceClassListVisibility(this)"
                 title="Toggle visibility of source class list">▶
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
@@ -40,8 +40,8 @@ import java.util.Set;
 public abstract class JarLinkageReport {
 
   /**
-   * Returns the absolute path of the jar file that was checked. This is where the source classes
-   * were found.
+   * Returns the absolute path of the jar file that was checked. The source classes
+   * of the reported errors are all in this JAR.
    */
   public abstract Path getJarPath();
 
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
@@ -34,13 +34,14 @@ import java.util.List;
 import java.util.Set;
 
 /**
- * The result of checking linkages in one jar file.
+ * Linkage problems found in one jar file given a specific classpath.
  */
 @AutoValue
 public abstract class JarLinkageReport {
 
   /**
-   * Returns the absolute path of the jar file containing source classes of linkage errors
+   * Returns the absolute path of the jar file that was checked. This is where the source classes
+   * were found.
    */
   public abstract Path getJarPath();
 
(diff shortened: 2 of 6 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
```diff
@@ -34,13 +34,14 @@
...
 /**
- * The result of checking linkages in one jar file.
+ * Linkage problems found in one jar file given a specific classpath.
  */
 @AutoValue
 public abstract class JarLinkageReport {
 
   /**
-   * Returns the absolute path of the jar file containing source classes of linkage errors
+   * Returns the absolute path of the jar file that was checked. This is where the source classes
```
netdpb:
```suggestion
   * Returns the absolute path of the jar file that was checked. The source classes of the reported errors are all in this JAR.
```

netdpb:
I feel like there's a bigger refactoring that could clarify a lot of names and concepts. Right now `SymbolReference` combines the (target) symbol being referred to with the (source) class that is referring to it. That means you have to later separate out the "cause" (which I think is the symbol being referred to and the problem with it, what I would call the "error") and the source of that "cause".

Rather than that, would it be simpler to have the following?
1. a `Symbol` type (subtypes `MethodSymbol`, `FieldSymbol`, `ClassSymbol`): basically `SymbolReference` without the source class
1. a `ClassInJar` type (JAR + class)
1. a `SymbolError` type (symbol + reason + optional `ClassInJar` for anything other than top-level class not found): replaces `SymbolNotResolvable`
1. a multimap from source `ClassInJar` to `SymbolError`

Then you don't have a notion of a "cause": you have `SymbolError`s representing erroneous references from source `ClassInJar`s. You can invert that multimap to show all source classes for a given error. You can also show all errors in a target class, separated by error type and source class.

Food for thought for a future CL.

dashboard/src/main/resources/templates/macros.ftl
```diff
@@ -15,7 +15,7 @@
       causing linkage errors referenced from
       ${pluralize(sourceClassCount, "source class", "source classes")}.
     </p>
-    <#list jarLinkageReport.getCauses() as cause >
+    <#list jarLinkageReport.getUnresolvableTargets() as cause >
```
netdpb:
```suggestion
    <#list jarLinkageReport.getUnresolvableTargets() as unresolvableTarget >
```

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/JarLinkageReportTest.java
```diff
@@ -153,7 +153,7 @@ public void testGetErrorString() {
   @Test
   public void testGetCauseToSourceClasses() {
```
netdpb: Rename the test method?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
```diff
@@ -92,15 +93,11 @@ public String toString() {
     return builder.toString();
   }
 
-  /** Returns map from the cause of linkage errors to class names affected by the errors. */
-  // TODO this is only used by formatJarLinkageReport in macros.ftl. We should be able to
-  // refactor this to make it a lot clearer by removing ImmutableMultimap from the API.
+  /** 
+   * Map missing classes and members to classes that refer to those items.
```
netdpb:
```suggestion
   * Returns a map from linkage errors to classes that have them.
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
```diff
@@ -92,15 +93,11 @@ public String toString() {
...
+   */
   @Memoized
-  ImmutableMultimap<LinkageErrorCause, String> getCauseToSourceClasses() {
-    return mapCauseToSourceClasses();
-  }
-
-  private ImmutableMultimap<LinkageErrorCause, String> mapCauseToSourceClasses() {
+  ImmutableMultimap<LinkageErrorCause, String> getTargetToSources() {
```
netdpb: The `LinkageErrorCause` isn't the target, right? It's the error (minus the source).

Should this be `getErrorSources()`?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
```diff
@@ -141,29 +138,29 @@ public String toString() {
    * @return the total number of classes that refer to missing or inaccessible members
    */
   public int getErrorCount() {
-    return getCauseToSourceClasses().size();
+    return getTargetToSources().size();
   }
   
   /**
-   * @return the total number of missing or inaccessible members
+   * @return the total number of missing or inaccessible targets
```
netdpb: This isn't true, is it? You can have more than one LEC for a target class. This is the total number of errors (reasons for problematic symbols) independent of how many source classes trigger that error.

The method name is wrong too.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/JarLinkageReport.java
```diff
@@ -141,29 +138,29 @@ public String toString() {
...
    */
   public int getTargetClassCount() {
-    return getCauses().size();
+    return getUnresolvableTargets().size();
   }
   
   /**
-   * @return the references to missing or inaccessible members
+   * @return the references to missing or inaccessible targets
```
netdpb: See above.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorCause.java
```diff
@@ -22,6 +22,8 @@
 @AutoValue
 // TODO "cause" is unclear. Is it the target or the source? I don't know.
 // Rename things to remove the word "cause" and use more specific terminology.
+// why do we need this class at all? It's just a wrapper around a SymbolNotResolvable.
```
netdpb: I think SNR has the bad target symbol and the reason it's bad (and whether the source of the error is reachable [?]). This has just the target symbol.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorCause.java
```diff
@@ -32,7 +34,7 @@
    * This is a class name, field name, or method name. 
    */
   abstract String getSymbol();
-
+
```
netdpb: Revert

PR #579, opened 2019-04-22
To remove unnecessary argument validation in DashboardMain

### LINKED ISSUE #578: Fix DashboardMain argument length validation
Fix DashboardMain argument length validation. When DashboardMain is passed with `-f boms/cloud-oss-bom/pom.xml`, it complains with "Please specify path to pom.xml or Maven coordinates for a BOM." (screenshot below)

image

The following code in `DashboardMain.main` expects the argument to contain a single string "-f boms/cloud-oss-bom/pom.xml". (This is how Maven-exec module passes arguments)

```
    if (arguments.length != 1) {
      System.err.println("Please specify path to pom.xml or Maven coordinates for a BOM.");
      return;
    }
```

However this is not compatible with normal main method invocation.

suztomo:
Fixes #578 

Without the unnecessary validation, DashboardMain can reject invalid arguments:

image

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -95,10 +95,6 @@ public class DashboardMain {
       throws IOException, TemplateException, RepositoryException, URISyntaxException,
           PlexusContainerException, ComponentLookupException, ProjectBuildingException,
           ParseException {
-    if (arguments.length != 1) {
-      System.err.println("Please specify path to pom.xml or Maven coordinates for a BOM.");
-      return;
-    }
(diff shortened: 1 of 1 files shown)


dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -95,10 +95,6 @@ public static void main(String[] arguments)
       throws IOException, TemplateException, RepositoryException, URISyntaxException,
           PlexusContainerException, ComponentLookupException, ProjectBuildingException,
           ParseException {
-    if (arguments.length != 1) {
```
suztomo: This validation was not a problem when Maven-exec-plugin passes arguments. (See https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/578)

PR #580, opened 2019-04-22
To distinguish BOM update and dependency update in  Global Upper Bounds Fixes

### LINKED ISSUE #568: Dashboard can suggest a global upper bounds fix for an artifact itself:
This is interesting. Not sure whether we should filter this out completely or replace it with a message that the BOM is out of date.

Global Upper Bounds Fixes

Suggested updates to bring this artifact into sync with the highest versions of its dependencies used by any artifact in the BOM:

    Upgrade com.google.protobuf:protobuf-java:jar:3.6.1 to com.google.protobuf:protobuf-java:jar:3.7.1:

    Add this dependency element to the pom.xml for com.google.protobuf:protobuf-java:3.6.1:

    <dependency>
      <groupId>com.google.protobuf</groupId>
      <artifactId>protobuf-java</artifactId>
      <version>3.7.1</version>
    </dependency>

If the pom.xml for com.google.protobuf:protobuf-java:3.6.1 already includes this dependency, update the version of the existing dependency element. Otherwise add a new dependency element to the dependencyManagement section.

suztomo:
Fixes #568.

Screenshot:

image

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -326,9 +326,7 @@ public class DashboardMain {
                   .equals(Artifacts.toCoordinates(dependencyPath.get(0)))));
 
       Map<String, Object> templateData = new HashMap<>();
-      templateData.put("groupId", artifact.getGroupId());
-      templateData.put("artifactId", artifact.getArtifactId());
-      templateData.put("version", artifact.getVersion());
+      templateData.put("artifact", artifact);
       templateData.put("updates", convergenceIssues);
       templateData.put("upperBoundFailures", upperBoundFailures);
       templateData.put("globalUpperBoundFailures", globalUpperBoundFailures);
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -95,10 +95,6 @@ public class DashboardMain {
       throws IOException, TemplateException, RepositoryException, URISyntaxException,
           PlexusContainerException, ComponentLookupException, ProjectBuildingException,
           ParseException {
-    if (arguments.length != 1) {
-      System.err.println("Please specify path to pom.xml or Maven coordinates for a BOM.");
-      return;
-    }
     DashboardArguments dashboardArguments = DashboardArguments.readCommandLine(arguments);
     Path output =
         dashboardArguments.hasFile()
--- a/dashboard/src/main/resources/templates/component.ftl
+++ b/dashboard/src/main/resources/templates/component.ftl
@@ -6,12 +6,13 @@
   <#assign version = artifact.getVersion()>
   <head>
     <meta charset="utf-8" />
-    <title>Google Cloud Platform Dependency Analysis Report for ${artifact}</title>
+    <title>Google Cloud Platform Dependency Analysis Report for ${groupId}:${artifactId}:${version}
+    </title>
     <link rel="stylesheet" href="dashboard.css" />
     <script src="dashboard.js"></script>
   </head>
   <body>
-    <h1>Dependency Analysis of ${artifact}</h1>
+    <h1>Dependency Analysis of ${groupId}:${artifactId}:${version}</h1>
     
     
     <h2 id="global-upper-bounds">Global Upper Bounds Check</h2>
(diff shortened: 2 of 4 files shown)


dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -326,9 +326,7 @@ private static ArtifactResults generateArtifactReport(
                   .equals(Artifacts.toCoordinates(dependencyPath.get(0)))));
 
       Map<String, Object> templateData = new HashMap<>();
-      templateData.put("groupId", artifact.getGroupId());
-      templateData.put("artifactId", artifact.getArtifactId());
-      templateData.put("version", artifact.getVersion());
```
suztomo: These three variables are now defined in component.ftl using `artifact`.

dashboard/src/main/resources/templates/component.ftl
```diff
@@ -24,21 +27,42 @@
...
         <#list globalUpperBoundFailures as lower, upper>
-          <li>Upgrade ${lower} to ${upper}:
-          
-          <p>Add this dependency element to the pom.xml for ${groupId}:${artifactId}:${version}:</p>
-          
-<pre class="suggested-dependency-mediation"><code>&lt;dependency>
-  &lt;groupId>${upper.getGroupId()}&lt;/groupId>
-  &lt;artifactId>${upper.getArtifactId()}&lt;/artifactId>
-  &lt;version>${upper.getVersion()}&lt;/version>
+          <#if lower == artifact>
```
suztomo: Now it distinguish whether the upgrade is the BOM member or not.

dashboard/src/main/resources/templates/component.ftl
```diff
@@ -1,14 +1,17 @@
 <!DOCTYPE html>
 <html lang="en-US">
   <#include "macros.ftl">
+  <#assign groupId = artifact.getGroupId()>
+  <#assign artifactId = artifact.getArtifactId()>
+  <#assign version = artifact.getVersion()>
   <head>
     <meta charset="utf-8" />
-    <title>Google Cloud Platform Dependency Analysis Report for ${groupId}:${artifactId}:${version}</title>
+    <title>Google Cloud Platform Dependency Analysis Report for ${artifact}</title>
```
elharo: revert; it's unclear what toString does here

dashboard/src/main/resources/templates/component.ftl
```diff
@@ -1,14 +1,17 @@
...
     <link rel="stylesheet" href="dashboard.css" />
     <script src="dashboard.js"></script>
   </head>
   <body>
-    <h1>Dependency Analysis of ${groupId}:${artifactId}:${version}</h1>
+    <h1>Dependency Analysis of ${artifact}</h1>
```
elharo: ditto; weird toString output has been a problem lately

dashboard/src/main/resources/templates/component.ftl
elharo: What's the equals semantics here? identity or equality?

dashboard/src/main/resources/templates/component.ftl
```diff
@@ -24,21 +27,42 @@
...
-          
-          </li>
+
+            </li>
+          </#if>
         </#list>
       </ul>
       
-      <p>If the pom.xml for ${groupId}:${artifactId}:${version} already includes this dependency,
+      <p>If the pom.xml for ${artifact} already includes this dependency,
```
elharo: ditto, no toString

dashboard/src/main/resources/templates/component.ftl
suztomo: Makes sense. Reverted.

dashboard/src/main/resources/templates/component.ftl
suztomo: Reverted.

dashboard/src/main/resources/templates/component.ftl
suztomo: Reverted.

dashboard/src/main/resources/templates/component.ftl
suztomo: It's equality ([Freemarker Expressions - Comparison](https://freemarker.apache.org/docs/dgui_template_exp.html#dgui_template_exp_comparison)). But I updated the code to compare only the coordinates, without 'file' attribute of Artifact instance.

suztomo:
Updated; the sentence should be within the list element. Updated the screenshot of this PR.
PTAL.

example-problems/no-such-method-error-signature-mismatch/pom.xml
```diff
@@ -39,6 +39,34 @@
           <mainClass>io.grpc.internal.App</mainClass>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
elharo: This change doesn't seem related to this PR.

suztomo: Sorry, missed git operation. Removed.

suztomo:
I see "TOOL ABORTED" message in Fusion. Going to ask Kokoro team.

PR #621, opened 2019-05-23
cloud-oss-bom: import google-http-client-bom

chingor13:
google-http-client provides a bom of it's artifacts. We should use that rather than duplicating the list of artifacts.

### DIFF
--- a/boms/cloud-oss-bom/pom.xml
+++ b/boms/cloud-oss-bom/pom.xml
@@ -84,54 +84,11 @@
 
       <!-- google-http-java-client from https://github.com/googleapis/google-http-java-client -->
      <dependency>
-       <groupId>com.google.http-client</groupId>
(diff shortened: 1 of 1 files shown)


elharo:
java.lang.AssertionError: expected:<187> but was:<188>
	at com.google.cloud.tools.opensource.dependencies.RepositoryUtilityTest.testReadBom_path(RepositoryUtilityTest.java:59)


What extra dep is being added here?

I recall we deliberatley excluced some of the google-http-java-client dependencies, but those might have been dropped in the latest version.

chingor13:
Looks like it's the `google-http-client-findbugs` artifact

elharo:
There might be a reason for that. That plugin is pulling in findbugs 3.0.1 whereas 3.0.2 is used in the BOM. You can try updating the test and see if anything else fails.

PR #631, opened 2019-05-28
Dashboard to use new value class

### LINKED ISSUE #630: Dashboard to use new value classes
The last item of the refactoring (#574)

Given `ImmutableSetMultimap<ClassFile, SymbolProblem>`, show symbol errors grouped by 
- jar in source ClassFile
  - SymbolProblem
    - list class names in the source class file

suztomo:
Fixes #630 .

- inversed `findSymbolProblems()`'s return value because the caller was always inversing it.
- Deletion of the old value classes will come in another PR after this.
- A symbol problem caused by CLASS_NOT_FOUND has a ClassSymbol regardless of its origin (class reference, method reference, or field reference). This helps uniquely count symbol problems.
- Dashboard shows counting of references rather than source classes.

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -18,7 +18,6 @@ package com.google.cloud.tools.opensource.dashboard;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
-import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.ImmutableTable.toImmutableTable;
 
 import com.google.common.collect.Table;
@@ -35,7 +34,6 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
@@ -52,6 +50,7 @@ import freemarker.template.Template;
 import freemarker.template.TemplateException;
 import freemarker.template.TemplateHashModel;
 import freemarker.template.Version;
+import javax.annotation.Nullable;
 import org.apache.commons.cli.ParseException;
 import org.eclipse.aether.RepositoryException;
 import org.eclipse.aether.artifact.Artifact;
@@ -165,19 +164,15 @@ public class DashboardMain {
     copyResource(output, "js/dashboard.js");
     Configuration configuration = configureFreemarker();
 
-    // BOM member coordinates -> SymbolProblem -> Set<ClassFile>
-    ImmutableTable<String, SymbolProblem, ImmutableSet<ClassFile>> symbolProblemTable =
-        createSymbolProblemTable(symbolProblems, jarToDependencyPaths);
-
     List<ArtifactResults> table =
-        generateReports(configuration, output, cache, symbolProblemTable, jarToDependencyPaths);
+        generateReports(configuration, output, cache, symbolProblems, jarToDependencyPaths);
 
     generateDashboard(
         configuration,
         output,
         table,
         cache.getGlobalDependencies(),
-        symbolProblemTable,
+        symbolProblems,
         jarToDependencyPaths);
 
     return output;
@@ -202,61 +197,21 @@ public class DashboardMain {
   }
 
   /**
-   * Returns a table where rows are the Maven coordinates of BOM members, columns are symbol
-   * problems, and the values are sets of class files.
-   *
-   * <p>For example, {@code classFiles = symbolProblemTable.get("com.google.abc:foo:1.0.0",
-   * ProblemA)} means that the jar files containing the {@code classFiles} are in the dependency
-   * tree of{@code "com.google.abc:foo:1.0.0"} and that the {@code classFiles} reference the
-   * symbol of {@code ProblemA}.
+   * Returns mapping from BOM member to jar files that are in the dependency tree of the BOM member.
    */
-  @VisibleForTesting
-  static ImmutableTable<String, SymbolProblem, ImmutableSet<ClassFile>> createSymbolProblemTable(
-      ImmutableSetMultimap<SymbolProblem, ClassFile> symbolProblems,
+  private static ImmutableSetMultimap<String, Path> bomMemberToJars(
       ListMultimap<Path, DependencyPath> jarToDependencyPaths) {
-    // Coordinates of BOM member
-    //   -> Symbol Problem
-    //     -> ClassFiles in the BOM member or its dependencies
-    Table<String, SymbolProblem, ImmutableSet.Builder<ClassFile>> symbolProblemTable
-        = HashBasedTable.create();
-
     // Coordinates of BOM member -> One or more jar files (the BOM member and its dependencies)
-    ImmutableMultimap.Builder<Path, String> jarToCoordinatesBuilder = ImmutableMultimap.builder();
+    ImmutableSetMultimap.Builder<String, Path> jarToCoordinatesBuilder = ImmutableSetMultimap
+        .builder();
     for (Path path : jarToDependencyPaths.keySet()) {
       for (DependencyPath dependencyPath : jarToDependencyPaths.get(path)) {
         Artifact artifact = dependencyPath.get(0);
-        jarToCoordinatesBuilder.put(path, Artifacts.toCoordinates(artifact));
+        jarToCoordinatesBuilder.put(Artifacts.toCoordinates(artifact), path);
       }
     }
 
-    ImmutableMultimap<Path, String> jarToBomMembers = jarToCoordinatesBuilder.build();
-
-    for(SymbolProblem problem : symbolProblems.keySet()) {
-      ImmutableSet<ClassFile> classFiles = symbolProblems.get(problem);
-      ImmutableListMultimap<Path, ClassFile> jarToClassFile =
-          Multimaps.index(classFiles, ClassFile::getJar);
-
-      for (Path jar : jarToClassFile.keySet()) {
-        // For each BOM member that contains this jar in dependencies
-        for(String coordinates: jarToBomMembers.get(jar)) {
-          ImmutableSet.Builder<ClassFile> classFilesForProblem = symbolProblemTable
-              .get(coordinates, problem);
-          if (classFilesForProblem == null) {
-            classFilesForProblem = ImmutableSet.builder();
-            symbolProblemTable.put(coordinates, problem, classFilesForProblem);
-          }
-
-          ImmutableList<ClassFile> classFilesForJar = jarToClassFile.get(jar);
-          classFilesForProblem.addAll(classFilesForJar);
-        }
-      }
-    }
-
-    return symbolProblemTable.cellSet().stream().collect(ImmutableTable.toImmutableTable(
-        Cell::getRowKey,
-        Cell::getColumnKey,
-        cell -> cell.getValue().build()
-    ));
+    return jarToCoordinatesBuilder.build();
   }
 
   @VisibleForTesting
@@ -264,8 +219,10 @@ public class DashboardMain {
       Configuration configuration,
       Path output,
       ArtifactCache cache,
-      ImmutableTable<String, SymbolProblem, ImmutableSet<ClassFile>> symbolProblemTable,
+      ImmutableSetMultimap<SymbolProblem, ClassFile> symbolProblems,
       ListMultimap<Path, DependencyPath> jarToDependencyPaths) {
+
+    ImmutableSetMultimap<String, Path> bomMemberToJars = bomMemberToJars(jarToDependencyPaths);
 
     Map<Artifact, ArtifactInfo> artifacts = cache.getInfoMap();
     List<ArtifactResults> table = new ArrayList<>();
@@ -278,6 +235,9 @@ public class DashboardMain {
           table.add(unavailable);
         } else {
           Artifact artifact = entry.getKey();
+          ImmutableSet<Path> jarsInDependencyTree = bomMemberToJars
+              .get(Artifacts.toCoordinates(artifact));
+
           ArtifactResults results =
               generateArtifactReport(
                   configuration,
@@ -285,7 +245,7 @@ public class DashboardMain {
                   artifact,
                   entry.getValue(),
                   cache.getGlobalDependencies(),
-                  symbolProblemTable.row(Artifacts.toCoordinates(artifact)),
+                  indexByJar(symbolProblems, jarsInDependencyTree),
                   jarToDependencyPaths);
           table.add(results);
         }
@@ -341,7 +301,7 @@ public class DashboardMain {
       Artifact artifact,
       ArtifactInfo artifactInfo,
       List<DependencyGraph> globalDependencies,
-      Map<SymbolProblem, ImmutableSet<ClassFile>> symbolProblems,
+      ImmutableTable<Path, SymbolProblem, ImmutableSet<String>> symbolProblemTable,
       ListMultimap<Path, DependencyPath> jarToDependencyPaths)
       throws IOException, TemplateException {
 
@@ -366,7 +326,7 @@ public class DashboardMain {
 
       List<DependencyPath> dependencyPaths = completeDependencies.list();
 
-      int totalLinkageErrorCount = symbolProblems.keySet().size();
+      int totalLinkageErrorCount = symbolProblemTable.columnKeySet().size();
 
       ListMultimap<DependencyPath, DependencyPath> dependencyTree =
           DependencyTreeFormatter.buildDependencyPathTree(dependencyPaths);
@@ -379,9 +339,6 @@ public class DashboardMain {
               dependencyPath -> coordinates
                   .equals(Artifacts.toCoordinates(dependencyPath.get(0)))));
 
-      ImmutableTable<Path, SymbolProblem, ImmutableSet<String>> jarToSymbolProblemToClasses =
-          indexByJar(symbolProblems);
-
       Map<String, Object> templateData = new HashMap<>();
       templateData.put("artifact", artifact);
       templateData.put("updates", convergenceIssues);
@@ -390,7 +347,7 @@ public class DashboardMain {
       templateData.put("lastUpdated", LocalDateTime.now());
       templateData.put("dependencyTree", dependencyTree);
       templateData.put("dependencyRootNode", Iterables.getFirst(dependencyTree.values(), null));
-      templateData.put("jarToSymbolProblemToClasses", jarToSymbolProblemToClasses);
+      templateData.put("jarToSymbolProblemToClasses", symbolProblemTable);
       templateData.put("jarToDependencyPaths", jarToDependencyPathsForArtifact);
       templateData.put("totalLinkageErrorCount", totalLinkageErrorCount);
       report.process(templateData, out);
@@ -433,50 +390,35 @@ public class DashboardMain {
   }
 
   /**
-   * Returns a table where rows are jar files, columns are {@link SymbolProblem}, and values are
-   * class names referencing the problems.
+   * Returns a table where rows are jar files in {@code jarFilter}, columns are {@link
+   * SymbolProblem}, and values are class names referencing the problems. When {@code jarFilter} is
+   * {@code null}, jar files are not filtered.
+   *
+   * <p>For example, {@code classes = table.get("com.google.abc:foo:1.0.0", ProblemA)} means that
+   * the jar files containing the {@code classFiles} are in the dependency tree of{@code
+   * "com.google.abc:foo:1.0.0"} and that the {@code classes} reference the symbol of {@code
+   * ProblemA}.
    */
   private static ImmutableTable<Path, SymbolProblem, ImmutableSet<String>> indexByJar(
-      ImmutableTable<String, SymbolProblem, ImmutableSet<ClassFile>> coordinatesToProblems) {
-    ImmutableSetMultimap.Builder<SymbolProblem, ClassFile> builder = ImmutableSetMultimap.builder();
-    coordinatesToProblems.columnMap().forEach((symbolProblem, coordinatesToClassFile) -> {
-      ImmutableSet<ClassFile> classFilesCombined = coordinatesToClassFile
-          .values().stream()
-          .flatMap(Collection::stream)
-          .collect(toImmutableSet());
-      builder.putAll(symbolProblem, classFilesCombined);
-    });
-
-    // To adjust type for the input of the other indexByJar; ImmutableSetMultimap.asMap()'s value
-    // type is Collection, not ImmutableSet.
-    ImmutableMap<SymbolProblem, ImmutableSet<ClassFile>> problemToClassFiles = builder.build()
-        .asMap().entrySet().stream().collect(
-            toImmutableMap(
-                Entry::getKey,
-                entry -> ImmutableSet.copyOf(entry.getValue())));
-    return indexByJar(problemToClassFiles);
-  }
-
-  /**
-   * Returns a table where rows are jar files, columns are {@link SymbolProblem}, and values are
-   * class names referencing the problems.
-   */
-  private static ImmutableTable<Path, SymbolProblem, ImmutableSet<String>> indexByJar(
-      Map<SymbolProblem, ImmutableSet<ClassFile>> problemWithClassFiles) {
+      ImmutableSetMultimap<SymbolProblem, ClassFile> symbolProblems,
+      @Nullable Set<Path> jarFilter) {
     Table<Path, SymbolProblem, ImmutableSet.Builder<String>> table = HashBasedTable.create();
 
-    problemWithClassFiles.forEach((symbolProblem, classFiles) -> {
+    symbolProblems.asMap().forEach((symbolProblem, classFiles) -> {
       ImmutableListMultimap<Path, ClassFile> jarToClassFiles =
           Multimaps.index(classFiles, ClassFile::getJar);
       jarToClassFiles.forEach((jar, classFilesWithJar) -> {
+        if (jarFilter != null && !jarFilter.contains(jar)) {
+          return;
+        }
         ImmutableSet.Builder<String> classNames = table.get(jar, symbolProblem);
         if (classNames == null) {
           classNames = ImmutableSet.builder();
           table.put(jar, symbolProblem, classNames);
         }
 
-        ImmutableSet<String> newClassNames =
-            classFiles.stream().map(ClassFile::getClassName).collect(toImmutableSet());
+        ImmutableSet<String> newClassNames = classFiles.stream().map(ClassFile::getClassName)
+            .collect(toImmutableSet());
         classNames.addAll(newClassNames);
       });
     });
@@ -496,14 +438,14 @@ public class DashboardMain {
       Path output,
       List<ArtifactResults> table,
       List<DependencyGraph> globalDependencies,
-      ImmutableTable<String, SymbolProblem, ImmutableSet<ClassFile>> coordinatesToProblems,
+      ImmutableSetMultimap<SymbolProblem, ClassFile> symbolProblems,
       ListMultimap<Path, DependencyPath> jarToDependencyPaths)
       throws IOException, TemplateException {
     
     Map<String, String> latestArtifacts = collectLatestVersions(globalDependencies);
 
     ImmutableTable<Path, SymbolProblem, ImmutableSet<String>> jarToSymbolProblemToClasses =
-        indexByJar(coordinatesToProblems);
+        indexByJar(symbolProblems, null);
 
     Map<String, Object> templateData = new HashMap<>();
     templateData.put("table", table);
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -20,8 +20,6 @@ import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.collect.ImmutableTable.toImmutableTable;
 
-import com.google.common.collect.Table;
-import com.google.common.collect.Table.Cell;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -87,6 +85,8 @@ import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Multimaps;
 import com.google.common.collect.Sets;
+import com.google.common.collect.Table;
+import com.google.common.collect.Table.Cell;
 
 public class DashboardMain {
   public static final String TEST_NAME_LINKAGE_CHECK = "Linkage Errors";
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -17,6 +17,8 @@
 package com.google.cloud.tools.opensource.dashboard;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
 
 import com.google.common.collect.Table;
 import java.io.File;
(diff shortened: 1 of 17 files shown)


dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -192,8 +195,17 @@ static Configuration configureFreemarker() {
     return configuration;
   }
 
+  /**
+   * Returns a table where rows are the Maven coordinates of BOM members, columns are symbol
+   * problems, and the values are set of class files.
```
elharo: set --> sets

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -192,8 +195,17 @@ static Configuration configureFreemarker() {
...
+   *
+   * <p>{@code classFiles = symbolProblemTable.get("com.google.abc:foo:1.0.0", ProblemA)} means that
+   * the jar files of the {@code classFiles} are in the dependency tree of {@code
```
elharo: of --> containing?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -192,8 +195,17 @@ static Configuration configureFreemarker() {
...
+   * "com.google.abc:foo:1.0.0"} and that the {@code classFiles} are referencing the symbol of
```
elharo: are referencing --> reference

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -226,30 +237,25 @@ static Configuration configureFreemarker() {
         for(String coordinates: jarToBomMembers.get(jar)) {
           Set<ClassFile> classFilesForProblem = symbolProblemTable.get(coordinates, problem);
           if (classFilesForProblem == null) {
-            classFilesForProblem = new HashSet<>();
+            classFilesForProblem = Sets.newHashSet();
```
elharo: Per API docs "Note for Java 7 and later: this method is now unnecessary and should be treated as deprecated. Instead, use the HashSet constructor directly, taking advantage of the new "diamond" syntax."

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Fixed

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Updated

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Updated

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Now it uses `ImmutableSet.Builder` to have deterministic iteration order.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/FreemarkerTest.java
```diff
@@ -69,12 +69,9 @@ public static void cleanUp() throws IOException {
   }
 
   @Test
-  public void testCountFailures() throws IOException, TemplateException, ValidityException, ParsingException {
```
suztomo: This exception was grayed out.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -201,7 +199,8 @@ public LinkageCheckReport findLinkageErrors() {
       if (classDumper.catchesNoClassDefFoundError(sourceClassName)) {
         return Optional.empty();
       }
-      return Optional.of(new SymbolProblem(symbol, ErrorType.CLASS_NOT_FOUND, null));
+      ClassSymbol classSymbol = new ClassSymbol(symbol.getClassName());
+      return Optional.of(new SymbolProblem(classSymbol, ErrorType.CLASS_NOT_FOUND, null));
```
suztomo: Now a symbol problem caused by missing class for method references has class symbol.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -246,7 +245,8 @@ public LinkageCheckReport findLinkageErrors() {
...
```
suztomo: Now a symbol problem caused by missing class for field references has class symbol.

enforcer-rules/src/test/java/com/google/cloud/tools/dependencies/enforcer/LinkageCheckerRuleTest.java
```diff
@@ -287,40 +286,4 @@ public void testExecute_shouldFailBadBom() {
       // pass
     }
   }
-
-  @Test
-  public void testFormatSymbolProblems() {
```
suztomo: This has been moved to SymbolProblemTest

enforcer-rules/src/main/java/com/google/cloud/tools/dependencies/enforcer/LinkageCheckerRule.java
```diff
@@ -172,28 +170,6 @@ public void execute(@Nonnull EnforcerRuleHelper helper) throws EnforcerRuleExcep
     }
   }
 
-  @VisibleForTesting
-  static String formatSymbolProblems(ImmutableSetMultimap<ClassFile, SymbolProblem> symbolProblems) {
```
suztomo: This has been moved to SymbolProblem.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerTest.java
```diff
@@ -289,17 +310,37 @@ public void testFindInvalidReferences_nonExistentField() throws IOException {
             new ClassFile(firestoreJar, LinkageCheckReportTest.class.getName()),
             new FieldSymbol(
                 "com.google.firestore.v1beta1.FirestoreGrpc",
-                // This method is marked as private
```
suztomo: This was wrong comment.

enforcer-rules/src/test/java/com/google/cloud/tools/dependencies/enforcer/LinkageCheckerRuleTest.java
```diff
@@ -164,9 +164,8 @@ public void testExecute_shouldFailForBadProject() throws RepositoryException {
           "The rule should raise an EnforcerRuleException for artifacts missing dependencies");
     } catch (EnforcerRuleException ex) {
       // pass
-      verify(mockLog).error(ArgumentMatchers.startsWith("Linkage Checker rule found 384 errors."));
-      assertEquals(
-          "Failed while checking class path. See above error report.", ex.getMessage());
+      verify(mockLog).error(ArgumentMatchers.startsWith("Linkage Checker rule found 123 errors."));
```
suztomo: Now the dashboard uniquely count the missing class, regardless of method or field references.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -200,12 +200,6 @@ public void testDashboard_statisticBox() {
 
   @Test
   public void testLinkageReports() {
-    Nodes reports = details.query("//p[@class='jar-linkage-report']");
```
elharo: Why were these assertions deleted? This refactoring shouldn't materially change the output in the dashboard, right?

suztomo: Reverting that part.

suztomo:
@elharo PTAL. Restored the message in the dashboard.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -188,28 +196,34 @@ static Configuration configureFreemarker() {
...
       ListMultimap<Path, DependencyPath> jarToDependencyPaths) {
-    ImmutableMap<Path, JarLinkageReport> jarToLinkageReport =
-        linkageCheckReport.getJarLinkageReports().stream()
-            .collect(
-                toImmutableMap(JarLinkageReport::getJarPath, jarLinkageReport -> jarLinkageReport));
-
-    // Map from Artifact's coordinates to (unique) JarLinkageReports.
-    // Using string coordinates rather than Artifact class because its equality includes file.
-    ImmutableSetMultimap.Builder<String, JarLinkageReport> builder = ImmutableSetMultimap.builder();
+    // Coordinates of BOM member -> One or more jar files (the BOM member and its dependencies)
```
netdpb: This repeats information that's in the Javadoc, right?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -188,28 +196,34 @@ static Configuration configureFreemarker() {
...
+    ImmutableSetMultimap.Builder<String, Path> bomMemberToJars = ImmutableSetMultimap
+        .builder();
     for (Path path : jarToDependencyPaths.keySet()) {
```
netdpb: This could use `jarToDependencyPaths.forEach((jar, dependencyPath) -> {...});`

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -375,23 +390,68 @@ private static ArtifactResults generateArtifactReport(
     return upperBoundFailures;
   }
 
+  /**
+   * Returns a table where rows are jar files in {@code jarFilter}, columns are {@link
```
netdpb: Not every JAR in the filter will be in the table, right—because it's just a filter?

How about

Returns a table where the row keys are JAR files with problems, the column keys are problems, and the values are the classes in those JAR files with those problems.

If `jarFilter` is not null, the table will contain only rows for JAR files in that set.

netdpb: I wonder if this table is as helpful as it should be. You're partitioning the problems multimap by jar file, right? And as far as I can tell in the templates you're iterating over the jars. With the table, you're getting a `Map<SymbolProblem, ImmutableSet<String>>` for each jar row key.

If this method returned an `ImmutableMap<Path, ImmutableSetMultimap<SymbolProblem, ClassFile>>`, then you'd get an `ImmutableSetMultimap<SymbolProblem, ClassFile>` for each jar key, which you could then invert as you needed to, or `forEach` over, or what have you.

And the implementation of this method would become very simple. Something like:

```java
return ImmutableMap.copyOf(
    Maps.transformValues(
        Multimaps.index(symbolProblems.entrySet(), entry -> entry.getValue().getJar()).asMap(), 
        ImmutableSetMultimap::copyOf));
```

Something to consider.

suztomo: Trying that approach.

suztomo: Now `indexByJar` returns map of multimap.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -125,13 +129,7 @@ private LinkageChecker(
           }
         });
 
-    return problemToClass.build().inverse();
-  }
-
-  /** Finds linkage errors in the input classpath and generates a linkage check report. */
-  public LinkageCheckReport findLinkageErrors() {
```
suztomo: This removal is the goal of this PR.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -203,8 +203,7 @@ public void testLinkageReports() {
     Nodes reports = details.query("//p[@class='jar-linkage-report']");
     // appengine-api-sdk, shown as first item in linkage errors, has these errors
     Truth.assertThat(trimAndCollapseWhiteSpace(reports.get(0).getValue()))
-        .isEqualTo(
-            "106 target classes causing linkage errors referenced from 516 source classes.");
+        .isEqualTo("115 symbols causing linkage errors referenced from 1,648 classes.");
```
elharo: why did these numbers change?

suztomo: With new value class, it counts linkage errors more uniquely.

A symbol problem caused by CLASS_NOT_FOUND has a ClassSymbol regardless of its origin (class reference, method reference, or field reference). This helps uniquely count symbol problems.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -220,11 +214,12 @@ static Configuration configureFreemarker() {
       Configuration configuration,
       Path output,
       ArtifactCache cache,
-      ImmutableSetMultimap<SymbolProblem, ClassFile> symbolProblems,
+      ImmutableMap<Path, ImmutableSetMultimap<SymbolProblem, String>> symbolProblemTable,
       ListMultimap<Path, DependencyPath> jarToDependencyPaths) {
 
     ImmutableSetMultimap<String, Path> bomMemberToJars = bomMemberToJars(jarToDependencyPaths);
 
+
```
netdpb: Delete extra blank line.

Have you run google-java-format?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -357,7 +358,7 @@ private static ArtifactResults generateArtifactReport(
       results.addResult(TEST_NAME_UPPER_BOUND, upperBoundFailures.size());
       results.addResult(TEST_NAME_GLOBAL_UPPER_BOUND, globalUpperBoundFailures.size());
       results.addResult(TEST_NAME_DEPENDENCY_CONVERGENCE, convergenceIssues.size());
-      results.addResult(TEST_NAME_LINKAGE_CHECK, totalLinkageErrorCount);
+      results.addResult(TEST_NAME_LINKAGE_CHECK, (int) totalLinkageErrorCount);
```
netdpb: Do you really need this explicit cast?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -391,45 +392,31 @@ private static ArtifactResults generateArtifactReport(
   }
 
   /**
-   * Returns a table where rows are jar files in {@code jarFilter}, columns are {@link
-   * SymbolProblem}, and values are class names referencing the problems. When {@code jarFilter} is
-   * {@code null}, jar files are not filtered.
+   * Returns a nested map where first keys are JAR files with symbol problems, second keys are the
```
netdpb: How about

`Partitions {@code symbolProblems} by the JAR file that contains the {@link ClassFile}.`

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -391,45 +392,31 @@ private static ArtifactResults generateArtifactReport(
...
-    });
-
-    return table.cellSet()
-        .stream()
-        .collect(toImmutableTable(
-            Cell::getRowKey,
-            Cell::getColumnKey,
-            cell -> cell.getValue().build()
-        ));
+  private static ImmutableMap<Path, ImmutableSetMultimap<SymbolProblem, String>> indexByJar(
```
netdpb: Small question: Why bother extracting the class file from the values of the multimap? If you return `IM<Path, ISM<SymbolProblem, ClassFile>>` you can use that just as easily, and it takes less transformation.

Feel free to keep doing what you're doing now if you prefer.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -391,45 +392,31 @@ private static ArtifactResults generateArtifactReport(
...
+      ImmutableSetMultimap<SymbolProblem, ClassFile> symbolProblems) {
+
+    ImmutableMap<Path, Collection<Entry<SymbolProblem, ClassFile>>> jarMap = Multimaps
+        .index(
+            symbolProblems.entries(),
+            entry -> entry.getValue().getJar()).asMap();
```
netdpb: Please run google-java-format. It'll make this expression easier to read because it'll be clear that `asMap()` is being called on the result of `Multimaps.index()`.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -391,45 +392,31 @@ private static ArtifactResults generateArtifactReport(
...
+
+    return ImmutableMap.copyOf(
+        Maps.transformValues(
+            jarMap,
+            entries -> entries // These entries have same JAR file for entry.getValue.getJar()
```
netdpb: If you still want to transform the `ClassFile` to a class name, you can do this instead:

```java
ImmutableMap.copyOf(
    Maps.transformValues(
        jarMap,
        entries ->
            ImmutableSetMultimap.copyOf(
                Multimaps.transformValues(
                    ImmutableSetMultimap.copyOf(entries), ClassFile:getClassName))));
```

dashboard/src/main/resources/templates/macros.ftl
```diff
@@ -8,23 +8,25 @@
...
-  <!-- problemsToClassFiles: ImmutableMap<SymbolProblem, Set<String>> -->
-  <#if problemsToClassFiles?size gt 0>
+<#macro formatJarLinkageReport jar problemsWithClass jarToDependencyPaths dependencyPathRootCauses>
+  <!-- problemsToClasses: ImmutableSetMultimap<SymbolProblem, String> converted to
+    ImmutableMap<SymbolProblem, Collection<String>>; Multimaps don't work well with Freemarker -->
+  <#assign problemsToClasses = problemsWithClass.asMap() />
+  <#if problemsToClasses?size gt 0>
     <h3>${jar.getFileName()?html}</h3>
-    <#assign symbolProblemCount = problemsToClassFiles?keys?size />
+    <#assign symbolProblemCount = problemsToClasses?keys?size />
```
netdpb: `problemsToClasses.keys().size()` is the same as `problemsToClasses.size()` now that it's a map instead of a multimap.

dashboard/src/main/resources/templates/macros.ftl
```diff
@@ -8,23 +8,25 @@
...
     <#assign sourceClassCount = 0 />
-    <#list problemsToClassFiles?values as classFiles>
-      <#assign sourceClassCount = sourceClassCount + classFiles?size />
+    <#list problemsToClasses as symbolProblem, sourceClasses>
+      <#assign sourceClassCount = sourceClassCount + sourceClasses?size />
```
netdpb: This won't be the number of unique source classes, since if a source class has more than one problem it'll appear more than once.

Maybe you want `problemsWithClass.inverse().keySet().size()`?

dashboard/src/main/resources/templates/macros.ftl
```diff
@@ -8,23 +8,25 @@
   <#return plural?string(pluralNoun, singularNoun)>
 </#function>
 
-<#macro formatJarLinkageReport jar problemsToClassFiles jarToDependencyPaths dependencyPathRootCauses>
-  <!-- problemsToClassFiles: ImmutableMap<SymbolProblem, Set<String>> -->
-  <#if problemsToClassFiles?size gt 0>
+<#macro formatJarLinkageReport jar problemsWithClass jarToDependencyPaths dependencyPathRootCauses>
+  <!-- problemsToClasses: ImmutableSetMultimap<SymbolProblem, String> converted to
+    ImmutableMap<SymbolProblem, Collection<String>>; Multimaps don't work well with Freemarker -->
```
netdpb: What's the problem with Freemarker? Is it that `<#list multimap as key, value>` doesn't work?

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -203,8 +203,7 @@ public void testLinkageReports() {
     Nodes reports = details.query("//p[@class='jar-linkage-report']");
     // appengine-api-sdk, shown as first item in linkage errors, has these errors
     Truth.assertThat(trimAndCollapseWhiteSpace(reports.get(0).getValue()))
-        .isEqualTo(
-            "106 target classes causing linkage errors referenced from 516 source classes.");
+        .isEqualTo("106 symbols causing linkage errors referenced from 51 classes.");
```
suztomo: The symbol (target class) number remains same. This PR has enhancement to count unique source classes. Previously it was not counting unique number (screenshot in PR description).

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolProblem.java
```diff
@@ -38,7 +37,11 @@
   private final ClassFile containingClass;
 
   public SymbolProblem(Symbol symbol, ErrorType errorType, @Nullable ClassFile containingClass) {
-    this.symbol = checkNotNull(symbol);
+    checkNotNull(symbol);
+
+    // After finding symbol problem, there is no need to have SuperClassSymbol over ClassSymbol.
```
netdpb: What does this change do for you? `SuperClassSymbol` is already a subclass of `ClassSymbol`. Why bother removing that information?

(Also, I'm not sure what `SuperClassSymbol` does that's different from `ClassSymbol`. If there's a problem with a reference to a class as a superclass, then that's a kind of `SymbolProblem`, not a different kind of symbol.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
netdpb: I think that option looks only at changes that haven't been committed yet, compared to the last commit. There really should be a way of looking only at lines that have been changed on this branch, compared to master.

Aha! From [google-java-format.diff.py](https://github.com/google/google-java-format/blob/master/scripts/google-java-format-diff.py):

```shell
$ git diff -U0 master...HEAD | google-java-format-diff.py -p1 -i
```

That should use the diff between (the common ancestor of master and the current branch) and HEAD to determine which lines to format. Note the three dots.

dashboard/src/main/resources/templates/macros.ftl
netdpb: OK, so the comment that "Multimaps don't work well with Freemarker" isn't accurate. Can you remove it?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Thanks. Installed the script but it seems touching other files than the diff. Will show you later.

dashboard/src/main/resources/templates/macros.ftl
suztomo: Updated.

dashboard/src/main/resources/templates/macros.ftl
```diff
@@ -14,11 +14,8 @@
   <#assign problemsToClasses = problemsWithClass.asMap() />
   <#if problemsToClasses?size gt 0>
     <h3>${jar.getFileName()?html}</h3>
-    <#assign symbolProblemCount = problemsToClasses?keys?size />
-    <#assign sourceClassCount = 0 />
-    <#list problemsToClasses as symbolProblem, sourceClasses>
-      <#assign sourceClassCount = sourceClassCount + sourceClasses?size />
-    </#list>
+    <#assign symbolProblemCount = problemsToClasses?size />
```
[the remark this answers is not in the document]
suztomo: Updated. Somehow `problemsWithClass?size` returns 30 (expected 106). So not using this multimap with `?size`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassSymbol.java
```diff
@@ -26,4 +26,6 @@ public ClassSymbol(String className) {
   public String toString() {
     return "Class " + getClassName();
   }
+
```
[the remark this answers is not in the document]
suztomo: Deleted

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -129,7 +130,14 @@ private LinkageChecker(
...
 
-    return problemToClass.build();
+    // Remove duplicate references from inner classes of same Java class file
+    return ImmutableSetMultimap.copyOf(
+        Multimaps.transformValues(
+            problemToClass.build(),
+            classFile ->
+                classFile.getClassName().contains("$")
+                    ? new ClassFile(classFile.getJar(), classFile.getClassName().split("\\$")[0])
+                    : classFile));
```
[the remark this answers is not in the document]
suztomo: Nice. Moved the implementation to ClassFile.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolProblem.java
suztomo: Because of the inequality (`equals` method) between `ClassSymbol` and `SuperClassSymbol`, the dashboard was showing the symbol problem twice for the same (missing) class; one for normal class symbol and the other for class symbol referenced by subclass.


`SuperClassSymbol` is a marker that the reference is for class inheritance. Once the class inheritance validation ([link](https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SuperClassSymbol.java#L20)) is done, these symbols should be treated the same.

dashboard/src/main/resources/templates/macros.ftl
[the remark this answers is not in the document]
suztomo: `problemsWithClass?size` returns 30.
`problemsToClasses?size` returns 106.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: It worked. This is what I needed:

    ~/cloud-opensource-java$ git fetch; git diff -U0 origin/master...HEAD | google-java-format-diff.py -p1 --skip-sorting-imports -i

suztomo:
@netdpb PTAL.

dashboard/src/main/resources/templates/macros.ftl
```diff
@@ -2,54 +2,52 @@
...
 </#function>
-
 <!-- same as above but without the number -->
 <#function plural number singularNoun pluralNoun>
   <#local plural = number gt 1 />
   <#return plural?string(pluralNoun, singularNoun)>
 </#function>
 
-<#macro formatJarLinkageReport jarLinkageReport jarToDependencyPaths dependencyPathRootCauses>
-  <#if jarLinkageReport.getErrorCount() gt 0>
```
suztomo: With this PR, this if statement is always true.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -375,23 +389,45 @@ private static ArtifactResults generateArtifactReport(
...
-    
+
     Map<String, String> latestArtifacts = collectLatestVersions(globalDependencies);
 
     Map<String, Object> templateData = new HashMap<>();
     templateData.put("table", table);
     templateData.put("lastUpdated", LocalDateTime.now());
     templateData.put("latestArtifacts", latestArtifacts);
-    templateData.put("jarLinkageReports", linkageCheckReport.getJarLinkageReports());
+    templateData.put("jarToSymbolProblemToClasses", symbolProblemTable);
```
elharo: The name jarToSymbolProblemToClasses is getting unwieldy. I'm not sure I have a better idea though. Maybe just symbolProblems.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
elharo: So are you claiming the previous 516 number was incorrect?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Updated to symbolProblems. (I thought [Hangarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) is easy to understand in Freemarker template, but `jarToSymbolProblemToClasses` is too much)

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
suztomo: Yes, 516 source classes was not unique. The screenshot in [the PR summary](https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/631#issue-282943824) explains the unique count should be 51 in old (current) dashboard.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -203,8 +203,7 @@ public void testLinkageReports() {
     Nodes reports = details.query("//p[@class='jar-linkage-report']");
     // appengine-api-sdk, shown as first item in linkage errors, has these errors
     Truth.assertThat(trimAndCollapseWhiteSpace(reports.get(0).getValue()))
-        .isEqualTo(
-            "106 target classes causing linkage errors referenced from 516 source classes.");
+        .isEqualTo("106 symbols causing linkage errors on 516 references.");
```
suztomo: @elharo As discussed, the dashboard shows "references" rather than "source classes" for each symbol problem.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -280,18 +279,18 @@ public void testComponent_linkageCheckResult() throws IOException, ParsingExcept
     Nodes reports = document.query("//p[@class='jar-linkage-report']");
     Assert.assertEquals(1, reports.size());
     Truth.assertThat(trimAndCollapseWhiteSpace(reports.get(0).getValue()))
-        .isEqualTo(
-            "106 target classes causing linkage errors referenced from 516 source classes.");
+        .isEqualTo("106 symbols causing linkage errors on 516 references.");
```
elharo: This phrasing still doesn't quite work. In any case, we should make this change in a separate PR, not as part of the refactoring.

dashboard/src/main/resources/templates/macros.ftl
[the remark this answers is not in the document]
netdpb: I thought `?size` was equivalent to `.size()`. I don't know what Freemarker is trying to do then.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/SymbolProblem.java
netdpb: What is the difference between a normal class symbol and a superclass symbol? They're both a class symbol, right? Is the latter just used to report a missing superclass problem? That should be represented as a missing symbol problem, shouldn't it? Just like any other reference to a class.

suztomo: The difference is only to check class inheritance condition when validating reference to `ClassSymbol` including SuperClassSymbol; for example a subclass cannot override `final` methods on superclass.

> That should be represented as a missing symbol problem, shouldn't it?
Yes.

PR #650, opened 2019-06-04
http-client 1.29.2

elharo:
@suztomo @chingor13


chingor13:
FYI, there is a 1.30.0 release coming today. 1.29.2 had a regression fix.

PR #661, opened 2019-06-05
Short lived ClassRepository to use less memory while parsing JAR files

### LINKED ISSUE #660: Short-lived ClassPathRepository
Short-lived ClassPathRepository to avoid the class consuming 400MB while parsing hundreds of JAR files.

Torwards #653 .

suztomo:
Fixes  #660.

- `ClassDumper.listClassesInJar` to use short-lived ClassPathRepository to avoid consuming 400 MB when it parses JAR files for spring-cloud-gcp-kotlin-app-sample.
- `ClassDumper` remembers special location for classes in `BOOT-INF/classes`. `ClassPathRepository` used to remember such location but now we use different instances of the class.


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -74,6 +76,12 @@
 
   private final ImmutableList<Path> inputClassPath;
   private final Repository classRepository;
+
+  // Sometimes classes are not placed in the root of a JAR file. For example, Spring Boot Gradle
+  // Java plugin places class files under "BOOT-INF/classes". ClassDumper needs to remember the
```
elharo: https://docs.spring.io/spring-boot/docs/current/reference/html/executable-jar.html

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
```diff
@@ -313,4 +314,26 @@ public void testIsUnusedClassSymbolReference_classSymbolReferenceNotFound()
...
+    SymbolReferenceMaps symbolReferences = classDumper.findSymbolReferences();
+    symbolReferences
+        .getClassToClassSymbols()
+        .forEach(
+            (classFile, classSymbol) -> {
+              try {
+                classDumper.loadJavaClass(classFile.getClassName());
+              } catch (ClassNotFoundException ex) {
+                fail(
+                    "The class in findSymbolReferences should be able to load again: "
```
elharo: I'm not sure what this message means

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Updated the URL

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
suztomo: Simplified the test and removed the message.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -74,6 +76,12 @@
 
   private final ImmutableList<Path> inputClassPath;
   private final Repository classRepository;
+
+  // Sometimes classes are not placed in the root of a JAR file. For example, Spring Boot Gradle
+  // Java plugin places class files under "BOOT-INF/classes" (
+  // https://docs.spring.io/spring-boot/docs/current/reference/html/executable-jar.html).
+  // ClassDumper needs to remember the special location to load JavaClass by a class name.
+  private final Map<String, String> specialClassFileLocation;
```
elharo: This maps what to what? i.e. what are thw two strings?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -115,7 +124,8 @@ private ClassDumper(
    *     API</a>
    */
   JavaClass loadJavaClass(String className) throws ClassNotFoundException {
-    return classRepository.loadClass(className);
+    String classFileName = specialClassFileLocation.getOrDefault(className, className);
```
elharo: why are the two arguments the same?

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
```diff
@@ -313,4 +314,18 @@ public void testIsUnusedClassSymbolReference_classSymbolReferenceNotFound()
...
+      throws IOException, URISyntaxException, ClassNotFoundException {
+    // This JAR file contains com.google.firestore.v1beta1.FirestoreGrpc under BOOT-INF/classes.
+    // Spring-cloud-gcp's spring-cloud-gcp-kotlin-app-sample module generates such file.
+    Path path = absolutePathOfResource("testdata/dummy-boot-inf-prefix.jar");
+    ClassDumper classDumper = ClassDumper.create(ImmutableList.of(path));
+
+    classDumper.findSymbolReferences(); // This builds specialClassFileLocation
+
+    // This should not raise ClassNotFoundException
+    classDumper.loadJavaClass("com.google.firestore.v1beta1.FirestoreGrpc");
```
elharo: You could assert various tinings about the class returned here.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Added comment.

`getOrDefault(key, defaultValue)` This is shorthand of:

```
if (specialClassFileLocation.containsKey(className)) {
  classFileName = specialClassFileLocation.get(className);
} else {
  classFileName = className;
}
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: Added comment for key and value

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassDumperTest.java
suztomo: Added assertions

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
elharo: These are the names with slashes, right? That is, https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.2.1

Given that I'd call this a pathToClass rather than a class name (since classPath is already taken) and definitely path rather than a simple name.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -359,10 +372,18 @@ Path findClassLocation(String className) {
    */
   private ImmutableSet<JavaClass> listClassesInJar(Path jar) throws IOException {
     ImmutableSet.Builder<JavaClass> javaClasses = ImmutableSet.builder();
+
+    // Not to keep JavaClasses retained by class repository while parsing hundreds of jar files
+    Repository classRepository = createClassRepository(inputClassPath);
     for (String className : listClassNamesInJar(jar)) {
```
elharo: Again, I think you're dealing with a path rather than a name. Distinguishing that and adding some comments should make it clearer what this is doing.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: These are name with "."

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: It's a path separated by "." (usually it's same as name)

suztomo:
I came up with a better idea to create our own ClassPathRepository and keeping the mapping there. Closing this for now.

PR #694, opened 2019-06-11
Linkage Monitor to return 1 if it finds new linkage error

suztomo:
Fixes #681 

- As discussed moved findLinkageProblem from bom/integration-test to `LinkageChecker.create`
- Test with mock object

PR #710, opened 2019-06-12
Follow up of PR 694

### LINKED ISSUE #682: LinkageMonitor: copy the BOM with locally-installed snapshot versions
It seems it can use VersionRangeRequest to query repository.

```
  /**
   * Returns the highest snapshot version installed in {@code repositorySystem}. Null if highest
   * version is not a snapshot.
   */
  private static String findSnapshotVersion(
      RepositorySystem repositorySystem, RepositorySystemSession session, Artifact artifact)
      throws VersionRangeResolutionException {
    Artifact artifactWithVersionRange = artifact.setVersion("(0,]");
    VersionRangeRequest request = new VersionRangeRequest(artifactWithVersionRange, null, null);
    VersionRangeResult versionResult = repositorySystem.resolveVersionRange(session, request);
    String version = versionResult.getHighestVersion().toString();
    if (version.contains("-SNAPSHOT")) {
      return version;
    }
    return null;
  }
```

suztomo:
Fixes #682


linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -90,12 +111,10 @@ static Bom copyWithSnapshot(RepositorySystem repositorySystem, Bom bom)
...
-      Optional<String> snapshotVersion =
-          findSnapshotVersion(repositorySystem, session, managedDependency);
-      if (snapshotVersion.isPresent()) {
-        managedDependency = managedDependency.setVersion(snapshotVersion.get());
-      }
-      managedDependencies.add(managedDependency);
+      managedDependencies.add(
+          findSnapshotVersion(repositorySystem, session, managedDependency)
+              .map(managedDependency::setVersion)
+              .orElse(managedDependency));
```
suztomo: This is from David's comment in #694

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorException.java
```diff
@@ -16,7 +16,7 @@
 
 package com.google.cloud.tools.dependencies.linkagemonitor;
 
-/** Error when {@link LinkageMonitor} reports new linkage errors. */
+/** Exception thrown if {@link LinkageMonitor} reports new linkage errors. */
```
suztomo: This is Elliotte's comment in #694

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
netdpb: Ha! Now I think you're going a bit further than you need to.

`managedDependency.setVersion(foo)` returns its receiver, so this works, but both the `map` and the `orelse` are actually returning the same object, and that is not what I expect when seeing `orElse`.

I recommend

```java
findSnapshotVersion(repositorySystem, session, managedDependency)
    .ifPresent(managedDependency::setVersion);
managedDependencies.add(managedDependency);
````

Alternatively, I guess you could change `findSnapshotVersion` into `applySnapshotVersion`, and have it set the version if it finds it. Then you'd have:

```java
applySnapshotVersion(repositorySystem, session, managedDependency);
managedDependencies.add(managedDependency);
````
Up to you between those.

Comment (suztomo) [reply linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java]: `Artifact` is designed as immutable.

```
/**
 * A specific artifact. In a nutshell, an artifact has identifying coordinates and optionally a file that denotes its
 * data. <em>Note:</em> Artifact instances are supposed to be immutable, e.g. any exposed mutator method returns a new
 * artifact instance and leaves the original instance unchanged. <em>Note:</em> Implementors are strongly advised to
 * inherit from {@link AbstractArtifact} instead of directly implementing this interface.
 * 
 * @noimplement This interface is not intended to be implemented by clients.
 * @noextend This interface is not intended to be extended by clients.
 */
public interface Artifact

```

AbstractArtifact:

```
    /**
     * Creates a new artifact with the specified coordinates, properties and file.
     * 
     * @param version The version of the artifact, may be {@code null}.
     * @param properties The properties of the artifact, may be {@code null} if none. The method may assume immutability
     *            of the supplied map, i.e. need not copy it.
     * @param file The resolved file of the artifact, may be {@code null}.
     * @return The new artifact instance, never {@code null}.
     */
    private Artifact newInstance( String version, Map<String, String> properties, File file )
    {
        return new DefaultArtifact( getGroupId(), getArtifactId(), getClassifier(), getExtension(), version, file,
                                    properties );
    }

    public Artifact setVersion( String version )
    {
        String current = getVersion();
        if ( current.equals( version ) || ( version == null && current.length() <= 0 ) )
        {
            return this;
        }
        return newInstance( version, getProperties(), getFile() );
    }
```

Comment (netdpb) [reply linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java]: Oof. "setVersion" is a bad name for that method. OK.

File: linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java [hunk 3]
```diff
@@ -61,9 +64,22 @@ private void run(String bomCoordinates)
...
         LinkageChecker.create(baseline).findSymbolProblems().keySet();
-
     Bom snapshot = copyWithSnapshot(repositorySystem, baseline);
 
+    // Compare coordinates of the two BOM. No need to run comparison if they are the same.
+    ImmutableList<String> baselineCoordinates =
+        baseline.getManagedDependencies().stream()
+            .map(Artifacts::toCoordinates) // DefaultArtifact does not have equal overridden
+            .collect(toImmutableList());
+    ImmutableList<String> snapshotCoordinates =
```
netdpb: Why are you using a list instead of a set? Do you care if the order differs?

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -76,6 +92,14 @@ private void run(String bomCoordinates)
       throw new LinkageMonitorException(
           String.format("Found %d new linkage error%s", errorSize, errorSize > 1 ? "s" : ""));
     }
+    Set<SymbolProblem> disappearedErrors = Sets.difference(problemsInBaseline, problemsInSnapshot);
```
netdpb: "Fixed" errors?

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -76,6 +92,14 @@ private void run(String bomCoordinates)
...
+    if (!disappearedErrors.isEmpty()) {
+      System.out.println("The following errors disappeared:");
+      System.out.println(disappearedErrors);
+    } else {
+      System.out.println(
+          "Snapshot versions have the same " + problemsInBaseline.size() + " errors as baseline");
```
netdpb: Would it be simpler to do the following?

```
if problemsInSnapshot == problemsInBaseline:
  no change!
  return

if problemsInSnapshot - problemsInBaseline is not empty:
  new errors!
  don't return, so you can also report fixed errors below

if problemsInBaseline - problemsInSnapshot is not empty:
  fixed errors!
```

linkage-monitor/src/test/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorTest.java
```diff
@@ -60,20 +60,31 @@ public void testFindSnapshotVersion() throws VersionRangeResolutionException {
...
             versionScheme.parseVersion("2.1.1"),
             versionScheme.parseVersion("2.1.2-SNAPSHOT")));
+    versionWithoutSnapshot.setVersions(
+        ImmutableList.of(versionScheme.parseVersion("1.2.3"), versionScheme.parseVersion("1.1.1")));
 
     RepositorySystem mockSystem = mock(RepositorySystem.class);
     when(mockSystem.resolveVersionRange(
             any(RepositorySystemSession.class), any(VersionRangeRequest.class)))
-        .thenReturn(dummyVersionRangeResult);
+        .thenReturn(versionWithSnapshot) // first invocation
```
netdpb: This seems brittle because it depends on the order of calls. Are there other options?

boms/cloud-oss-bom/pom.xml
```diff
@@ -45,7 +45,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <guava.version>28.0-android</guava.version>
-    <google.cloud.java.version>0.84.0-alpha</google.cloud.java.version>
+    <google.cloud.java.version>0.89.0-alpha</google.cloud.java.version>
```
suztomo: ???

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtilityTest.java
```diff
@@ -58,7 +58,7 @@ public void testReadBom_path() throws Exception {
     Bom bom = RepositoryUtility.readBom(pomFile);
     
     ImmutableList<Artifact> artifacts = bom.getManagedDependencies();
-    Assert.assertEquals(186, artifacts.size());
+    Assert.assertEquals(203, artifacts.size());
```
suztomo: ??? I've merged origin/master

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -83,7 +83,18 @@ private void run(String bomCoordinates)
     ImmutableSet<SymbolProblem> problemsInSnapshot =
         LinkageChecker.create(snapshot).findSymbolProblems().keySet();
 
+    Set<SymbolProblem> fixedErrors = Sets.difference(problemsInBaseline, problemsInSnapshot);
     Set<SymbolProblem> newErrors = Sets.difference(problemsInSnapshot, problemsInBaseline);
+
+    if (newErrors.equals(fixedErrors)) {
```
netdpb: These cannot be equal unless they're both empty. I think you mean `problemsInSnapshot.equals(problemsInBaseline)`. In that case you can move the `fixedErrors` and `newErrors` declarations to after this `if` block.

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -83,7 +83,18 @@ private void run(String bomCoordinates)
...
+      System.out.println(
+          "Snapshot versions have the same " + problemsInBaseline.size() + " errors as baseline");
+      return;
+    }
+    if (!fixedErrors.isEmpty()) {
+      System.out.println("The following errors are fixed:");
```
netdpb: How about this? It's slightly clearer: "The following errors in the baseline no longer appear in the snapshot:"

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -61,13 +64,37 @@ private void run(String bomCoordinates)
     Bom baseline = RepositoryUtility.readBom(bomCoordinates);
     ImmutableSet<SymbolProblem> problemsInBaseline =
         LinkageChecker.create(baseline).findSymbolProblems().keySet();
-
     Bom snapshot = copyWithSnapshot(repositorySystem, baseline);
 
+    // Compare coordinates of the two BOM. No need to run comparison if they are the same.
+    ImmutableList<String> baselineCoordinates =
+        baseline.getManagedDependencies().stream()
+            .map(Artifacts::toCoordinates) // DefaultArtifact does not have equal overridden
```
elharo: does not override equals

linkage-monitor/src/test/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorTest.java
netdpb: What about something like this?

```java
when(
    mockSystem.resolveVersionRange(
        any(RepositorySystemSession.class),
        argThat(
            request -> request.getArtifact().getArtifactId().equals("something"))))
    .thenReturn(somethingResult);
```

and repeat for each expected artifact.

suztomo: Great. That worked well. Thank you.

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -61,12 +64,36 @@ private void run(String bomCoordinates)
     Bom baseline = RepositoryUtility.readBom(bomCoordinates);
     ImmutableSet<SymbolProblem> problemsInBaseline =
         LinkageChecker.create(baseline).findSymbolProblems().keySet();
-
     Bom snapshot = copyWithSnapshot(repositorySystem, baseline);
 
+    // Compare coordinates of the two BOM. No need to run comparison if they are the same.
```
elharo: BOMs (plural)

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -61,12 +64,36 @@ private void run(String bomCoordinates)
...
+    ImmutableList<String> baselineCoordinates =
+        baseline.getManagedDependencies().stream()
+            .map(Artifacts::toCoordinates) // DefaultArtifact does not override equals
+            .collect(toImmutableList());
+    ImmutableList<String> snapshotCoordinates =
+        snapshot.getManagedDependencies().stream()
+            .map(Artifacts::toCoordinates)
+            .collect(toImmutableList());
+    if (baselineCoordinates.equals(snapshotCoordinates)) {
+      System.out.println("No snapshot version for the BOM found. Not running comparison.");
```
elharo: This doesn't feel quite accurate. The snapshot was found, but the answer is the same as below. This is just an optimization.

linkage-monitor/src/test/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorTest.java
```diff
@@ -60,20 +61,47 @@ public void testFindSnapshotVersion() throws VersionRangeResolutionException {
...
+        snapshotBom.getManagedDependencies().get(0).getVersion());
+
+    int bomSize = bom.getManagedDependencies().size();
+    assertEquals(
+        "Snapshot BOM should have the same length as original BOM.",
+        bomSize,
+        snapshotBom.getManagedDependencies().size());
+    for (int i = 1; i < bomSize; ++i) {
+      assertEquals(
+          "The other artifacts than protobuf-java should have the original version",
```
elharo: "The other artifacts than" --> "Artifacts other than"

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
suztomo: Fixed

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
suztomo: Updated. Does this look good now?

linkage-monitor/src/test/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorTest.java
suztomo: Updated

linkage-monitor/src/test/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorTest.java
```diff
@@ -65,26 +66,42 @@ public void testBomSnapshot()
...
     RepositorySystem mockSystem = mock(RepositorySystem.class);
     when(mockSystem.resolveVersionRange(
             any(RepositorySystemSession.class), any(VersionRangeRequest.class)))
-        .thenReturn(versionWithSnapshot) // first invocation
-        .thenReturn(versionWithoutSnapshot); // subsequent invocation
+        .thenReturn(versionWithoutSnapshot); // other invocations than protobuf-java
+    when(mockSystem.resolveVersionRange(
+            any(RepositorySystemSession.class),
+            argThat(request -> "protobuf-java".equals(request.getArtifact().getArtifactId()))))
+        .thenReturn(versionWithSnapshot); // invocation for protobuf-java
```
netdpb: Maybe rename `versionWithSnapshot` to `protobufJavaVersion`?

PR #711, opened 2019-06-12
Move google-cloud-java ahead 5 versions

elharo:
@suztomo @chingor13


elharo:
We don't need to bump BOM version on every PR, just between releases. Since we have a major version bump scheduled, we don't need to go further than that.

I still need to see how far ahead I can move google-cloud-java before releasing. HEAD doesn't quite work. The next release @chingor13 is working on might work, but even if it doesn't, we might be able to push ahead one or two more versions.

suztomo:
Yes, as long as we remember PRs between releases, that's fine.

Issue #702, opened 2019-06-12
Dashboard index page for different versions


suztomo:
Dashboard index page for different versions. After #697 and #662 is implemented.

The dashboard for a version will be at  `<groupId>/<artifactId>/<versions>/index.html`.

Index page would be at `<groupId>/<artifactId>/index.html`

suztomo:
Idea
`DashboardMain -a com.google.cloud:libraries-bom` to specify version-less artifact. It will

- fetches versions via [VersionRangeRequest](https://maven.apache.org/resolver/apidocs/org/eclipse/aether/resolution/VersionRangeRequest.html)
- runs `DashboardMain.generate(String bomCoodinates)` for each version

PR #738, opened 2019-06-19
DashboardMain to have option to run all versions of BOM

### LINKED ISSUE #737: DashboardMain to have option to run all versions of a BOM
DashboardMain to take a parameter `-a <groupId>:<artifactId>`. For example

```
DashboardMain -f ../bom/pom.xml
DashboardMain -c com.google.cloud:libraries-bom:1.2.2
DashboardMain -a com.google.cloud:libraries-bom  # New!
```

It will

- fetch versions via [VersionRangeRequest](https://maven.apache.org/resolver/apidocs/org/eclipse/aether/resolution/VersionRangeRequest.html)
- run `DashboardMain.generate(String bomCoodinates)` for each version

This does not touch index page: #702 yet.

suztomo:
Fixes #737

I borrowed the term "versionless coordinates" to specify GAV without version:
https://cwiki.apache.org/confluence/display/MAVEN/NoGoalSpecifiedException and https://maven.apache.org/resolver/maven-resolver-api/apidocs/org/eclipse/aether/graph/DependencyCycle.html use the term.

This does not touch index page yet (#702 )


linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -164,36 +161,13 @@ static Bom copyWithSnapshot(RepositorySystem repositorySystem, Bom bom)
...
-        findHighestVersion(
+        RepositoryUtility.findHighestVersion(
             repositorySystem, session, artifact.getGroupId(), artifact.getArtifactId());
     if (version.contains("-SNAPSHOT")) {
       return Optional.of(version);
     }
     return Optional.empty();
   }
-
-  private static String findHighestVersion(
```
suztomo: These functions have been moved to RepositoryUtility

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -98,11 +100,38 @@ public static void main(String[] arguments)
...
+        elements.size() == 2,
+        "The version-less coordinates should have one colon character: " + versionlessCoordinates);
+    String groupId = elements.get(0);
+    String artifactId = elements.get(1);
+
+    RepositorySystem repositorySystem = RepositoryUtility.newRepositorySystem();
+    ImmutableList<String> versions =
+        RepositoryUtility.findVersions(repositorySystem, groupId, artifactId);
+    for (String version : versions) {
+      if (version.contains("SNAPSHOT")) {
```
suztomo: This if statement filters locally-installed snapshot. Kokoro Daily job will run DashboardMain for snapshot separately.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardArguments.java
```diff
@@ -101,6 +122,16 @@ private static Options configureOptions() {
             .build();
     inputGroup.addOption(inputCoordinatesOption);
 
+    Option versionlessCoordinatesOption =
+        Option.builder("a")
+            .longOpt("versionless-coordinates")
```
netdpb: Should this be "all-versions"? Why "a"?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardArguments.java
```diff
@@ -101,6 +122,16 @@ private static Options configureOptions() {
             .build();
     inputGroup.addOption(inputCoordinatesOption);
 
+    Option versionlessCoordinatesOption =
```
netdpb: Instead of a separate option, would it be simpler to allow the version part of the coordinates of -c to be * or some other sentinel value?

-c "com.google.cloud:libraries-bom:*"

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -98,11 +100,38 @@ public static void main(String[] arguments)
       throws IOException, TemplateException, RepositoryException, URISyntaxException,
           ParseException, MavenRepositoryException {
     DashboardArguments dashboardArguments = DashboardArguments.readCommandLine(arguments);
-    Path output =
-        dashboardArguments.hasFile()
-            ? generate(dashboardArguments.getBomFile())
-            : generate(dashboardArguments.getBomCoordinates());
-    System.out.println("Wrote dashboard into " + output.toAbsolutePath());
+
+    if (dashboardArguments.hasVersionlessCoordinates()) {
```
netdpb: Should we report an error if someone passes more than one of (-c, -a, -f)?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -98,11 +100,38 @@ public static void main(String[] arguments)
...
+      generateAllVersions(dashboardArguments.getVersionlessCoordinates());
+    } else {
+      Path output =
+          dashboardArguments.hasFile()
+              ? generate(dashboardArguments.getBomFile())
+              : generate(dashboardArguments.getBomCoordinates());
+      System.out.println("Wrote dashboard into " + output.toAbsolutePath());
```
netdpb: Why not have `generate()` print this? The one with the file could write "Wrote dashboard for _file_ into _output_", and the one with coordinates could write "Wrote dashboard for _coordinates_ into _output_". Then `generateAllVersions` doesn't have to do it too.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtilityTest.java
```diff
@@ -63,4 +64,23 @@ public void testReadBom_path() throws MavenRepositoryException {
...
+
+  @Test
+  public void testFindHighestVersions() throws MavenRepositoryException {
+    RepositorySystem system = RepositoryUtility.newRepositorySystem();
+    String guavaHighestVersion =
+        RepositoryUtility.findHighestVersion(
+            system, RepositoryUtility.newSession(system), "com.google.guava", "guava");
+    Assert.assertNotNull(guavaHighestVersion);
+    // Latest guava release is greater than or equal to 28.0.
+    Truth.assertThat(guavaHighestVersion).isAtLeast("28.0");
```
netdpb: This is doing case sensitive string comparisons. I'm not sure that's what you want.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: commons-cli throws `AlreadySelectedException` when more than one option group is set.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: Nice idea. Updated.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtilityTest.java
suztomo: Good catch; it will fail when guava's "100.0-jre" release. Updated.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardArguments.java
suztomo: a for artifacts. Updated the longOpt to have "all-versions"

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardArguments.java
suztomo: I implemented one with version range https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/739 . Elliotte does not like it.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -104,11 +104,11 @@ public static void main(String[] arguments)
     if (dashboardArguments.hasVersionlessCoordinates()) {
       generateAllVersions(dashboardArguments.getVersionlessCoordinates());
     } else {
-      Path output =
-          dashboardArguments.hasFile()
-              ? generate(dashboardArguments.getBomFile())
-              : generate(dashboardArguments.getBomCoordinates());
-      System.out.println("Wrote dashboard into " + output.toAbsolutePath());
+      if (dashboardArguments.hasFile()) {
```
netdpb: What about using `else if` here? It's simpler.

```java
if (getVersionlessCoordinates()) {
  generate...
} else if (hasFile()) {
  generate...
} else {
  generate...
}
```

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -129,15 +129,16 @@ private static void generateAllVersions(String versionlessCoordinates)
...
+      generate(String.format("%s:%s:%s", groupId, artifactId, version));
     }
   }
 
   @VisibleForTesting
   static Path generate(String bomCoordinates)
       throws IOException, TemplateException, RepositoryException, URISyntaxException {
-    return generate(RepositoryUtility.readBom(bomCoordinates));
+    Path output = generate(RepositoryUtility.readBom(bomCoordinates));
+    System.out.println("Wrote dashboard for " + bomCoordinates + " at " + output);
```
netdpb: Nit-pick: I think "to" is a little better here than "at".

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -146,7 +147,9 @@ static Path generate(Path bomFile)
           MavenRepositoryException {
     checkArgument(Files.isRegularFile(bomFile), "The input BOM %s is not a regular file", bomFile);
     checkArgument(Files.isReadable(bomFile), "The input BOM %s is not readable", bomFile);
-    return generate(RepositoryUtility.readBom(bomFile));
+    Path output = generate(RepositoryUtility.readBom(bomFile));
+    System.out.println("Wrote dashboard for " + bomFile + " at " + output);
```
netdpb: ...and here.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
suztomo: That's more clear. Updated.

PR #767, opened 2019-07-03
Simplify language; omit needless words

elharo:
@garrettjonesgoogle

### DIFF
--- a/library-best-practices/JLBP-6.md
+++ b/library-best-practices/JLBP-6.md
@@ -2,11 +2,10 @@
 -----------------------------------------------
 
 When a library B in Java depends on another library A through the Maven
-repository system, there are two identifiers that library B needs to
-access classes in library A:
+repository system, library B needs two identifiers  to find classes in library A:
 
 1. The Maven coordinates of library A, following the form
-   `${groupId}:${artifactId}:${version}`, for example
+   `${groupId}:${artifactId}:${version}`; for example
    `com.google.guava:guava:26.0-jre`. The Maven coordinates are used to pull the
    library's jar from Maven Central. For each pair of group ID and artifact ID
    (hereafter referenced as "Maven ID"), the user's build system (for example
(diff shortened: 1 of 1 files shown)


library-best-practices/JLBP-6.md
```diff
@@ -2,7 +2,7 @@
 --------------------------------------------
 
 When a library B in Java depends on another library A through the Maven
-repository system, there are two types of handles that library B needs to use to
+repository system, there are two identifiers that library B needs to
```
garrettjonesgoogle: This creates a garden path - "identifiers" fills in the object slot of "access" until "classes" intrudes and bumps it out. Even though it feels more wordy, I think we need either "needs to use to access" or "needs in order to access."

library-best-practices/JLBP-6.md
```diff
@@ -17,9 +17,7 @@ access classes in library A:
    generally share a Java package (the package as defined in `package`
    statements, for example `package com.google.common.collect`). The classpath,
    which is formed from Maven artifacts and possibly non-Maven sources, is
-   searched for each fully-qualified class name at runtime. In order to simplify
-   the discussion, we only talk about the Java package, assuming that the
```
garrettjonesgoogle: I think this clarification was added because one of the readers of the JLBP was confused, and this needed to be clarified.

library-best-practices/JLBP-6.md
elharo: "use" is a weak, non-descript verb that's almost never needed, especially in conjunction with any other verb. And "in order to" is simply a wordy synonym for "to".

library-best-practices/JLBP-6.md
elharo: Except that this sentence is actively incorrect. It says "we only talk about the Java package..." and then immediately proceed to talk about the Maven coordinates too.

If this is trying to say something else, it's unclear and I did not follow it so it should be rephrased.

library-best-practices/JLBP-6.md
garrettjonesgoogle: I don't disagree with what you're saying about "use" and "in order to", but the garden path still needs to be fixed somehow.

library-best-practices/JLBP-6.md
garrettjonesgoogle: Oh, we were trying to say that as far as the fully-qualified class name goes, we only talk about the Java package part of that fully-qualified class name. (The scope of the comment was limited to that.)

library-best-practices/JLBP-6.md
elharo: not sure what you mean by "garden path" here

garrettjonesgoogle: https://en.wikipedia.org/wiki/Garden-path_sentence

Read this in isolation:

"there are two identifiers that library B needs to access"

then read:

"there are two identifiers that library B needs to access classes"

The "classes" feels hanging if you read it this way, because you have to re-interpret the grammar (specifically, what "access" applies to).

elharo: rewritten

library-best-practices/JLBP-6.md
elharo: I don't see that we use that distinction below.

library-best-practices/JLBP-6.md
```diff
@@ -2,11 +2,10 @@
 -----------------------------------------------
 
 When a library B in Java depends on another library A through the Maven
-repository system, there are two identifiers that library B needs to
-access classes in library A:
+repository system, library B needs two identifiers  to find classes in library A:
```
garrettjonesgoogle: two consecutive spaces -> one space

Issue #699, opened 2019-06-12
com.google.guava:guava-jdk5:13.0


elharo:
This is in the BOM tree. WTF?!

Figure out where it's coming from and remove it at the source.

suztomo:
![image](https://user-images.githubusercontent.com/28604/60107743-6c724600-9735-11e9-821c-76484cd08da4.png)

Let me see.

suztomo:
It's this.

- com.google.cloud:google-cloud-datastore:1.73.0
  - com.google.cloud.datastore:datastore-v1-proto-client:1.6.0
    - com.google.api-client:google-api-client:1.20.0
      - com.google.guava:guava-jdk5:13.0`

![image](https://user-images.githubusercontent.com/28604/60109068-af351d80-9737-11e9-94b0-c33f2f637e18.png)

 The latest google-api-client does not depend on guava-jdk5.


datastore-v1-proto-client is available in https://github.com/GoogleCloudPlatform/google-cloud-datastore/blob/master/java/datastore/pom.xml. It's parent pom defines google-api-client version 1.28.0 ([source](https://github.com/GoogleCloudPlatform/google-cloud-datastore/blob/master/java/pom.xml)). But it has not yet released?

elharo:
seem to be fixed

PR #774, opened 2019-07-09
google-cloud-datastore also propagates guava-jdk5

elharo:
@garrettjonesgoogle see #699

### DIFF
--- a/library-best-practices/JLBP-5.md
+++ b/library-best-practices/JLBP-5.md
@@ -23,9 +23,6 @@ Example 1: Guava's main artifact ID is `guava`, but from versions 13.0 to 17.0,
   `guava-jdk5` with `guava` because the artifact names are different.
   When Guava classes overlapped in the two jars, users suffered from 
   runtime errors resulting from classes and methods not being found.
-  The only Google Cloud library propagating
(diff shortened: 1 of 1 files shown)


library-best-practices/JLBP-5.md
```diff
@@ -23,9 +23,6 @@ Example 1: Guava's main artifact ID is `guava`, but from versions 13.0 to 17.0,
   `guava-jdk5` with `guava` because the artifact names are different.
   When Guava classes overlapped in the two jars, users suffered from 
   runtime errors resulting from classes and methods not being found.
-  The only Google Cloud library propagating
```
garrettjonesgoogle: I think the example is helpful; I think we should just remove the incorrect claim that it is the only case.

elharo: I though that's what I did.

garrettjonesgoogle: Well, I meant we should keep the example of the example (google-api-client depending on guava-jdp5)

elharo: I don't think that helps the reader very much. It's quite specific and the general description with guava-jdk5 is clear enough.

PR #782, opened 2019-07-11
Enforcer rule to work with "bundle" packaging

suztomo:
Towards #781 

Jaxen uses [packaging: bundle.](https://github.com/jaxen-xpath/jaxen/blob/master/core/pom.xml#L15) The previous PR skipped such packaging.

Issue #781, opened 2019-07-11
Enforcer rule to work for jaxen project


suztomo:
Try running the enforcer rule in https://github.com/jaxen-xpath/jaxen and see how it works or not.

suztomo:
After #782 and #783 Enforcer rule detected following linkage errors in jaxen/dom4j-navigator:

```
[INFO] --- maven-enforcer-plugin:3.0.0-M2:enforce (enforce) @ dom4j-navigator ---
[ERROR] Linkage Checker rule found 7 errors. Linkage error report:
Class org.jaxen.pattern.PatternParser is not found;
  referenced by 1 class file
    org.dom4j.xpath.XPathPattern (dom4j-1.6.1.jar)
(dom4j-navigator-2.0.0-BETA.jar) org.jaxen.pattern.Pattern's method getText() is not found;
  referenced by 1 class file
    org.dom4j.xpath.XPathPattern (dom4j-1.6.1.jar)
(dom4j-navigator-2.0.0-BETA.jar) org.jaxen.pattern.Pattern's method matches(Object arg1, org.jaxen.Context arg2) is not found;
  referenced by 1 class file
    org.dom4j.xpath.XPathPattern (dom4j-1.6.1.jar)
(dom4j-navigator-2.0.0-BETA.jar) org.jaxen.pattern.Pattern's method getPriority() is not found;
  referenced by 1 class file
    org.dom4j.xpath.XPathPattern (dom4j-1.6.1.jar)
(dom4j-navigator-2.0.0-BETA.jar) org.jaxen.pattern.Pattern's method getUnionPatterns() is not found;
  referenced by 1 class file
    org.dom4j.xpath.XPathPattern (dom4j-1.6.1.jar)
(dom4j-navigator-2.0.0-BETA.jar) org.jaxen.pattern.Pattern's method getMatchType() is not found;
  referenced by 1 class file
    org.dom4j.xpath.XPathPattern (dom4j-1.6.1.jar)
(dom4j-navigator-2.0.0-BETA.jar) org.jaxen.pattern.Pattern's method getMatchesNodeName() is not found;
  referenced by 1 class file
    org.dom4j.xpath.XPathPattern (dom4j-1.6.1.jar)

[WARNING] Rule 0: com.google.cloud.tools.dependencies.enforcer.LinkageCheckerRule failed with message:
Failed while checking class path. See above error report.

```

I will check the validity.

elharo:
[...]

PR #783, opened 2019-07-11
Enforcer rule not to count test scoped dependencies

suztomo:
Towards #781 .

After applying #782 locally, Jaxen project still does not work with the enforcer rule because of the following IndexOutOfBoundsException:

```
Caused by: java.lang.IndexOutOfBoundsException: end index (2) must not be greater than size (1)
    at com.google.common.base.Preconditions.checkPositionIndexes (Preconditions.java:1418)
    at com.google.common.collect.SingletonImmutableList.subList (SingletonImmutableList.java:64)
    at com.google.cloud.tools.dependencies.enforcer.LinkageCheckerRule.execute (LinkageCheckerRule.java:137)
    at org.apache.maven.plugins.enforcer.EnforceMojo.execute (EnforceMojo.java:194)
```

This is because the jaxen has junit in dependency (inherited from parent), while the enforcer rule does not include test-scoped dependencies.

The enforcer rule was trying to create sublist of classpath (which only contains jaxen.jar) with length 2 (jaxen and junit)

### DIFF
--- a/enforcer-rules/src/test/java/com/google/cloud/tools/dependencies/enforcer/LinkageCheckerRuleTest.java
+++ b/enforcer-rules/src/test/java/com/google/cloud/tools/dependencies/enforcer/LinkageCheckerRuleTest.java
@@ -53,6 +53,7 @@ import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.artifact.DefaultArtifact;
 import org.eclipse.aether.collection.CollectRequest;
+import org.eclipse.aether.graph.DefaultDependencyNode;
 import org.eclipse.aether.graph.Dependency;
 import org.eclipse.aether.graph.DependencyNode;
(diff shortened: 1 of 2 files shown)


enforcer-rules/src/test/java/com/google/cloud/tools/dependencies/enforcer/LinkageCheckerRuleTest.java
```diff
@@ -114,15 +119,13 @@ private void setupMock()
...
   }
 
-  /** Returns a dependency graph node resolved from {@link Artifact} of {@code coordinates}. */
+  /**
+   * Returns a dependency graph node resolved from {@link Artifact} of {@code coordinates}.
+   */
   private DependencyNode createResolvedDependencyGraph(String... coordinates)
       throws RepositoryException, URISyntaxException {
     CollectRequest collectRequest = new CollectRequest();
-    // This dummy artifact must be something that exists in a repository
```
suztomo: This variable has been moved to an instance variable

PR #798, opened 2019-07-23
Linkage Checker to record first JAR file when multiple JARs contain same class

### LINKED ISSUE #797: Linkage Checker to record first JAR file when multiple JARs contain same class
Problem found while investigating #794 . Both `guide-3.0.jar` and `sisu-guice-3.2.6-no_aop.jar` contain `com.google.inject.internal.InjectorImpl$BindingsMultimap`. Because `guide-3.0.jar` supersedes in the class path, `com.google.inject.internal.InjectorImpl$BindingsMultimap` should not be associated with `sisu-guice-3.2.6-no_aop.jar`.

suztomo:
Fixes #797 .


suztomo:
Thanks

PR #808, opened 2019-07-26
Remove DependencyGraph.levelOrder in favor of Null DependencyGraphTransformer

### LINKED ISSUE #806: DependencyGraphBuilder.levelOrder can be replaced with DependencyGraphTransformer?
DependencyGraphBuilder.levelOrder is complex and not following how maven handles exclusions. Can the function be replaced with DependencyGraphTransformer attached to RepositorySystemSession?

image

Key differences between Maven and DependencyGraphBuilder

- DependencyGraphBuilder.levelOrder gives dependency tree that has duplicates
- DependencyGraphBuilder.levelOrder returns our DependencyGraph, not aether's DependencyNode. The former contains list of `DependencyPath` (our class) to represent tree structure.

DependencyGraphTransformer takes a node that has not yet pruned duplicates.

image

The default ConflictResolver has 
- versionSelector (no need for us)
- scopeSelector (no need for us. Maybe only for test)
- optionalitySelector (include optional for now)
- scopeDeriver (JavaScopeDeriver; no need for us)

# Usage of DependencyGraphBuilder

## getDirectDependencies
getDirectDependencies is only used by DirectReport.java which is unused for dashboard or linkage checker.

## getCompleteDependencies and getTransitiveDependencies

How are they different? getCompleteDependencies specified levelOrder(FULL_DEPENDENCY), meaning to call resolveCompileTimeDependencies for each node in the tree.

getTransitiveDependencies does not try to resolve dependency in each node.

```
  private static ArtifactCache loadArtifactInfo(List<Artifact> artifacts) {
    Map<Artifact, ArtifactInfo> infoMap = new LinkedHashMap<>();
    List<DependencyGraph> globalDependencies = new ArrayList<>();

      try {
            DependencyGraphBuilder.getCompleteDependencies(artifact);
        globalDependencies.add(completeDependencies);

        // picks versions according to Maven rules
        DependencyGraph transitiveDependencies =
            DependencyGraphBuilder.getTransitiveDependencies(artifact);`
```

# includeProvidedScope?

Q: What's the purpose of the flag?

A: When I wanted to use DependencyGraphBuilder for LinkageChecker, there was a problem
(https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/187#issuecomment-431921964; ). The existing report for each artifact, which uses `getTransitiveDependencies`, did not include provided-scoped artifacts, while Linkage Checker needed to have them.

Initially I tried to change the behavior to include provided ([commit](https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/5b045c3dac293b941e8a2ec28b8fa51a8b17bf3a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java#L85)), but I ended up creating the flag includeProvidedScope to select configuration for RepositorySystemSession.

Q. Given levelOrder(FULL_DEPENDENCY) calls `resolveCompileTimeDependencies` for each node, is there are difference between levelOrder(FULL_DEPENDENCY)  and levelOrder(FULL_DEPENDENCY_WITH_PROVIDED)? 

A: resolveCompileTimeDependencies() does not include provided dependencies. Therefore linkage checker with FULL_DEPENDENCY outputs false positives.

By default, RepositorySystemSession does not include provided scope:

image

By default, RepositorySystemSession includes optional direct dependencies:

image

It omits optional non-direct dependencies:

image

Depenency convergence seems incorrect when using NoopGraphTransformer:

image

NoopGraphTransformer receives a tree that already have managed dependencies. ClassicDependencyManager sets the versions. Its `managedVersion` map is passed around the ClassicDependencyManager instances.

suztomo:
Fixes #806 

To discuss the idea with Elliotte before fixing tests.


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
```diff
@@ -58,11 +62,50 @@
   
   // map of groupId:artifactId:version to paths
   private SetMultimap<String, DependencyPath> paths = HashMultimap.create();
-  
+
   @VisibleForTesting
-  public DependencyGraph() {
+  public DependencyGraph() {}
```
elharo: I'm surprised this is needed. Could we add an empty root instead/?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
```diff
@@ -58,11 +62,50 @@
...
+
+      DependencyPath path = new DependencyPath();
+      item.forEach(
+          element -> {
+            if (element.getArtifact() != null) {
+              path.add(element.getArtifact());
+            }
+          });
+      if (path.size() > 0) {
+        // Root node may be empty
```
elharo: may be or is?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -157,20 +157,22 @@ public DependencySelector deriveChildSelector(
...
+        includeProvidedScope ? new String[] {"test"} : new String[] {"provided", "test"};
+
     DependencySelector dependencySelector =
         new AndDependencySelector(
             // ScopeDependencySelector takes exclusions. 'Provided' scope is not here to avoid
             // false positive in LinkageChecker.
-            new ScopeDependencySelector("test"),
-            new OptionalDependencySelector(),
+            new ScopeDependencySelector(excludedScopes),
             new ExclusionDependencySelector(),
```
suztomo: This is configuration to respect exclusion elements.

suztomo:
I'll need to figure out the difference in tree generation between master and this branch. Currently at `DependencyGraphIntegrationTest.testJaxen` shows unexplained difference.

suztomo:
Jaxen's dependency tree by DependencyGraphBuilder.levelOrder cannot be replicated by Null DependencyGraphTransformer. DefaultDependencyCollector handles cyclic dependency differently than DependencyGraphBuilder. This makes the resulting tree different when dealing with Jaxen's dependency than original.

------

When a cycle node is found (key: groupId and artifactId), it adds the new node node to the tree but tits children are the one already in the tree.

For example in our case, jaxen:jaxen:1.1.6 > dom4j:dom4j:1.6.1 is in the stack. dom4j:dom4j:1.6.1 has a dependency jaxen:jaxen:jar:1.1-beta-6. This is a cycle node (key: jaxen:jaxen). DefaultDependencyCollector adds jaxen:jaxen:jar:1.1-beta-6 but the children of the node is the same as jaxen:jaxen:1.1.6's: dom4j:dom4j:1.6.1. That's why the tree does not get dom4j:dom4j:1.5.2.

elharo:
That's weird. Which behavior (if either) is the one we need?

PR #812, opened 2019-07-29
With Java 9 it's not sufficient to advise the developer to pay attention to where the class is coming from.

elharo:
@garrettjonesgoogle @netdpb 

While working on JLBP-19 I realized that overlapping classes imply overlapping packages, which is a compile time error in Java 9+. It's no longer sufficient to advise the developer to pay attention to where the class is coming from.

### DIFF
--- a/library-best-practices/JLBP-5.md
+++ b/library-best-practices/JLBP-5.md
@@ -36,8 +36,20 @@ Example 2: There are multiple artifacts that provide classes under
     might be used anywhere that is not a servlet. Define these status codes
     yourself or choose a different library to provide them.
 
+In Java 9 and later overlapping classes become compile-time errors when
+named modules are used.
+
 Problems caused by overlapping classes can be extremely difficult to resolve.
-If at all possible, eliminate all but one of the
-overlapping dependencies. Otherwise pay very close attention to which version of
-each overlapping class is chosen. Make sure the project does not depend on any
-behavior or API of the class that is not selected.
+Nonetheless it is critical, especially in Java 9 and later, to eliminate all but one of the
+overlapping dependencies. 
+
+If possible, upgrade all dependencies on the overlapping classes
+to import them only by the same group ID and artifact ID. 
+
+If this isn't possible, for instance because the dependency that imports the renamed
+artifact is unmaintained, then add
+[dependency exclusions](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html) for all but one of the renamed artifacts
+in your own pom.xml.
+
+
+
--- a/library-best-practices/JLBP-5.md
+++ b/library-best-practices/JLBP-5.md
@@ -36,20 +36,14 @@ Example 2: There are multiple artifacts that provide classes under
     might be used anywhere that is not a servlet. Define these status codes
     yourself or choose a different library to provide them.
 
-In Java 9 and later overlapping classes become compile-time errors when
-named modules are used.
-
-Problems caused by overlapping classes can be extremely difficult to resolve.
-Nonetheless it is critical, especially in Java 9 and later, to eliminate all but one of the
-overlapping dependencies. 
+In Java 9 and later overlapping classes become compile-time and runtime errors when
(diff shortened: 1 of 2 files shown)


library-best-practices/JLBP-6.md
```diff
@@ -18,7 +18,7 @@ repository system, library B needs two identifiers to find classes in library A:
    which is formed from Maven artifacts and possibly non-Maven sources, is
    searched for each fully-qualified class name at runtime.
 
-When breaking changes are introduced to library A between major version 1 and
+When breaking changes are introduced to library between major version 1 and
```
netdpb: I think this "A" was referring to the references in the above paragraphs.

library-best-practices/JLBP-5.md
```diff
@@ -36,8 +36,14 @@ Example 2: There are multiple artifacts that provide classes under
...
-Problems caused by overlapping classes can be extremely difficult to resolve.
-If at all possible, eliminate all but one of the
-overlapping dependencies. Otherwise pay very close attention to which version of
-each overlapping class is chosen. Make sure the project does not depend on any
-behavior or API of the class that is not selected.
+In Java 9 and later overlapping classes become compile-time and runtime errors when
+named modules are used. It is critical, especially in Java 9 and later,
+to remove all but one of the overlapping classes from the classpath.
+
+If possible, upgrade all dependencies on the overlapping classes
```
garrettjonesgoogle: The wording "upgrade all dependencies on the overlapping classes" sounds odd and requires a re-interpretation in the middle of the sentence. I would suggest: "upgrade all dependency declarations on the artifacts containing the overlapping classes".

library-best-practices/JLBP-5.md
```diff
@@ -36,8 +36,14 @@ Example 2: There are multiple artifacts that provide classes under
...
+to import them only by the same group ID and artifact ID.
+
+If this isn't possible, for instance because a dependency that imports the
+renamed artifact is unmaintained, then add
```
garrettjonesgoogle: "the renamed artifact" doesn't seem to have an antecedent. Also, "dependency" here is confusing because it's a different part of the dependency problem than the part having overlapping classes, unlike the previous paragraph.

I would suggest: "... for instance because you depend on an unmaintained library that imports an undesired artifact containing overlapping classes, ..."

library-best-practices/JLBP-5.md
```diff
@@ -36,8 +36,14 @@ Example 2: There are multiple artifacts that provide classes under
...
+[dependency exclusions](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html) for all but one of the renamed artifacts
```
garrettjonesgoogle: I would suggest "undesired" instead of "renamed"

(It's entirely possible that a "renamed" artifact becomes the desired one)

library-best-practices/JLBP-6.md
```diff
@@ -18,14 +18,15 @@ repository system, library B needs two identifiers to find classes in library A:
    which is formed from Maven artifacts and possibly non-Maven sources, is
    searched for each fully-qualified class name at runtime.
 
-When breaking changes are introduced to library A between major version 1 and
+When breaking changes are introduced to a library between major version 1 and
```
garrettjonesgoogle: "library A" was an intentional example name introduced up above.

garrettjonesgoogle:
Can you change the PR description to the change you are making instead of some statement about the world?

library-best-practices/JLBP-5.md
elharo: rewritten

library-best-practices/JLBP-5.md
elharo: rewritten

library-best-practices/JLBP-5.md
elharo: done

library-best-practices/JLBP-6.md
elharo: The principle applies to all libraries equally, B included, not to just library A

PR #828, opened 2019-08-06
update protobuf version

elharo:
@suztomo


elharo:
It does, but this version includes other fixes.

PR #835, opened 2019-08-08
JLBP for adding module names

### LINKED ISSUE #832: JLBP on minimal module info
That is, module name

elharo:
@netdpb @garrettjonesgoogle 
fixes #832 

First draft. Also see https://github.com/googleapis/google-http-java-client/issues/782

### DIFF
--- a/library-best-practices/JLBP-20.md
+++ b/library-best-practices/JLBP-20.md
@@ -1,0 +1,48 @@
+[JLBP-19] Give each jar a module name
+--------------------------------------------------
+
+For compatibility with the Java Platform Module System (JPMS) in Java 9 and
+later, every JAR you publish should have a module name, even if the library
+does not itself use modules. More precisely, the JAR manifest in 
+META-INF/MANIFEST.MF should have an Automatic-Module-Name field such as
+this one for `com.google.http-client:google-http-client`:
+
+```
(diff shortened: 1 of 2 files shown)


library-best-practices/JLBP-20.md
```diff
@@ -0,0 +1,55 @@
...
+later, every JAR you publish should have a module name, even if the library
+does not itself use modules. More precisely, the JAR manifest in 
+META-INF/MANIFEST.MF should have an Automatic-Module-Name field such as
+this one for `com.google.http-client:google-http-client`:
+
+```
+Automatic-Module-Name: com.google.api.client
+```
+
+The module name should be globally unique and composed of
```
netdpb: Might want to link to Stephen Colebourne's [discussion of module naming best practices](https://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html), and even recommend more strongly that we follow them (Replace "often has the same name" with "should have the same name", etc.).

elharo: changed to "should" as suggested

@garrettjonesgoogle What do you think about linking to othe non-normative references?

garrettjonesgoogle: @elharo I don't have a fundamental problem with it.

Issue #834, opened 2019-08-07
Enforcer rule: Paths to the missing artifact: (empty)


suztomo:
The following output is from Ray's [grpc-java-by-example/chat-example/chat-vaadin-client](https://github.com/saturnism/grpc-java-by-example/tree/master/chat-example/chat-vaadin-client). Investigate why the message does not have path.

```
[ERROR] Could not find artifact org.jboss.logging:jboss-logging:pom:3.0.0.Beta5-SNAPSHOT
[ERROR] Paths to the missing artifact: 
[WARNING] Rule 0: com.google.cloud.tools.dependencies.enforcer.LinkageCheckerRule warned with message:
Unable to build a dependency graph: Could not resolve dependencies for project org.test:chat-vaadin-client:war:1.0-SNAPSHOT: Failed to collect dependencies at
 com.vaadin:vaadin-server:jar:8.1.6
 -> com.vaadin:vaadin-push:jar:8.1.6
 -> com.vaadin.external.atmosphere:atmosphere-runtime:jar:2.4.11.vaadin2
 -> org.atmosphere.jboss.as:jboss-as-websockets:jar:0.5
 -> org.jboss.as:jboss-as-server:jar:7.1.1.Final
 -> org.jboss.as:jboss-as-controller:jar:7.1.1.Final
 -> org.jboss.msc:jboss-msc:jar:1.0.2.GA
 -> org.jboss.logging:jboss-logging-processor:jar:1.0.0.Beta5
 -> org.jboss.logging:jboss-logging:jar:3.0.0.Beta5-SNAPSHOT
```

Among the lines, the bug is the following line:

```
[ERROR] Paths to the missing artifact: 
```

The dependency graph has been modified when it is passed to ArtifactTransferException. The graph only has org.jboss.logging:jboss-logging-processor:jar:**1.0.0.Final**.

```
[ERROR] Path to jboss logging:org.test:chat-vaadin-client:war:1.0-SNAPSHOT > com.vaadin:vaadin-push:jar:8.1.6 (compile) > com.vaadin.external.atmosphere:atmosphere-runtime:jar:2.4.11.vaadin2 (compile) > org.atmosphere.jboss.as:jboss-as-websockets:jar:0.5 (provided) > org.jboss.as:jboss-as-server:jar:7.1.1.Final (provided) > org.jboss.logging:jboss-logging-processor:jar:1.0.0.Final (provided)
```

[...]

PR #843, opened 2019-08-19
To use PathRecordingDependencyVisitor instead of our own findArtifact

suztomo:
Towards #834 .

Refactoring to remove `LinkageCheckerRule.findArtifact` in favor of Maven's PathRecordingDependencyVisitor.

Maven-resolver-util provides [PathRecordingDependencyVisitor](https://maven.apache.org/resolver/apidocs/org/eclipse/aether/util/graph/visitor/PathRecordingDependencyVisitor.html) to do the exact same job as `LinkageCheckerRule.findArtifact` to find dependency path in a dependency graph.


suztomo:
Extracted the lambda as variable `filter`. More concise than an unnamed class.

PR #857, opened 2019-08-29
findHighestVersion to return second highest snapshot

### LINKED ISSUE #856: Linkage Monitor to use second highest snapshot version
Protobuf repository keeps the latest released version in its master without incrementing
its version. Maven's `RepositorySystem.resolveVersionRange` considers version "X.Y.Z" is higher than "X.Y.Z-SNAPSHOT". For example "3.9.1" is considered as higher than "3.9.1-SNAPSHOT". This ordering makes Linkage Monitor not picking up latest local build marked as "SNAPSHOT".

Therefore, `RepositoryUtility.findHighestVersion` needs to take this into
account to pick up the version from master branch ("X.Y.Z-SNAPSHOT" over "X.Y.Z")

suztomo:
Fixes #856 .

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -114,6 +114,11 @@
       <artifactId>guava-testlib</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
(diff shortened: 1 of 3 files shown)


suztomo:
Makes sense. I'll ask protobuf team first.

PR #859, opened 2019-08-29
minimal docs directory

elharo:
@garrettjonesgoogle @suztomo For initial testing and experimentation purposes I'm creating a simple docs directory so we can turn on github pages.


elharo:
Yes, and for the moment I don't really care. I just need to put something here so I can test out github pages.

PR #861, opened 2019-08-30
Move best practices onto google pages

elharo:
@garrettjonesgoogle https://googlecloudplatform.github.io/cloud-opensource-java/


elharo:
I don't think I can use a 302 redirect but I can leave a pointer page behind.

PR #870, opened 2019-09-03
Use a custom theme to give us total control over HTML generation

elharo:
@suztomo Still experimenting, but this gives us an even simpler page without any extras added by the jekyll-primer-theme.

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -1,0 +1,24 @@
+<!DOCTYPE html>
+<html lang="{{ site.lang | default: "en-US" }}">
+  <head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+{% seo %}
+  </head>
+  <body>
+    {{ content }}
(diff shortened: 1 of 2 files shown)


docs/_layouts/default.html
```diff
@@ -0,0 +1,24 @@
...
+<html lang="{{ site.lang | default: "en-US" }}">
+  <head>
+    <meta charset="UTF-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+{% seo %}
+  </head>
+  <body>
+    {{ content }}
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/anchor-js/4.1.0/anchor.min.js" integrity="sha256-lZaRhKri35AyJSypXXs4o6OPFTbTmUoltBbDCbdzegg=" crossorigin="anonymous"></script>
```
suztomo: Hmm, depending on Cloudflare... 
I assume you don’t like this.

elharo: We can take this out if we like. We were doing this already. I stared by copying and shrinking the default theme.

PR #875, opened 2019-09-04
Add Copyright

elharo:
@suztomo

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -1,9 +1,9 @@
 <!DOCTYPE html>
-<html lang="{{ site.lang | default: "en-US" }}">
+<html lang="en-US">
   <head>
-    <meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


docs/_layouts/default.html
```diff
@@ -1,9 +1,9 @@
 <!DOCTYPE html>
-<html lang="{{ site.lang | default: "en-US" }}">
+<html lang="en-US">
   <head>
-    <meta charset="UTF-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <meta charset="UTF-8" />
```
suztomo: > do not close void elements, i.e. write <br>, not <br />

https://google.github.io/styleguide/htmlcssguide.html

Meta is a void element. https://www.w3.org/TR/2011/WD-html-markup-20110405/syntax.html#syntax-elements

PR #883, opened 2019-09-05
use at least one paragraph on each JLBP, not just lists

elharo:
@garrettjonesgoogle I also added blank lines between list items to make Kramdown (part of Jekyll/github pages include p elements inside its li elements.) That is, this is going to turn a lot of

```
<ul>
<li>Lorem ipsum</li>
</ul>
```

into 

```
<ul>
<li><p>Lorem ipsum</p></li>
</ul>
```

### DIFF
--- a/docs/JLBP-1.md
+++ b/docs/JLBP-1.md
@@ -25,13 +25,15 @@ Some specific notes about minimizing dependencies:
   dependencies are pulled in, consider a different direct dependency.
   Alternatively, if the functionality you need is small, reimplement
   it in your own library.
+
   - Maven: Run `mvn dependency:tree` (after running
     `mvn install -DskipTests` to build the library).
+    
   - Gradle: Run `./gradlew dependencies`
 
 - Prefer JDK classes where available. For example, XOM and JDOM
-  are very convenient and far easier to use than DOM. However, most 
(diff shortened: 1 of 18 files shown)


garrettjonesgoogle:
This... doesn't seem standard. Why does it need to be done?

elharo:
This adds whitespace between list items. Otherwise they appear cramped and less readable. It also makes list items that contain more than one paragraph consistent with those that contain only a single paragraph.

garrettjonesgoogle:
@elharo it's something I've only see you do. @suztomo What are your thoughts?

suztomo:
CSS should control the visual margin between list items, not Markdown. Therefore we should not add unnecessary spaces between list items in Markdown.

elharo:
See https://googlecloudplatform.github.io/cloud-opensource-java/JLBP-2.html which is pre this PR. Having only a list on the page feels much stranger in rendered HTML than in markdown.

Re, paragraphs in list items this is allowed and accepted in HTML. If each list item is a paragraph, generally a complete sentence or more, then I expect `<p>` elements to be used. If, on the other hand, each list item is just a word or two, then I don't expect `<p>`.

In our case there are a few places where we are already using two or more paragraphs in a single list item. That suggests the rest of these should be paragraphs too.

PR #892, opened 2019-09-06
Update to gRPC 1.23.0

elharo:


suztomo:
@elharo boms/upper-bound-check/pom.xml needs to update the version of new BOM.

PR #896, opened 2019-09-06
Linkage Checker not to read classes from multi-release JAR

### LINKED ISSUE #895: Linkage Checker to exclude classes in multi-release JAR
Linkage Checker to exclude classes in multi-release JAR

#890 's error `com.google.common.base.VerifyException: When checking a class reference from org.graalvm.libgraal.LibGraal to jdk.vm.ci.services.Services, the reference to the target class is no longer found in the source class's constant pool.` is caused by ClassRepository's incorrect handling of  classes in multi-release JAR. Classes in `META-INF.versions.9` or `META-INF.versions.11` have the same name as normal classes and thus the cache in `ClassRepository` was not working as expected; it returned different instances than the one used in extracting symbol references.

Currently Linkage Checker supports Java 8 or earlier (as documented in [isCompatibleClassFileVersion](https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java#L169)). And thus there's no need to read classes in `META-INF.versions.`.

suztomo:
Fixes #895.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -381,6 +381,7 @@ class ClassDumper {
     for (String classFileName : listClassFileNames(jar)) {
       if (classFileName.startsWith("META-INF.versions.")) {
         // Linkage Checker does not support multi-release JAR (for Java 9+) yet
+        // https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/897
         continue;
       }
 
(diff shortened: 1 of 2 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -379,6 +379,11 @@ Path findClassLocation(String className) {
   private ImmutableSet<JavaClass> listClasses(Path jar) throws IOException {
     ImmutableSet.Builder<JavaClass> javaClasses = ImmutableSet.builder();
     for (String classFileName : listClassFileNames(jar)) {
+      if (classFileName.startsWith("META-INF.versions.")) {
+        // Linkage Checker does not support multi-release JAR (for Java 9+) yet
```
elharo: add a link to the issue to support Java 9+ here

suztomo: Added issue URL one line below.

PR #901, opened 2019-09-12
use paragraphs for top level items, not lists JLBP-2

elharo:
@garrettjonesgoogle as discussed in meeting


elharo:
Look at https://jlbp.dev/JLBP-2.html for the current HTML version. This one looks a little strange to me.  What do you think?

garrettjonesgoogle:
The current HTML version seems fine (ish)...

elharo:
Again https://g3doc.corp.google.com/company/engedu/writing/lists-and-tables.md?cl=head would suggest using paragraphs here. We shouldn't be opening with a list item.

garrettjonesgoogle:
It just suggests opening with at least an introductory sentence that isn't a list item. That could be added before the existing list.

PR #907, opened 2019-09-12
convert to paragraphs

elharo:
@garrettjonesgoogle


elharo:
Can we compromise in the middle with two paragraphs? I rewrote this some since I was finding "ecosystem" and "snapshots" to be unclear.

garrettjonesgoogle:
I don't see the rewrite yet...

PR #911, opened 2019-09-12
Specify exactly one version of each dependency

### LINKED ISSUE #1002: Recommend against single version hard dependencies in JLBP

elharo:
@suztomo @garrettjonesgoogle fixes #1002


elharo:
Done

garrettjonesgoogle:
Hmm... by "adjusting the paragraphs" I meant to imply that they would be combined into larger paragraphs. The short paragraphs have a choppy feel.

elharo:
Sounds good. I combined several paragraphs.

garrettjonesgoogle:
I don't see any updates to the content since your last message...

elharo:
Please check again. I see updates.

garrettjonesgoogle:
Timeline:

- Sep 12: your commit "use paragraphs for top level items, not lists"
- Sep 13: my comment "I would actually recommend removing all bullet points on this one, and adjusting the paragraphs so it flows better."
- Sep 25: your commit "rewrite without lists"
- Sep 26: my comment "Hmm... by "adjusting the paragraphs" I meant to imply that they would be combined into larger paragraphs. The short paragraphs have a choppy feel."

All other commits are merge commits. There is no non-merge commit since my Sep 26 comment.

elharo:
OK, I think we're getting out of sync here. It would help if you could make specific comments on particular lines so I can see what you''re asking for.

docs/JLBP-14.md
```diff
@@ -1,29 +1,29 @@
 # [JLBP-14] Do not use version ranges
 
-- If the pom.xml of your library specifies a version range instead of a
-  single version for any particular dependency, builds at different points
-  in time can see different versions of that dependency (as your dependency
-  releases new versions), which can break your library unexpectedly.
+Always specify a single version instead of a version range for dependencies.
```
garrettjonesgoogle: I would combine this paragraph with the next one.

docs/JLBP-14.md
```diff
@@ -1,29 +1,29 @@
...
-    times can produce different artifacts with arbitrarily different behavior.
+When a pom.xml specifies a version range instead of a
+single version for any particular dependency, builds at different points
+in time can see different versions of that dependency (as your dependency
+releases new versions), which can break your product unexpectedly.
 
-  - Version ranges can pull in incomplete releases that break your
-    library's build (example:
-    https://github.com/GoogleCloudPlatform/appengine-gcs-client/issues/71).
+Version ranges cause builds to be non-reproducible. Builds of the
```
garrettjonesgoogle: I would combine this paragraph with the next two (starting with "Version"). When you do that, I would suggest adding words to make the list flow better, e.g. adding "also", "lastly", etc.

docs/JLBP-14.md
```diff
@@ -1,29 +1,29 @@
...
+Version ranges can be a security hole. When a new version of a dependency
+is pushed to Maven Central, it is picked up by downstream dependencies
+automatically. This can be used a by a malicious actor with release privileges
+to slip new malicious code into projects without proper review. A version
+of this attack in the node.js ecosystem was used to steal
+Bitcoins](https://www.theregister.co.uk/2018/11/26/npm_repo_bitcoin_stealer/).
 
-- Single-element version ranges ("hard requirements") have a much different
-  impact, and this rule does not apply to them.
+Single-element version ranges ("hard requirements") have a much different
```
garrettjonesgoogle: I would combine this paragraph with the next one.

docs/JLBP-14.md
```diff
@@ -1,29 +1,29 @@
...
+build. For example:
+https://github.com/GoogleCloudPlatform/appengine-gcs-client/issues/71).
 
-  Thus, always specify a single version instead of a version range for dependencies.
+Version ranges can be a security hole. When a new version of a dependency
+is pushed to Maven Central, it is picked up by downstream dependencies
+automatically. This can be used a by a malicious actor with release privileges
+to slip new malicious code into projects without proper review. A version
+of this attack in the node.js ecosystem was used to steal
+Bitcoins](https://www.theregister.co.uk/2018/11/26/npm_repo_bitcoin_stealer/).
```
garrettjonesgoogle: The link title is missing the starting square bracket (`[`).

I would strongly recommend looking at the preview for your edits...

docs/JLBP-14.md
```diff
@@ -1,29 +1,29 @@
...
-    of this attack has been used in the node.js ecosystem to steal 
-    Bitcoins](https://www.theregister.co.uk/2018/11/26/npm_repo_bitcoin_stealer/). 
+Version ranges can pull in incomplete releases that break the
+build. For example:
+https://github.com/GoogleCloudPlatform/appengine-gcs-client/issues/71).
 
-  Thus, always specify a single version instead of a version range for dependencies.
+Version ranges can be a security hole. When a new version of a dependency
+is pushed to Maven Central, it is picked up by downstream dependencies
+automatically. This can be used a by a malicious actor with release privileges
```
garrettjonesgoogle: existing grammar problem: "used a by a" -> "used by a"

docs/JLBP-14.md
elharo: done

docs/JLBP-14.md
elharo: done

docs/JLBP-14.md
elharo: done

docs/JLBP-14.md
elharo: does the link only apply to hard requirements or to everything in this article?

garrettjonesgoogle: Hmm, based on the fact that the bullet point in the original was at the top level, I think it applies to everything. So, it makes sense to keep it separate.

elharo: makes sense, though looking at it now I realize I'm not sure what a hard requirement means and Maven doesn't say. Filed https://issues.apache.org/jira/browse/MNG-6791?filter=-2 to clarify

docs/JLBP-14.md
```diff
@@ -11,18 +12,16 @@ same code with the same compiler at different
...
 Version ranges can pull in incomplete releases that break the
 build. For example, a [version range in appengine-gcs-client's
-dependency on google-http-java-client broke multiple customers' 
+dependency on google-http-java-client broke multiple customers'
 builds](https://github.com/GoogleCloudPlatform/appengine-gcs-client/issues/71)
-when only half the artifacts in the project were pushed one afternoon. 
+when only half the artifacts in the project were pushed one afternoon.
 
-Version ranges can be a security hole. When a new version of a dependency
-is pushed to Maven Central, it is picked up by downstream dependencies
```
garrettjonesgoogle: "library with a version range" implies that the library you depend on is using a version range, which isn't what you're going for. I think you want working like "A project that uses a version range when depending on a library picks up new versions..."

elharo: done

docs/JLBP-14.md
```diff
@@ -1,7 +1,7 @@
-# [JLBP-14] Do not use version ranges
+# [JLBP-14] Specify exactly one version of each dependency
```
garrettjonesgoogle: I support the general idea of using affirmative statements instead of negated ones, but this particular affirmative statement doesn't feel right. To someone who is used to seeing only single version dependencies used, this is very confusing. I feel like "range" needs to be used somehow, because that's the poor practice.

docs/JLBP-14.md
```diff
@@ -16,10 +16,10 @@ dependency on google-http-java-client broke multiple customers'
 builds](https://github.com/GoogleCloudPlatform/appengine-gcs-client/issues/71)
 when only half the artifacts in the project were pushed one afternoon.
 
-Version ranges can be a security hole. A project that depends on a
-library with a version range picks up new versions of that dependency as they're
-pushed to Maven Central. This can be exploited by someone with release privileges
-to slip malicious code into projects without proper review. A variant
+Version ranges can be a security hole. A project that uses a range to specify
+the version of a dependency picks up new versions of that dependency when they're
```
garrettjonesgoogle: This doesn't sound right because a "range" is not "the version of a dependency", it is multiple allowable versions.

docs/JLBP-14.md
elharo: You're right that we rarely see ranges in practice. Given that, I think it's more confusing to put ranges, something many developers do not encounter, in the title.

docs/JLBP-14.md
elharo: makes sense. changed "specify" to "select"

docs/JLBP-14.md
garrettjonesgoogle: But avoiding ranges is the entire point of this JLBP. Someone who does not encounter ranges  and wouldn't think of using them doesn't even need to think about this JLBP.

suztomo: How about "Avoid version ranges"? (not using negated sentence)

I agree with Garrett. This JLBP is about version ranges; it's clear that having it in the title.

docs/JLBP-14.md
```diff
@@ -1,29 +1,47 @@
...
+requiring version 2.3 and no other. If two `dependency` elements in the
+graph have non-overlapping version ranges, the build fails.
 
-- Single-element version ranges ("hard requirements") have a much different
-  impact, and this rule does not apply to them.
+Even if all hard requirements can be satisfied, version ranges override
+the normal Maven algorithm for selecting among different versions of the
+same artifact. They pin the dependency tree to that version or versions,
+even when other artifacts in the classpath specify newer compatible versions
+that include important bug and security fixes.
```
garrettjonesgoogle: I think this point makes sense in general. The usage of hard requirements by grpc makes sense though: they are trying to control intra-grpc dependencies so that dependencies across grpc artifacts don't break when there are dependencies on internal unstable features. 

Maybe another way to state this: People shouldn't use hard requirements across libraries (nobody should depend on grpc using a hard requirement), but using them among the artifacts of a single library can actually be beneficial.

docs/JLBP-14.md
```diff
@@ -1,29 +1,47 @@
...
 
-  - Version ranges can be a security hole. When a new version of the dependency
-    is pushed to Maven Central, it will be picked up by downstream dependencies
-    automatically. This can be used a by a malicious actor with release privileges
-    to slip new malicious code into projects without proper review. [A version
-    of this attack has been used in the node.js ecosystem to steal 
-    Bitcoins](https://www.theregister.co.uk/2018/11/26/npm_repo_bitcoin_stealer/). 
+Version ranges have another, less obvious problem. A normal soft version
+requirement such as `<version>2.3</version>` says that Maven prefers
+version 2.3 but will accept version 2.0 or 3.0 if other dependencies
```
garrettjonesgoogle: I would argue that accepting 2.0 rarely works in practice. The reason is that as versions increase, features are often added in non-breaking ways. Thus, using a higher version than specified is generally safe; downgrading would lead to breakages due to missing features.

elharo: Every upper bounds failure is a case of accepting a lower version than expected. This happens fairly often in real projects. It's just a question of which node in the tree declares the dependency first.

docs/JLBP-14.md
elharo: I didn't realize gRPC did that. Its' built with gradle, fwiw. Where does it use a hard dependency? I didn't se any in the first couple of artifacts I looked at.

docs/JLBP-14.md
```diff
@@ -1,29 +1,66 @@
...
+pick that first. This flexibility is important
+for large projects with many dependencies, since it's very rare for
+all artifacts that depend on a particular library to agree on which version they
+need.
 
-  Thus, always specify a single version instead of a version range for dependencies.
+A version range does not have this flexibility. When Maven sees
+`<version>[2.3,2.9]</version>`, it accepts any version between 2.3 and 2.9 inclusive. However it does not accept 2.0 or 3.1.5.
```
suztomo: Sentences between this "However" is saying the same thing.

```suggestion
`<version>[2.3,2.9]</version>`, it accepts any version between 2.3 and 2.9 inclusive. For example, it does not accept 2.0 or 3.1.5.
```

Or just omit the conjunction.

docs/JLBP-14.md
```diff
@@ -1,29 +1,47 @@
...
+when only half the artifacts in the project were pushed one afternoon.
 
-  - Version ranges can pull in incomplete releases that break your
-    library's build (example:
-    https://github.com/GoogleCloudPlatform/appengine-gcs-client/issues/71).
+Version ranges can be a security hole. A project that uses a range to select
+the version of a dependency picks up new versions of that dependency when
+they're pushed to Maven Central. Someone with release privileges can take
+advantage of this to slip malicious code into projects without proper review. A
+variant of this [attack in the node.js ecosystem was used to steal
```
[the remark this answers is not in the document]
suztomo: There is a version in the link text.

docs/JLBP-14.md
```diff
@@ -1,29 +1,66 @@
-# [JLBP-14] Do not use version ranges
+# [JLBP-14] Specify a single, soft version of each dependency
 
-- If the pom.xml of your library specifies a version range instead of a
-  single version for any particular dependency, builds at different points
-  in time can see different versions of that dependency (as your dependency
-  releases new versions), which can break your library unexpectedly.
+Give the version of each dependency as a single value such as `2.3`
+instead of a range such as `[2.3,2.9]`, `[2.3,)`, or even `[2.3]`.
+When a pom.xml specifies a range instead of a
```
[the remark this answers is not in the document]
elharo: rewrote

docs/JLBP-14.md
elharo: done

docs/JLBP-14.md
```diff
@@ -1,29 +1,66 @@
...
+```
+<dependency>
+  <groupId>io.grpc</groupId>
+  <artifactId>grpc-core</artifactId>
+  <version>[1.25.0]</version>
+</dependency>
+```
+
+`io.grpc:grpc-alts:1.25.0` and `io.grpc:grpc-alts:1.23.0` cannot appear in 
+the classpath at the same time. If a dependency tree contains both
```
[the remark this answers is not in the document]
elharo: you're right, fixed

docs/JLBP-14.md
```diff
@@ -1,29 +1,66 @@
...
+`io.grpc:grpc-alts:1.25.0` and `io.grpc:grpc-alts:1.23.0`, the build fails, and the developer must fix the conflict before they can proceed.
+OpenCensus did not do this, and as a result there have been several problems
```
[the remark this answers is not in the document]
elharo: I think we're encouraging single-element version ranges for intra project dependencies that have tight coupling between artifacts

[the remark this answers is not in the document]
suztomo: Would you explicitly mention that? Without that, this idea contradicts 

> even `[2.3]`

 in line 4.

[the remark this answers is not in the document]
elharo: rewritten, but @garrettjonesgoogle asked for this exception.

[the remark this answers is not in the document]
suztomo: Thanks. But gRPC is not "_pre-1.0 projects with unstable APIs_". If we want to allow the exception for gRPC, the criteria should match gRPC.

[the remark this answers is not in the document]
elharo: Yes, and in writing this I started thinking that although gRPC does this, I'm not at all sure it should. The case for doing this in OpenCensus is a lot stronger than the case for doing ti in gRPC. I don't think we have to bless all existing behavior.

suztomo:
As per last comment, this needs to either
- disapprove gRPC's use of single version range, or
- change the criteria `Rapidly changing, pre-1.0 projects` part to match gRPC.

elharo:
These are two separate cases, though they are related. One is for intraproject dependencies. One is for Rapidly changing, pre-1.0 projects.

docs/JLBP-14.md
```diff
@@ -49,9 +48,9 @@ cause the build to fail early when multiple versions are mixed. For example,
 </dependency>
 ```
 
-`io.grpc:grpc-alts:1.25.0` and `io.grpc:grpc-alts:1.23.0` cannot appear in 
-the classpath at the same time. If a dependency tree contains both
-`io.grpc:grpc-alts:1.25.0` and `io.grpc:grpc-alts:1.23.0`, the build fails, and the developer must fix the conflict before they can proceed.
+`io.grpc:grpc-alts:1.25.0` and `io.grpc:grpc-core:1.23.0` cannot appear in 
+the classpath at the same time. If dependency mediation selects both
+`io.grpc:grpc-alts:1.25.0` and `io.grpc:grpc-core:1.23.0`, the build fails; and the developer must fix the conflict before they can proceed.
```
garrettjonesgoogle: The semicolon seems incorrect here

docs/JLBP-14.md
```diff
@@ -51,15 +51,21 @@ cause the build to fail early when multiple versions are mixed. For example,
 `io.grpc:grpc-alts:1.25.0` and `io.grpc:grpc-core:1.23.0` cannot appear in 
 the classpath at the same time. If dependency mediation selects both
 `io.grpc:grpc-alts:1.25.0` and `io.grpc:grpc-core:1.23.0`, the build fails; and the developer must fix the conflict before they can proceed.
+
+Rapidly changing, pre-1.0 projects with unstable APIs can use hard
```
garrettjonesgoogle: Is it a problem for post-1.0 projects with internal unstable APIs to use hard requirements for intraproject dependencies?

docs/JLBP-14.md
elharo: changed to comma

docs/JLBP-14.md
elharo: Yes, it's a problem for dependents no matter why or when you do it.

garrettjonesgoogle: How is it a problem? The only time it causes failures is when it *should* cause failures (i.e. when other means of detecting version mismatches have failed). The way grpc uses it, I don't think it has false positives.

elharo: It's a problem for the same reasons hard dependencies are always a problem: they make builds brittle and lead to hard to diagnose and fix errors. gRPC has imposed extra cost on dependents by  using this technique.

garrettjonesgoogle: The technique moves obscure runtime failures into build time failures. It's the same effect as dependency convergence, except that it only fails at the correct time.

@ejona86 can you weigh in? I'm arguing in favor of grpc's usage of hard requirements for internal dependencies, @elharo is arguing against it.

elharo: These shouldn't be obscure runtime failures either. They shouldn't be failures at all. Tight coupling between components causes problems for users. If the code can't be more decoupled and robust, then it shouldn't be in separate artifacts.

ejona86: This sort of stuff doesn't exist in a vacuum. I would agree such internal unstable APIs should be avoided.

But if gRPC can't use internal APIs then our options are to 1) bump major versions ~yearly (and sometimes more frequently since some changes are time sensitive) along with package renaming or 2) have larger artifacts with more dependencies (which actually doesn't work because Netty vs OkHttp wouldn't really fit with this model). (Or option (3), give up on performance.) Shading can help ease (2), but would cause us to use much more resources at runtime (like Threads) and only solves version pinning with grpc-core, not grpc-api. Remember that if it weren't for dependencies of the artifacts, they _would_ be all in one artifact. We don't _want_ them to be separate artifacts; they _need_ to be separate artifacts.

So you sort of need to pick your poison, and having to fight Maven to match the versions doesn't seem outrageous since you are _already_ fighting Maven to get upperBoundDeps which is similar complexity.

elharo: Yes, I agree with this, and this agrees with the proposed changes. You have to pick your poison, but this is poison. It does impose real costs on dependents, and should only be done if it's the least bad option. I don't want to mislead library designers into believing that using hard dependencies for intraproject dependencies won't cause problems.

garrettjonesgoogle:
I'm ok-ish with the text now, but I still have an issue with the title - I think the term "soft version" is infrequent enough to confuse people and isn't really transparent. What about: "Specify a single, overridable version of each dependency"

(I think "overridable" is sufficiently transparent)

elharo:
done

docs/README.md
```diff
@@ -31,7 +31,7 @@ list is open-ended, so new ones may be added from time to time.
 - [JLBP-12](JLBP-12.md): Make level of support and API stability clear
 - [JLBP-13](JLBP-13.md): Quickly remove references to deprecated features in
    dependencies
-- [JLBP-14](JLBP-14.md): Do not use version ranges
+- [JLBP-14](JLBP-14.md): Specify a single, soft version of each dependency
```
garrettjonesgoogle: Update here also

elharo: done

PR #912, opened 2019-09-12
use paragraphs for top level items, not lists JLBP-15

elharo:
@garrettjonesgoogle


elharo:
I've tried reworking the paragraphs. I'm still not perfectly happy with this one. It might work better with some examples of BOM code.

docs/JLBP-15.md
```diff
@@ -1,27 +1,28 @@
...
-v1.15.0, etc.
+## The BOM (Bill of Materials)
+
+A BOM is a `pom.xml` file that has `<packaging>pom</packaging>`.
+The `<dependencyManagement>` section lists all the modules of a project.
+The BOM can omit the `<modules>` section and does not have a `<dependencies>`
+section.
+
+Consumers of a multi-module library import the library's BOM
+in their own `dependencyManagement` section. They can then omit the
```
garrettjonesgoogle: Put the html `<` and `>` around `dependencyManagement` here to be consistent

elharo: done

elharo:
PTAL @garrettjonesgoogle

garrettjonesgoogle:
The only new change I see is adding the angle brackets - what about combining the paragraphs more?

elharo:
I'm not sure the paragraphs should be combined further. I suggest merging this as is, and if you think they should be further combined, then sending a PR with that change for discussion.

docs/JLBP-15.md
```diff
@@ -1,51 +1,63 @@
...
+imported, a BOM operates as a filter. It does not cause any modules to be
+added as dependencies. Instead, for any dependency in a project's dependency
+tree, *if* that module appears in the BOM, the version from the BOM is used.
+
+## The BOM (Bill of Materials)
+
+A BOM is a `pom.xml` file that has `<packaging>pom</packaging>`.
+The `<dependencyManagement>` section lists all the modules of a project.
+The BOM can omit the `<modules>` section and does not have a `<dependencies>`
+section.
```
garrettjonesgoogle: I would combine this paragraph with the next one and the following one.

docs/JLBP-15.md
```diff
@@ -1,51 +1,63 @@
...
+
+A BOM is not necessary for projects that only have one module, since there is
+no consistency problem in that case.
+
+See the details specific to each build system in the following sections.
+
+## Libraries using Maven
+
+Create a module containing only the BOM `pom.xml` file and add the BOM module
+to the parent `<modules>` list.
```
garrettjonesgoogle: I would combine this paragraph with the following one, and also merge the two bullet points into the paragraph as well, leaving only the example BOM line on its own.

docs/JLBP-15.md
```diff
@@ -1,51 +1,63 @@
...
+
+Example of maintaining a `pom.xml` file: gax-java -
   [build.gradle](https://github.com/googleapis/gax-java/blob/master/gax-bom/build.gradle)
   and
   [pom.xml](https://github.com/googleapis/gax-java/blob/master/gax-bom/pom.xml).
-- There is not yet an example of a project that generates a `pom.xml` file.
+
+There is not yet an example of a project that generates a `pom.xml` file.
+
   - Note: This approach has a limitation that the BOM cannot include any
```
garrettjonesgoogle: I would merge this bullet point into the paragraph before it.

docs/JLBP-15.md
elharo: done

docs/JLBP-15.md
elharo: done

docs/JLBP-15.md
elharo: done

elharo: done

docs/JLBP-15.md
```diff
@@ -29,35 +27,28 @@ no consistency problem in that case.
...
-Example of maintaining a `pom.xml` file: gax-java -
   [build.gradle](https://github.com/googleapis/gax-java/blob/master/gax-bom/build.gradle)
   and
   [pom.xml](https://github.com/googleapis/gax-java/blob/master/gax-bom/pom.xml).
 
-There is not yet an example of a project that generates a `pom.xml` file.
-
-  - Note: This approach has a limitation that the BOM cannot include any
-    classifiers in the dependency specifications.
+This approach has a limitation that the BOM cannot include any classifiers in the dependency specifications.
```
garrettjonesgoogle: The limitation stated here applies to the approach of generating a `pom.xml` file, so since you removed the statement that there is no example, this limitation now seems to apply to the approach of using an explicit `pom.xml` file, which is incorrect.

elharo: Hmm, I didn't realize that. Why can't generated pom.xml include classifiers?

garrettjonesgoogle: It's a flaw in the gradle support for pom generation.

elharo: Do you have a link to that feature we can include here? Is it a standard part of gradle or an extra plugin?

garrettjonesgoogle:
I can't easily find the details of the pom generation any more. I think at this point, it sounds reasonable to remove the mention of it.

PR #914, opened 2019-09-12
use paragraphs for top level items, not lists JLBP-16

elharo:
@garrettjonesgoogle


elharo:
Per https://developers.google.com/style/lists  "In most cases, you should precede a list with an introductory sentence. The sentence can end with a colon or a period; usually a colon if it immediately precedes the list, usually a period if there's more material (such as a note paragraph) between the introduction and the list."

garrettjonesgoogle:
That guidance makes sense, but it's not really the change you are making. The guidance would suggest adding a sentence before the lists that are there.

PR #928, opened 2019-09-26
update protobuf to 3.10.0

elharo:
@suztomo

### DIFF
--- a/boms/cloud-oss-bom/pom.xml
+++ b/boms/cloud-oss-bom/pom.xml
@@ -46,7 +46,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <guava.version>28.1-android</guava.version>
     <google.cloud.java.version>0.100.0-alpha</google.cloud.java.version>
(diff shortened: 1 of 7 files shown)


elharo:
I think the kokoro failures are flaky:

[ERROR] Non-resolvable import POM: Failure to find com.google.protobuf:protobuf-bom:pom:3.10.0 in https://maven-central.storage-download.googleapis.com/repos/central/data/ was cached in the local repository, resolution will not be reattempted until the update interval of google-maven-central has elapsed or updates are forced @ com.google.cloud:libraries-bom:2.4.0-SNAPSHOT, /Volumes/BuildData/tmpfs/src/github/cloud-opensource-java/boms/cloud-oss-bom/pom.xml, line 73, column 19
[ERROR] 'dependencies.dependency.version' for com.google.cloud:google-cloud-core:jar is missing. @ line 56, column 17
[ERROR] 'dependencies.dependency.version' for com.google.cloud:google-cloud-storage:jar is missing. @ line 60, column 17

PR #939, opened 2019-09-30
Release BOM 2.4.0

elharo:
@suztomo


elharo:
I'm guessing that the next release will include an upgrade to google-cloud-java.

elharo:
Can you be more specific about "use the prefixed tag “bom-v2.4.0”. This does not match the release instructions. Do they need to be updated?

suztomo:
The release instruction remains the same. It continues to use Git tag for "committish" field. But our release tag has prefix "bom-" now:

image

PR #946, opened 2019-09-30
Bom 2.5.0

elharo:
@suztomo Is the script supposed to update the version number in pom.xml?

### DIFF
--- a/boms/cloud-oss-bom/pom.xml
+++ b/boms/cloud-oss-bom/pom.xml
@@ -7,7 +7,7 @@
 
   <groupId>com.google.cloud</groupId>
   <artifactId>libraries-bom</artifactId>
-  <version>2.5.0</version>
+  <version>2.5.0-SNAPSHOT</version>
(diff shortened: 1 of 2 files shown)


suztomo:
Yes, it is supposed to increment patch version when next version is omitted. I might introduce a bug when I updated it to take additional argument.

elharo:
Or when the next version isn't omitted. The command I ran was

../../scripts/prepare_release.sh bom 2.5.0 2.6.0

elharo:
Yes, that's it. It works unless you specify the new version. 

https://github.com/GoogleCloudPlatform/cloud-opensource-java/compare/bom-2.6.0

elharo:
I think

```
if [ -n "$3" ]; then
  NEXT_VERSION=$2
  CheckVersion ${NEXT_VERSION}
```

should perhaps be

```
if [ -n "$3" ]; then
  NEXT_VERSION=$3
  CheckVersion ${NEXT_VERSION}
```

PR #951, opened 2019-10-02
expand discussion of dependency scope

### LINKED ISSUE #950: JLBP-1: no compile-only scope
In JLBP-1 we say:

Use the smallest scope possible. For example, AutoValue doesn't need to use compile scope, and can instead use compile-only, since it doesn't need to appear on the classpath of consumers.

However Maven doesn't have a compile-only scope and in Gradle it's compileOnly. 

We should either fix this here, or perhaps write a completely new JLBP to "Use the smallest scope possible" that covers this as well as provided and optional.

elharo:
@garrettjonesgoogle @saturnism fixes #950

### DIFF
--- a/docs/JLBP-1.md
+++ b/docs/JLBP-1.md
@@ -4,47 +4,89 @@ Use the minimum number of dependencies that is reasonable.
 Every dependency of a library is a liability of both
 that library and its consumers.
 
-Adding a dependency for something that is difficult and complicated to do may be OK,
+Adding a dependency for something that is difficult and complicated may be OK,
 but avoid adding a dependency just to save a few lines of code.
+Remember you're also paying the cost of transitive dependencies.
+Before adding a dependency, consider what that dependency depends on.
 
-Some specific notes about minimizing dependencies:
-
-- Remember you're also paying the cost of transitive dependencies.
-  Before adding a dependency, also consider what that dependency
-  depends on.
-
-- Use the smallest scope possible. For example, AutoValue doesn't
-  need to use `compile` scope, and can instead use `compile-only`,
-  since it doesn't need to appear on the classpath of consumers.
-  
-  - Libraries used only for testing should have `test` scope
-    (for example junit, mockito, and truth).
-
-- Scrutinize all dependency additions. Whenever you add a new
-  dependency, check the full tree of transitive dependencies that
-  are pulled in as a result. If a large number of transitive
-  dependencies are pulled in, consider a different direct dependency.
-  Alternatively, if the functionality you need is small, reimplement
-  it in your own library.
+Scrutinize all dependency additions. Whenever you add a new
+dependency, check the full tree of transitive dependencies that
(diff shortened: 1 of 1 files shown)


docs/JLBP-1.md
```diff
@@ -4,47 +4,89 @@ Use the minimum number of dependencies that is reasonable.
...
+If you can reasonably reimplement functionality instead of adding
+another dependency, do so. For example, if the only classes you're
+using from Guava are `Preconditions` and `Strings`, it's not
+worth adding a dependency on Guava. You can easily reimplement
+any method in those classes.  
+
+## Minimize dependency scope
+
+When you do add a dependency, keep it scoped as narrowly as possible. 
+In Maven libraries used only for testing such as JUnit, Mockito, and Truth
```
garrettjonesgoogle: Remove "In" at the beginning of the sentence

docs/JLBP-1.md
```diff
@@ -4,47 +4,89 @@ Use the minimum number of dependencies that is reasonable.
...
+should have `test` scope:
+
+```
+  <dependency>
+    <groupId>junit</groupId>
+    <artifactId>junit</artifactId>
+    <version>4.12</version>
+    <scope>test</scope>
+  </dependency>
```
garrettjonesgoogle: missing the ending 3 ticks

docs/JLBP-1.md
```diff
@@ -4,47 +4,89 @@ Use the minimum number of dependencies that is reasonable.
...
+
+```
+  dependencies {
+    compileOnly 'com.google.auto.value:auto-value-annotations:1.6.6'
+  }
+```
+
+The `provided` scope should be used when the dependency is needed at runtime
+and compile time. However the specific jar will be supplied by the environment
+in which the code runs; for example a Java servlet container.
```
garrettjonesgoogle: I'm going to pull an Elliotte and say that you should avoid the semicolon here

docs/JLBP-1.md
elharo: done

docs/JLBP-1.md
elharo: done

docs/JLBP-1.md
elharo: done

docs/JLBP-1.md
```diff
@@ -82,11 +83,5 @@ In Gradle these libraries should have `compileOnly` scope:
 
 The `provided` scope should be used when the dependency is needed at runtime
 and compile time. However the specific jar will be supplied by the environment
-in which the code runs; for example a Java servlet container.
-
-
-
-- Use the smallest scope possible. For example, AutoValue doesn't
```
garrettjonesgoogle: I know that you use AutoValue in your examples, but I think it is still worth calling it out in the text because it's a common mistake in Google-owned Java libraries.

elharo: done

PR #957, opened 2019-10-14
Specify libraries-bom

elharo:
@garrettjonesgoogle


DECLARING_DEPENDENCIES.md
```diff
@@ -69,19 +69,18 @@ select compatible versions.
 
 Conflicts can occur when multiple artifacts from a single library are
 part of a dependency tree, and Maven resolves different artifacts from
-that library to versions from different releases. Using a BOM fixes
-this problem because a BOM dictates consistent versions for all
-artifacts from a library.
+that library to versions from different releases. The libraries-bom
```
garrettjonesgoogle: maybe put ticks around libraries-bom (e.g. `libraries-bom`)

DECLARING_DEPENDENCIES.md
```diff
@@ -69,19 +69,18 @@ select compatible versions.
...
+dictates consistent versions for all GCP orbit artifacts.
 
-You can use a BOM like this—this example is for `google-cloud-bom`:
+You use the BOM like this:
```
garrettjonesgoogle: The purpose of this section in the overall doc is to show how BOMs are imported in general. It is a section lower down that makes recommendations on which BOMs to use.

DECLARING_DEPENDENCIES.md
elharo: done

DECLARING_DEPENDENCIES.md
elharo: I found myself coming here to find info about how to add the BOM to other projects, and not finding what I needed. It's a simpler story and usually best to say just use the libraries-bom.

DECLARING_DEPENDENCIES.md
```diff
@@ -12,76 +12,29 @@ These recommendations apply to the following libraries:
 - [guava](https://github.com/google/guava)
 - [protobuf](https://github.com/protocolbuffers/protobuf)
 - [grpc-java](https://github.com/grpc/grpc-java)
-- [gax-java](https://github.com/googleapis/gax-java)
```
garrettjonesgoogle: Why are you removing libraries from this list?

DECLARING_DEPENDENCIES.md
```diff
@@ -12,76 +12,29 @@ These recommendations apply to the following libraries:
...
-          </execution>
-        </executions>
-      </plugin>
-    </plugins>
-  </build>
-```
-
-When the newest versions are not compatible, you can use BOMs to
-select compatible versions.
+libraries, import com.google.cloud:libraries-bom  and use that to
```
garrettjonesgoogle: add ticks around the group & artifact ID

DECLARING_DEPENDENCIES.md
```diff
@@ -12,76 +12,29 @@ These recommendations apply to the following libraries:
...
-- [beam](https://github.com/apache/beam)
-
-## Ensuring Compatibility
 
 To ensure that your own project uses compatible versions of these
-libraries, follow the guidance below for your build system.
-
-### Maven
-
-#### Use the requireUpperBoundDeps enforcer rule
```
garrettjonesgoogle: I don't think this should be removed. Users still need to use this enforcer rule for safety.

PR #966, opened 2019-10-18
Use a golden file to test that the BOM artifacts only change intentionally

netdpb:
Use a golden file to test that the BOM artifacts only change intentionally.


netdpb:
When the test fails, you can update the golden file by running:

```shell
$ UPDATE_GOLDEN_ARTIFACTS=yes mvn test -pl dependencies -am -Dtest=RepositoryUtilityTest
```

Then the test passes without that environment variable.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtilityTest.java
```diff
@@ -57,34 +70,75 @@ public void testReadBom_coordinates() throws ArtifactDescriptorException {
...
-    // However sometimes this list does change. If so, we want to 
-    // output the specific difference so we can manually verify whether
-    // the changes make sense. When they do make sense, we update the test. 
-    if (currentArtifacts.size() != 217) {
-      Truth.assertThat(currentArtifacts).containsExactlyElementsIn(oldArtifacts);
+    assertThat(coordinates).startsWith("com.google.cloud:libraries-bom:");
+    assertThat(coordinates).endsWith("-SNAPSHOT");
+
+    // Assert that the managed dependencies haven't unexpectedly changed.
+    // If this fails, update the golden file so the changes can be reviewed by running
```
suztomo: Now I got the idea. Nice enhancement.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtilityTest.java
```diff
@@ -57,34 +70,75 @@ public void testReadBom_coordinates() throws ArtifactDescriptorException {
...
+    // this test with UPDATE_GOLDEN_ARTIFACTS set to any nonempty string. Then
+    // rerun the test without that.
+
+    if (shouldUpdateGoldenArtifactsFile()) {
+      updateGoldenArtifactsFileAndFail(currentBom.getManagedDependencies());
     }
+
+    assertWithMessage(
+            "If BOM artifacts have changed, rerun this test setting the environment variable %s"
+                + " and then try again.",
```
suztomo:
```suggestion
                + " and then try again. If you see this in a presubmit, ensure your PR includes the change in goldenBomArtifacts.txt.",
```

How about adding the above? Just in case when I encounter this error in Circle CI for a PR after forgetting to include the file change.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtilityTest.java
netdpb: Thanks.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtilityTest.java
netdpb: Done.

netdpb:
@suztomo I made a small change, but I'm not sure it's better. Can you take a look and tell me if you'd rather I revert it?

suztomo:
@netdpb It's good learning for me especially `reduce`. However I now feel `Paths.get("src", "test", "resources", "goldemBomArtifacts.txt")` would just work and easier to read.

Is there any need to place the file under `src/test/resources/com/google/cloud/tools/opensource/dependencies` rather than `src/test/resources`?
I thought of the deep directory structure clarifying the correspondence between the test and the golden file, but the golden file's first few lines explain where it is referenced.

netdpb:
> Is there any need to place the file under `src/test/resources/com/google/cloud/tools/opensource/dependencies` rather than `src/test/resources`?
> I thought of the deep directory structure clarifying the correspondence between the test and the golden file, but the golden file's first few lines explain where it is referenced.

True. I'd want to rename the file as well. I'm used to Bazel, where you can place resources next to source files in the file system instead of in a separate "resources" directory. Let me try that.

dependencies/src/test/resources/RepositoryUtilityTestGoldenBomArtifacts.txt
```diff
@@ -0,0 +1,223 @@
+# Automatically generated by running tests with environment variable
```
elharo: I strongly prefer not to check any auto generated files into the repo. That's a big antipattern. We can autogenerate files once or every time. In this case, I think it makes sense to do it once and then manually edit it from there on in.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtilityTest.java
```diff
@@ -57,34 +69,73 @@ public void testReadBom_coordinates() throws ArtifactDescriptorException {
...
   }
 
+  private static final String UPDATE_GOLDEN_ARTIFACTS = "UPDATE_GOLDEN_ARTIFACTS";
+  private static final String GOLDEN_FILE_NAME =
+      RepositoryUtilityTest.class.getSimpleName() + "GoldenBomArtifacts.txt";
+  private static final Path GOLDEN_FILE = Paths.get("src", "test", "resources", GOLDEN_FILE_NAME);
+
   @Test
-  public void testReadBom_path() throws MavenRepositoryException, ArtifactDescriptorException {
+  public void testReadBom_path() throws MavenRepositoryException, IOException {
```
suztomo: The purpose of this `testReadBom_path` is to check `RepositoryUtility.readBom(Path)`, not the content of our BOM. Because of the earlier choice of test data `../boms/cloud-oss-bom/pom.xml` for this test, we're now having trouble maintaining the assertion.

If this test just uses a static pom.xml, such as `src/test/resources/libraries-bom-2.2.1-SNAPSHOT.pom` (new file), the problem of maintaining assertion is resolved.

dependencies/src/test/resources/RepositoryUtilityTestGoldenBomArtifacts.txt
netdpb: This is a common pattern for golden file tests. I think it's helpful to make this test easy to maintain, if we want the test at all (see below).

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtilityTest.java
netdpb: Going back to an earlier conversation: what is the purpose of this test?

If it's to test the code that reads a local BOM file to make sure it properly evaluates the indirect BOM references, etc., then a static test file as @suztomo suggests is perfect. It'll never change, and so it's a good data file for a regression test.

If it's to test that any changes to the libraries BOM are intentional and not inadvertent, then a golden file test like this makes sense because it forces code reviewers to approve any change to the effective BOM.

Based on previous discussions with @elharo, I thought that the purpose of the test was the latter. If that's not the case, then the golden file strategy doesn't make sense.

elharo: It's doing multiple things and perhaps we should split it as @suztomo suggests. The first purpose has always been to test that the `readBom(Path)` method work correctly.

dependencies/src/test/resources/RepositoryUtilityTestGoldenBomArtifacts.txt
elharo: No uneditable code in the repo. That's a hard line for me. I've bit my tongue and let this go in the past in the interests of immediate convenience -- e.g. in Cloud Tools for Eclipse target platform definitions -- and I've always ended up regretting it.

netdpb: Up to you.

PR #990, opened 2019-10-24
DependencyPath to hold scope and optional flag

### LINKED ISSUE #981: User-friendly Linkage Monitor error message
When investigating #980, the user could not find where gwt-user is coming from. So he was not able to take action upon the error message.

# Options 

## Option 1: Show Dependency Tree

The output would become too much to digest.

## Option 2: Show Path to the Problematic Artifact

This sounds reasonable.

## Option 3: Run Dashboard

This is what I usually do to investigate errors from Linkage Monitor.

# Scope of Dependency

It's important to add scope information to the nodes. As of now, the dashboard only shows artifacts:

```
The following paths to the jar file from the BOM are found in the dependency tree:
  com.google.cloud:google-cloud-firestore:1.28.1-SNAPSHOT / com.google.truth:truth:1.0 / com.google.gwt:gwt-user:2.8.2
```

suztomo:
Fixes #981 .

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
@@ -271,18 +271,18 @@ public class DependencyGraphBuilder {
     while (!queue.isEmpty()) {
       LevelOrderQueueItem item = queue.poll();
       DependencyNode dependencyNode = item.dependencyNode;
-      DependencyPath forPath = new DependencyPath();
+      DependencyPath path = new DependencyPath();
       Stack<DependencyNode> parentNodes = item.parentNodes;
       parentNodes.forEach(
           parentNode ->
-              forPath.add(
+              path.add(
                   parentNode.getArtifact(),
                   parentNode.getDependency().getScope(),
                   parentNode.getDependency().getOptional()));
       if (dependencyNode.getArtifact() != null) {
         // When requesting dependencies of 2 or more artifacts, root DependencyNode's artifact is
         // set to null
-        forPath.add(
+        path.add(
             dependencyNode.getArtifact(),
             dependencyNode.getDependency().getScope(),
             dependencyNode.getDependency().getOptional());
@@ -295,7 +295,7 @@ public class DependencyGraphBuilder {
           continue;
         }
         parentNodes.push(dependencyNode);
-        graph.addPath(forPath);
+        graph.addPath(path);
 
         if (resolveFullDependency && !"system".equals(dependencyNode.getDependency().getScope())) {
           Artifact dependencyNodeArtifact = dependencyNode.getArtifact();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
@@ -273,13 +273,17 @@ public class DependencyGraphBuilder {
       DependencyNode dependencyNode = item.dependencyNode;
       DependencyPath forPath = new DependencyPath();
       Stack<DependencyNode> parentNodes = item.parentNodes;
-      parentNodes.forEach(parentNode -> forPath.add(parentNode.getArtifact(),
-          parentNode.getDependency().getScope(),
-          parentNode.getDependency().getOptional()));
+      parentNodes.forEach(
+          parentNode ->
+              forPath.add(
+                  parentNode.getArtifact(),
+                  parentNode.getDependency().getScope(),
+                  parentNode.getDependency().getOptional()));
       if (dependencyNode.getArtifact() != null) {
         // When requesting dependencies of 2 or more artifacts, root DependencyNode's artifact is
         // set to null
-        forPath.add(dependencyNode.getArtifact(),
+        forPath.add(
+            dependencyNode.getArtifact(),
             dependencyNode.getDependency().getScope(),
             dependencyNode.getDependency().getOptional());
         if (resolveFullDependency && parentNodes.contains(dependencyNode)) {
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
@@ -273,11 +273,15 @@ public class DependencyGraphBuilder {
       DependencyNode dependencyNode = item.dependencyNode;
       DependencyPath forPath = new DependencyPath();
       Stack<DependencyNode> parentNodes = item.parentNodes;
-      parentNodes.forEach(parentNode -> forPath.add(parentNode.getArtifact()));
+      parentNodes.forEach(parentNode -> forPath.add(parentNode.getArtifact(),
+          parentNode.getDependency().getScope(),
+          parentNode.getDependency().getOptional()));
       if (dependencyNode.getArtifact() != null) {
         // When requesting dependencies of 2 or more artifacts, root DependencyNode's artifact is
         // set to null
-        forPath.add(dependencyNode.getArtifact());
+        forPath.add(dependencyNode.getArtifact(),
+            dependencyNode.getDependency().getScope(),
+            dependencyNode.getDependency().getOptional());
         if (resolveFullDependency && parentNodes.contains(dependencyNode)) {
           logger.severe(
               "Infinite recursion resolving "
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
@@ -18,10 +18,13 @@ package com.google.cloud.tools.opensource.dependencies;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Streams;
 import java.util.ArrayList;
 import java.util.List;
 
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import org.eclipse.aether.artifact.Artifact;
 
 /**
@@ -42,8 +45,16 @@ public final class DependencyPath {
   
   @Override
   public String toString() {
-    return Joiner.on(" / ")
-        .join(path.stream().map(Artifacts::toCoordinates).collect(Collectors.toList()));
+
+    ImmutableList.Builder<String> elements = ImmutableList.builder();
+    for (int i=0; i< path.size(); ++i) {
+      Artifact artifact = path.get(i);
+      String scope = scopes.get(i);
+      boolean optional = optionals.get(i);
+      elements.add(String.format("%s (%s%s)", Artifacts.toCoordinates(artifact),
+          scope, optional ? ", optional" : ""));
+    }
+    return Joiner.on(" / ").join(elements.build());
   }
   
   @Override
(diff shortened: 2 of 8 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -30,16 +30,29 @@
 public final class DependencyPath {
 
   private List<Artifact> path = new ArrayList<>();
+  private List<String> scopes = new ArrayList<>();
+  private List<Boolean> optionals = new ArrayList<>();
```
suztomo: These two lines are the purpose of this PR; DependencyPath to hold scope and optional flag.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -273,11 +273,19 @@ private static DependencyGraph levelOrder(
       DependencyNode dependencyNode = item.dependencyNode;
       DependencyPath forPath = new DependencyPath();
```
elharo: `forPath` is a very strange name. Any ideas for making it clearer?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
elharo: so we have three parallel lists? That feels off. We should have one list of a class that contains artifact, scope, and optional.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -30,16 +30,29 @@
...
-    return Joiner.on(" / ")
-        .join(path.stream().map(Artifacts::toCoordinates).collect(Collectors.toList()));
+
+    ImmutableList.Builder<String> elements = ImmutableList.builder();
+    for (int i = 0; i < path.size(); ++i) {
+      Artifact artifact = path.get(i);
+      String scope = scopes.get(i);
+      boolean optional = optionals.get(i);
+      elements.add(
+          String.format(
```
elharo: any reason not to use a local variable and string concatenation here?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
suztomo: That's a nice idea. Updated with DependencyPath.Node class.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
suztomo: I like joiner to add something inbetween elements; I don't have to worry about removing a separator from tail or head.

Updated with DependencyPath.Node's toString.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Updated to path.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -54,49 +71,99 @@ public boolean equals(Object o) {
...
+      }
+      if (thisNode.isOptional() != otherNode.isOptional()) {
+        return false;
+      }
     }
     return true;
   }
 
-  @Override
-  public int hashCode() {
```
suztomo: hashCode is moved 10 lines below.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -16,35 +16,52 @@
...
   @Override
   public String toString() {
-    return Joiner.on(" / ")
-        .join(path.stream().map(Artifacts::toCoordinates).collect(Collectors.toList()));
+    return Joiner.on(" / ").join(path);
   }
   
   @Override
   public boolean equals(Object o) {
-    if (o == null || !(o instanceof DependencyPath)) {
```
suztomo: o == null is already covered by `o instanceof DependencyPath`. (as per IntelliJ)

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -54,49 +71,99 @@ public boolean equals(Object o) {
...
+   * <p>{@link org.eclipse.aether.artifact.DefaultArtifact#equals(Object)} uses {@code file} field,
+   * which is unnecessary for our usage.
    */
   private static boolean artifactsEqual(Artifact artifact1, Artifact artifact2) {
     return artifact1.getArtifactId().equals(artifact2.getArtifactId())
         && artifact1.getGroupId().equals(artifact2.getGroupId())
         && artifact1.getVersion().equals(artifact2.getVersion());
   }
 
-  public int size() {
```
suztomo: size, getLeaf, getPath, and get are moved to upper part.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -54,49 +71,99 @@ public boolean equals(Object o) {
...
-  public List<Artifact> getPath() {
-    return path;
-  }
+    private Node(Artifact artifact, String scope, boolean optional) {
+      this.artifact = checkNotNull(artifact);
+      this.scope = checkNotNull(scope);
+      this.optional = optional;
+    }
 
-  // TODO think about index out of bounds
```
suztomo: This TODO is removed; I don't see this array index access is special. If out of bounds, let JVM raise IndexOutOfBoundsException.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -16,48 +16,38 @@
...
 import com.google.common.collect.ImmutableList;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 import org.eclipse.aether.artifact.Artifact;
 
 /**
- * A representation of a dependency path from a root Artifact to a node Artifact.
+ * Path from the root to a node Artifact in a dependency tree.
```
elharo: what's a "node Artifact"?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -67,20 +57,37 @@ public boolean equals(Object o) {
     }
     
     for (int i = 0; i < path.size(); i++) {
-      if (!artifactsEqual(path.get(i), other.path.get(i))) {
+      Node thisNode = path.get(i);
+      Node otherNode = other.path.get(i);
+      // DefaultArtifact's equality checks file, which we do not care
```
elharo: "care about"

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -100,16 +107,66 @@ public int size() {
...
   public Artifact get(int i) {
-    return path.get(i);
+    return path.get(i).getArtifact();
+  }
+
+  /** Node in a dependency tree, holding scope and optional flag. */
+  private static class Node {
+    private Artifact artifact;
+    private String scope;
+    private Boolean optional;
```
elharo: boolean?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -100,16 +107,66 @@ public int size() {
...
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hash(artifact, scope, optional);
+    }
+
+    @Override
+    public String toString() {
+      return String.format(
```
elharo: This opaque. Breaking it up into steps with local variables and perhaps using concatenation instead of format should make it clearer.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -16,35 +16,52 @@
...
 
   @VisibleForTesting
-  public void add(Artifact artifact) {
-    path.add(artifact);
+  public void add(Artifact artifact, String scope, Boolean optional) {
+    path.add(new Node(artifact, scope, optional));
   }
-  
+
+  public int size() {
```
elharo: do these new methods all need to be public? If so, they should probably have JavaDoc.

suztomo: They are not new. Size, getLeaf, getPath, and get just moved from the lower part. (getPath is renamed to getArtifacts)

Added Javadoc of these functions.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -16,35 +16,62 @@
...
+    return path.size();
+  }
+
+  /** Returns the artifact in the leaf (the furthest node from the node) of the path. */
+  public Artifact getLeaf() {
+    return path.get(size() - 1).getArtifact();
+  }
+
+  /** Returns the list of artifact in the path. */
+  public ImmutableList<Artifact> getArtifacts() {
```
suztomo: This is renamed from `getPath`.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -271,13 +271,21 @@ private static DependencyGraph levelOrder(
...
       LevelOrderQueueItem item = queue.poll();
       DependencyNode dependencyNode = item.dependencyNode;
-      DependencyPath forPath = new DependencyPath();
+      DependencyPath path = new DependencyPath();
       Stack<DependencyNode> parentNodes = item.parentNodes;
-      parentNodes.forEach(parentNode -> forPath.add(parentNode.getArtifact()));
+      parentNodes.forEach(
+          parentNode ->
+              path.add(
+                  parentNode.getArtifact(),
```
netdpb: Should `DependencyPath.add()` take a `DependencyNode` and decompose it itself?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyPath.java
```diff
@@ -54,49 +81,100 @@ public boolean equals(Object o) {
...
+    }
+    return hashCode;
   }
 
-  public Artifact getLeaf() {
-    return path.get(size() - 1);
-  }
+  /** Node in a dependency tree, holding scope and optional flag. */
+  private static class Node {
+    private Artifact artifact;
```
netdpb: These should be `final`.

suztomo: Updated to final.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: That's a nice suggestion. But I would keep the three parameters. The idea of taking `DependencyNode` would make this line simpler, but other places, such as DependencyPathTest, would become complex like this:

```
add(new DefaultDependencyNode(new Dependency(foo, "compile", false)));
```

netdpb: You could have an overload.

PR #997, opened 2019-10-25
MaximumlinakgeErrorsTest to automatically check with the latest version of the BOM

suztomo:
Followup of https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/993#pullrequestreview-307249115. With this PR, we no longer need to update the version number of this test.

Example output:
```
java.lang.AssertionError: Baseline BOM: com.google.cloud:libraries-bom:2.7.0
Newly introduced problems:
Class android.util.JsonToken is not found
Class android.util.JsonReader is not found
Class android.util.JsonWriter is not found
Class org.junit.rules.TestRule is not found
Class org.junit.runners.model.MultipleFailureException is not found
Class com.google.inject.Provider is not found
Class com.google.inject.util.Providers is not found
Class com.google.appengine.repackaged.com.google.common.html.HtmlEscapers is not found
```

### DIFF
--- a/boms/integration-tests/src/test/java/com/google/cloud/MaximumLinkageErrorsTest.java
+++ b/boms/integration-tests/src/test/java/com/google/cloud/MaximumLinkageErrorsTest.java
@@ -25,6 +25,7 @@ import com.google.cloud.tools.opensource.classpath.SymbolProblem;
 import com.google.cloud.tools.opensource.dependencies.Bom;
 import com.google.cloud.tools.opensource.dependencies.MavenRepositoryException;
 import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.collect.Sets;
@@ -32,6 +33,7 @@ import com.google.common.collect.Sets.SetView;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Optional;
 import org.eclipse.aether.RepositoryException;
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


boms/integration-tests/src/test/java/com/google/cloud/MaximumLinkageErrorsTest.java
```diff
@@ -70,13 +76,24 @@ public void testMaximumLinkageErrors()
...
+        message.append("Newly introduced classes linking to " + problem + ":\n");
         for (ClassFile classFile : newReferences) {
           message.append("Link from " + classFile + "\n");
         }
         Assert.fail(message.toString());
       }
     }
   }
+
+  private String findLatestNonSnapshotVersion() throws MavenRepositoryException {
```
elharo: There's already RepositoryUtility.findHighestVersion. I guess that returns snapshot versions too?

boms/integration-tests/src/test/java/com/google/cloud/MaximumLinkageErrorsTest.java
```diff
@@ -70,13 +76,24 @@ public void testMaximumLinkageErrors()
...
+    ImmutableList<String> versions =
+        RepositoryUtility.findVersions(newRepositorySystem(), "com.google.cloud", "libraries-bom");
```
elharo: This method doesn't seem to promise returning the version in any particular order.

suztomo: Clarified that in the javadoc in `findVersions`. [Javadoc for ` VersionRangeResult.getVersions()`](https://maven.apache.org/resolver/apidocs/org/eclipse/aether/resolution/VersionRangeResult.html#getVersions()) ensures the order.

boms/integration-tests/src/test/java/com/google/cloud/MaximumLinkageErrorsTest.java
suztomo: Yes. Clarified it with comment.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -366,11 +366,14 @@ public static String findHighestVersion(
 
   /**
    * Returns list of versions available for {@code groupId:artifactId} in {@code repositorySystem}.
+   * The returned list is in ascending order with regard to {@link
```
suztomo: Clarified the order of the elements.

boms/integration-tests/src/test/java/com/google/cloud/MaximumLinkageErrorsTest.java
```diff
@@ -17,19 +17,23 @@
 
 package com.google.cloud;
 
+import static com.google.cloud.tools.opensource.dependencies.RepositoryUtility.newRepositorySystem;
```
elharo: This method probably shouldn't be a static import since it's an uncommon method and no one will know where it comes from, unlike assertTrue, for example.

suztomo: Updated.

PR #999, opened 2019-10-28
Initial draft of "Separate tool classpath from product classpath"

### LINKED ISSUE #986: JLBP to recommend separating annotationProcessorPaths from dependencies / compile classpath?
Dagger users sometimes run into trouble because that they put the Dagger annotation processor on the compile-time classpath (generally as a `provided` or `optional` dependency), only to find that its dependencies conflict with their project's actual compile dependencies. This apparently was [required before `maven-compiler-plugin` 3.5](https://github.com/google/dagger/blob/master/README.md#maven), but with newer versions, it's better to [use `annotationProcessorPaths`](https://github.com/google/dagger/blob/master/README.md#maven) (or the equivalent in Gradle, etc., also shown [in the Dagger docs](https://github.com/google/dagger/blob/master/README.md#gradle)).

We've updated docs for [AutoService](https://github.com/google/auto/tree/master/service#getting-started) and [AutoValue](https://github.com/google/auto/blob/master/value/userguide/index.md#with-maven) accordingly, but likely we have other projects to update, and this advice might make a useful JLBP for others?

elharo:
@cpovirk @saturnism 

TBD: should this stand alone as in this PR, or is this another section for JLBP-1?

fixes #986


docs/JLBP-21.md
```diff
@@ -0,0 +1,31 @@
...
+It is critical not to mix the classpath of the tool with the classpath of the product.
+Dependencies of the tool should not be dependencies of the product.
+For example, there's no reason `javax.tools.JavaCompiler` should appear in the
+classpath of an MP3 player simply because the product was compiled by `javac`.
+
+Indeed `javac` does not transmit its own dependencies into the products
+it builds, but not all tools are this well behaved. Maven annotation processors
+such as AutoValue and Animal Sniffer are sometimes declared to be dependencies
+of the product itself in the pom.xml. Since they are only needed at
+compile time, they should instead be added to the annotation processor classpath.
```
cpovirk: Is it worth linking to https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#annotationProcessorPaths and https://docs.gradle.org/4.6/release-notes.html#convenient-declaration-of-annotation-processor-dependencies ? I haven't checked whether we normally provide such links to specific tools.

cpovirk:
(No opinion personally about whether this becomes part of JLBP1 or lives as a separate item.)

docs/JLBP-21.md
elharo: added a link to the Maven page.

saturnism:
I'm leaning towards JLBP-1, where we call out annotation processors not to be included as a direct dependency that gets exposed as a transitive dependency (or, any compile-only dependencies that shouldn't be exposed).

On the other hand, this proposal may have more information; perhaps it can be linked.  I feel current proposal seems to mix classloading (last paragraph) and annotation processors / compile time dependencies.

garrettjonesgoogle:
I lean slightly towards adding to JLBP-1, but it's not a hill I would die on. I lean towards it because this new content references AutoValue, which JLBP-1 also references (for essentially the same basic idea).

docs/JLBP-21.md
```diff
@@ -0,0 +1,32 @@
+# [JLBP-21] Separate tool classpath from product classpath
+
+The tools we use for Java projects are more often than not themselves
+written in Java. This includes build system such as Maven, javac, and Ant;
+runtime environments such as Tomcat, Hadoop, Beam, and App Engine; and editors
+such as jEdit, Eclipse, and IntelliJ IDEA. Tools like these should
+have their own classpaths from which they load their own code.
+
+It is critical not to mix the classpath of the tool with the classpath of the product.
+Dependencies of the tool should not be dependencies of the product.
```
garrettjonesgoogle: This line ("Dependencies of the tool...") seems redundant with the prior sentence.

elharo: It is, I suppose; but sometimes it's helpful to say the same thing in different ways to get the point across.

garrettjonesgoogle: This contradicts all of your prior edit suggestions on the JLBPs to ensure conciseness...

elharo:
seems like rough consensus is to include this in JLBP-1

PR #1012, opened 2019-10-31
Altternate version of JLBP-21

elharo:
This is an alternate vision for #999 that includes this content in JLBP-1 instead of a new JBLP-21. Please indicate which you prefer.


saturnism:
+1 I feel we can use concrete examples for an annotation processor here (e.g., autovalue)

saturnism:
Would it make sense to title the section something like? E.g., "Do not exposse annotation processors and annotations not needed at runtime in the transitive dependency"

And then giving examples of that? I.e., in most cases, we are dealing w/ Maven/Gradle and not javac. This will help codify concrete issue/solution/practice.

elharo:
The premise is much more general than that though. The annotations are just some recent examples we've found. The Hadoop issues, for instance, had nothing to do with annotations.

PR #1015, opened 2019-11-06
GCP Libraries BOM

elharo:
instead of Google Libraries BOM
@suztomo


suztomo:
(I might miss the discussion.) I prefer to use "Google Libraries BOM". "GCP Libraries BOM" reflects the groupId "com.google.cloud". But it obscures the different roles between libraries-bom and google-cloud-bom.

With "Google Libraries BOM", I thought we target wider libraries than just GCP.

PR #1021, opened 2019-11-06
Removed version range query from Linkage Monitor

### LINKED ISSUE #1020: Linkage Monitor to check the versions of artifacts in current working directory
As of now Linkage Monitor checks the version of artifacts through local repository. It picks up the highest snapshot version to be the one that just installed locally. But this method is not reliable if there is a higher snapshot version in Maven Central or a repository does not update version after a release (`1.0-SNAPSHOT` is lower than `1.0`).

If Linkage Monitor can check the versions of artifacts (including BOM) in current working directory, then it does not need to rely on local Maven repository, and thus can avoid using the convention of "highest snapshot version".

suztomo:
Fixes #1020 .

With this PR, Linkage Monitor checks current working directory rather than Maven repository. The directory where the tool runs matters; there should not be any `cd` after `mvn install` before Linkage Monitor runs:

- google-cloud-java: OK: https://github.com/googleapis/google-cloud-java/blob/master/.kokoro/linkage-monitor.sh#L33
- google-http-client: OK: https://github.com/googleapis/google-http-java-client/blob/master/.kokoro/linkage-monitor.sh#L33
- protobuf: OK: https://github.com/protocolbuffers/protobuf/blob/master/tests.sh#L271

### DIFF
--- a/linkage-monitor/src/test/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorTest.java
+++ b/linkage-monitor/src/test/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorTest.java
@@ -17,7 +17,6 @@
 package com.google.cloud.tools.dependencies.linkagemonitor;
 
 import static com.google.cloud.tools.opensource.dependencies.Artifacts.toCoordinates;
-import static com.google.cloud.tools.opensource.dependencies.RepositoryUtility.CENTRAL;
 import static com.google.common.collect.Iterables.skip;
 import static com.google.common.truth.Correspondence.transforming;
 import static com.google.common.truth.Truth.assertWithMessage;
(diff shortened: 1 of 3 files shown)


linkage-monitor/src/test/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorTest.java
```diff
@@ -37,142 +33,50 @@
...
 
   @Before
-  public void setup() throws IOException {
+  public void setup() {
     system = RepositoryUtility.newRepositorySystem();
 
-    // If possible, spy object should be avoided. But Maven is tightly coupled with RepositorySystem
-    // and thus normal mock objects on RepositorySystem would make the test even complicated.
-    // https://static.javadoc.io/org.mockito/mockito-core/3.0.0/org/mockito/Mockito.html#spy-T-
-    spySystem = spy(system);
```
suztomo: This no longer needs spy object.

linkage-monitor/src/test/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorTest.java
```diff
@@ -279,39 +184,18 @@ public void testBuildModelWithSnapshotBom_invalidCoordinates()
...
   @Test
   public void testBuildModelWithSnapshotBom_BomSnapshotUpdate()
-      throws MavenRepositoryException, ModelBuildingException, ArtifactResolutionException,
-          InvalidVersionSpecificationException, VersionRangeResolutionException {
+      throws ModelBuildingException, ArtifactResolutionException {
     // Linkage Monitor should update a BOM in Google Cloud Libraries BOM when it's available local
     // repository. This test case simulates the issue below where
     // google-cloud-bom:0.106.0-alpha-SNAPSHOT should provide gax:1.48.0.
     // https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/853
-
```
suztomo: These complex test setup is now gone.

PR #1026, opened 2019-11-07
Linkage Checker Enforcer Rule Release 1.0.1 

suztomo:
The enforcer rule now works with artifacts having `${os.detect.classifier}` in their dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Linkage Checker Enforcer Rule Change Log
-All notable changes to this project will be documented in this file.
 
 ## 1.0.1
 * The enforcer rule now works with artifacts having `${os.detect.classifier}` in their dependencies.
+  When such artifacts were resolved without [os-maven-plugin](https://github.com/trustin/os-maven-plugin)
+  and cached in Maven, the enforcer rule could not resolve the dependency tree.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Linkage Checker Enforcer Rule Change Log
+All notable changes to this project will be documented in this file.
+
+## 1.0.1
+* The enforcer rule now works with artifacts having `${os.detect.classifier}` in their dependencies.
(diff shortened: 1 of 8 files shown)


suztomo:
@elharo PTAL. Added CHANGELOG.md.

CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Linkage Checker Enforcer Rule Change Log
+All notable changes to this project will be documented in this file.
```
elharo: this line isn't needed

CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
...
+
+## 1.0.1
+* The enforcer rule now works with artifacts having `${os.detect.classifier}` in their dependencies.
```
elharo: Can you elaborate on this a little, perhaps link to the definition of this property in the Maven docs?

suztomo: Added explanation with a link to os-maven-plugin that defines the variable.

CHANGELOG.md
suztomo: Removed.

suztomo:
It's been pushed. The added explanation is the next line of the comment.

CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Linkage Checker Enforcer Rule Change Log
+
+## 1.0.1
+* The enforcer rule now works with artifacts having `${os.detect.classifier}` in their dependencies.
+  When such artifacts were resolved without [os-maven-plugin](https://github.com/trustin/os-maven-plugin)
```
suztomo: This is added explanation.

elharo: How about simply, "The enforcer rule now recognizes the `${os.detect.classifier}` property defined by the [os-maven-plugin](https://github.com/trustin/os-maven-plugin)."

That said, I wonder if we have a much bigger issue here. The current rule is is just patching over this one special case. There are many more cases involving os-maven-plugin and potentially other plugins where we are not resolving dependencies in the same way Maven does.

suztomo: Updated the changelog. Yes, there may be other cases. So far, only os-maven-plugin. I'm thinking to handle them in an ad-hoc manner.

PR #1038, opened 2019-11-14
Linkage monitor test not to rely on the linkage-monitor or root projects

### LINKED ISSUE #1035: System.out.println in LinkageMonitor
Can this really be ignored? If so, log at most. But maybe it can't be. Seems to be correlated with a test failure right now. 

```
      try {
        ModelBuildingResult modelBuildingResult = modelBuilder.build(modelRequest);
        Model model = modelBuildingResult.getEffectiveModel();
        artifactToVersion.put(model.getGroupId() + ":" + model.getArtifactId(), model.getVersion());
      } catch (ModelBuildingException ex) {
        // When there's a pom.xml file that is not (indirectly) referenced by the root pom.xml,
        // Maven may fail to build the model. Such pom.xml can be ignored.
        System.out.println("Ignoring bad model: " + path);
      }
```

suztomo:
Fixes #1035 . When I created the test I didn't think about "target" directory would contain invalid pom.xml. Fixing the test not to rely on the linkage-monitor directory or the root project.


linkage-monitor/src/test/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorTest.java
```diff
@@ -222,25 +222,14 @@ public void testBuildModelWithSnapshotBom_JdkVersionActivation()
...
 
-    Truth.assertThat(localArtifacts).hasSize(1);
-    Truth.assertThat(localArtifacts).containsKey("com.google.cloud.tools:linkage-monitor");
-  }
-
-  @Test
-  public void testFindLocalArtifacts_RootDirectory() {
-    // Root of cloud-opensource-java has more than 10 pom.xml files
-    ImmutableMap<String, String> localArtifacts =
-        LinkageMonitor.findLocalArtifacts(system, session, Paths.get(".."));
```
suztomo: This test was reading all `pom.xml` files under cloud-opensource-java directory.

linkage-monitor/src/test/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorTest.java
```diff
@@ -222,25 +222,14 @@ public void testBuildModelWithSnapshotBom_JdkVersionActivation()
   }
 
   @Test
-  public void testFindLocalArtifacts_currentDirectory() {
+  public void testFindLocalArtifacts() {
     // Current working directory of linkage-monitor should have one linkage monitor artifact
     ImmutableMap<String, String> localArtifacts =
-        LinkageMonitor.findLocalArtifacts(system, session, Paths.get("."));
+        LinkageMonitor.findLocalArtifacts(
+            system, session, Paths.get("src/test/resources/testproject"));
```
suztomo: Now this test only relies on the dummy directory containing the pom.xml files (added as part of this PR).

suztomo:
@elharo Would you confirm this enhancement resolve the flaky behavior? Branch linkage-monitor-test

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -133,7 +133,7 @@ public static void main(String[] arguments)
       } catch (ModelBuildingException ex) {
         // When there's a pom.xml file that is not (indirectly) referenced by the root pom.xml,
         // Maven may fail to build the model. Such pom.xml can be ignored.
-        System.out.println("Ignoring bad model: " + path);
+        System.out.println("Ignoring bad model: " + path + ": " + ex.getMessage());
```
elharo: do we want to System.out.println here? A logger would be better, or perhaps this case shouldn't be ignored.

suztomo: Now it uses logger.

suztomo: About ignoring this failure, the output in the comment below shows that target directory may contain pom.xml files. Linkage Monitor should not be stopped by them.
https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1035#issuecomment-553846199

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -133,7 +136,7 @@ public static void main(String[] arguments)
       } catch (ModelBuildingException ex) {
         // When there's a pom.xml file that is not (indirectly) referenced by the root pom.xml,
         // Maven may fail to build the model. Such pom.xml can be ignored.
-        System.out.println("Ignoring bad model: " + path);
+        logger.warning("Ignoring bad model: " + path + ": " + ex.getMessage());
```
elharo: should this even be a warning? Do we expect the user to act on it? If not, we don't need to notify them.

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -133,7 +136,7 @@ public static void main(String[] arguments)
       } catch (ModelBuildingException ex) {
         // When there's a pom.xml file that is not (indirectly) referenced by the root pom.xml,
```
elharo: I'm not sure what "not (indirectly)" means. Is this the same as "not" or a double negative? Can you clarify this at all?

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
suztomo: Lowered to info. This information of un-build-able pom.xml is useful when troubleshooting.

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
suztomo: Updated the comment.

linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -131,9 +134,9 @@ public static void main(String[] arguments)
         Model model = modelBuildingResult.getEffectiveModel();
         artifactToVersion.put(model.getGroupId() + ":" + model.getArtifactId(), model.getVersion());
       } catch (ModelBuildingException ex) {
-        // When there's a pom.xml file that is not (indirectly) referenced by the root pom.xml,
-        // Maven may fail to build the model. Such pom.xml can be ignored.
-        System.out.println("Ignoring bad model: " + path);
+        // Maven may fail to build pom.xml files found in irrelevant directories, such as "target"
```
elharo: should we deliberately and explicitly skip those files?

suztomo: I'd keep the logic simple until we need it.

PR #1097, opened 2020-01-10
Make DependencyGraphBuilder nonstatic to hold repositories

suztomo:
Make `DependencyGraphBuilder` and `ClassPathBuilder` non-static because
- `DependencyGraphBuilder` can hold `repositories` in its instance variable (rather than a global variable of RepositoryUtility)
- `ClassPathBuilder` to hold `DependencyGraphBuilder` instance

Removing `AggregatedRepositoryException` in favor of `ArtifactProblem`.


suztomo:
@elharo Although still in draft, what do you think about the idea of making DependencyGraphBuilder's methods non-static so that it can hold ArtifactProblem list in its instance?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilder.java
```diff
@@ -74,7 +77,7 @@
    * @return an ordered map of absolute paths of jar files to one or more Maven dependency paths
    * @throws RepositoryException when there is a problem retrieving jar files
    */
-  public static LinkedListMultimap<Path, DependencyPath> artifactsToDependencyPaths(
+  public LinkedListMultimap<Path, DependencyPath> artifactsToDependencyPaths(
```
elharo: why would this method need to change?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -57,13 +57,15 @@
...
+  private final List<UnresolvableArtifactProblem> artifactProblems = new ArrayList<>();
+
   static {
     detectOsProperties().forEach(System::setProperty);
   }
 
   // caching cuts time by about a factor of 4. Dependency class's equality includes exclusions.
-  private static final Map<Dependency, DependencyNode> cacheWithProvidedScope = new HashMap<>();
-  private static final Map<Dependency, DependencyNode> cacheWithoutProvidedScope = new HashMap<>();
+  private static Map<Dependency, DependencyNode> cacheWithProvidedScope = new HashMap<>();
```
elharo: why not final?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/UpdateReport.java
```diff
@@ -40,9 +39,9 @@ public static void main(String[] args) throws RepositoryException {
     
     try {
       DefaultArtifact artifact = new DefaultArtifact(args[0]);
-      
-      DependencyGraph graph =
-          DependencyGraphBuilder.getCompleteDependencies(artifact);
+
+      DependencyGraphBuilder dependencyGraphBuilder = new DependencyGraphBuilder();
```
elharo: There's no configuration here, which suggests this should still be static.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -57,13 +57,15 @@
   private static final CharMatcher LOWER_ALPHA_NUMERIC =
       CharMatcher.inRange('a', 'z').or(CharMatcher.inRange('0', '9'));
 
+  private final List<UnresolvableArtifactProblem> artifactProblems = new ArrayList<>();
```
suztomo: @elharo This is the reason to make DependencyGraphBuilder non-static. Once DependencyGraphBuilder becomes non-static, then ClassPathBuilder also becomes non-static to hold the instance of DependencyGraphBuilder.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilder.java
suztomo: The `levelOrder` method needs to be non-static to touch non-static `artifactProblems`. Then all other methods that may call levelOrder becomes non-static.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: This is typo.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/UpdateReport.java
suztomo: The reason of making this non-static is to hold internal state `artifactProblems`.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Having static `artifactProblem` (a global variable) would cause a problem when we call DependencyGraphBuilder to build a graph multiple times. We cannot distinguish which ArtfiactProblems correspond to which graph building invocation.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -336,17 +335,13 @@ private static DependencyGraph levelOrder(
               DependencyNode failedDependencyNode = artifactResult.getRequest().getDependencyNode();
               ExceptionAndPath failure =
                   ExceptionAndPath.create(parentNodes, failedDependencyNode, resolutionException);
-              if (requiredDependency(failure.getPath())) {
-                resolutionFailures.add(failure);
-              }
+              artifactProblems.add(new UnresolvableArtifactProblem(failure.getPath()));
```
suztomo: LevelOrder touching artifactProblems requires the method to be non-static.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerMain.java
```diff
@@ -44,12 +46,21 @@ public static void main(String[] arguments)
     LinkageCheckerArguments linkageCheckerArguments =
         LinkageCheckerArguments.readCommandLine(arguments);
 
-    RepositoryUtility.setRepositories(linkageCheckerArguments.getExtraMavenRepositoryUrls(),
-        linkageCheckerArguments.getAddMavenCentral());
+    DependencyGraphBuilder dependencyGraphBuilder =
+        new DependencyGraphBuilder(
+            linkageCheckerArguments.getExtraMavenRepositoryUrls(),
+            linkageCheckerArguments.getAddMavenCentral());
```
suztomo: In this idea, DependencyGraphBuilder takes repository URL configuration.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -108,20 +115,42 @@ private static String osDetectedArch() {
...
   }
 
-  private static DependencyNode resolveCompileTimeDependencies(
-      DependencyNode rootDependencyArtifact)
+  /**
+   * @param mavenRepositoryUrls Maven repository URLs to search for dependencies
+   * @param addMavenCentral if true, add Maven Central to the end of the repository list
+   * @throws IllegalArgumentException if a URL is malformed or not having an allowed scheme
+   */
+  public DependencyGraphBuilder(Iterable<String> mavenRepositoryUrls, boolean addMavenCentral) {
```
suztomo: DependencyGraphBuilder to take configuration of repository URLs. As of now, a static field of `RepositoryUtility` holds the repository URLs. I believe they should be an instance variable of DependencyGraphBuilder.

From old comment on the static mutable field in by Elliotte:

> Mutable static state suggests we're growing past the point where this class can plausibly be static. OK, for now but consider whether refactoring might help. Either to use more instance methods or to pass the repsositoryURLs as arguments to a method instead.
https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/349/files#r249024802

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -57,13 +60,17 @@
   private static final CharMatcher LOWER_ALPHA_NUMERIC =
       CharMatcher.inRange('a', 'z').or(CharMatcher.inRange('0', '9'));
 
+  private final List<UnresolvableArtifactProblem> artifactProblems = new ArrayList<>();
+
+  private final ImmutableList<RemoteRepository> repositories;
```
suztomo: Repository URLs are the configuration of DependencyGraphBuilder.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -90,8 +90,6 @@
   public static final RemoteRepository CENTRAL =
       new RemoteRepository.Builder("central", "default", "http://repo1.maven.org/maven2/").build();
 
-  private static ImmutableList<RemoteRepository> mavenRepositories = ImmutableList.of(CENTRAL);
```
suztomo: It was a mutable static variable.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -289,34 +296,6 @@ public static boolean shouldSkipBomMember(Artifact artifact) {
...
-
-    mavenRepositories = repositoryListBuilder.build();
-  }
-
-  /** Adds {@link #mavenRepositories} to {@code collectRequest}. */
-  public static void addRepositoriesToRequest(CollectRequest collectRequest) {
-    for (RemoteRepository repository : mavenRepositories) {
-      collectRequest.addRepository(repository);
-    }
-  }
```
suztomo: Now we don't need to have these methods that use `mavenRepositories`, a mutable global variable.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerArgumentsTest.java
```diff
@@ -16,29 +16,14 @@
...
-import org.junit.After;
 import org.junit.Assert;
 import org.junit.Test;
 
 public class LinkageCheckerArgumentsTest {
 
-  @After
-  public void cleanup() {
-    // Resets the effect of setRepositories
-    RepositoryUtility.setRepositories(ImmutableList.of(), true);
```
suztomo: No need to cleanup the mutable global variable.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerArgumentsTest.java
```diff
@@ -89,42 +74,6 @@ public void parseCommandLineArguments_invalidOption() {
...
-    // Spring-asm is used here because it does not have complex dependencies
-    Artifact artifact = new DefaultArtifact("androidx.lifecycle:lifecycle-common-java8:2.0.0");
-
-    List<Path> paths = ClassPathBuilder.artifactsToClasspath(ImmutableList.of(artifact));
-    Truth.assertThat(paths).isNotEmpty();
-  }
-
-  @Test
-  public void testConfigureAdditionalMavenRepositories_notToUseMavenCentral()
-      throws ParseException {
```
suztomo: These tests have moved to DependencyGraphBuilderTest.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Fixed.

suztomo:
@elharo PTAL. Now this is ready for review.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
elharo: I agree. That's a reason to not make it a field, static or instance. `artifactProblems` is invocation state, not object state. This can cause problems even in single threaded environments.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
elharo: This one does make sense as a field.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -108,20 +115,41 @@ private static String osDetectedArch() {
...
 
-  private static DependencyNode resolveCompileTimeDependencies(
-      DependencyNode rootDependencyArtifact)
+  public DependencyGraphBuilder() {
+    this(ImmutableList.of(CENTRAL.getUrl()));
+  }
+
+  /**
+   * @param mavenRepositoryUrls Maven repository URLs to search for dependencies
+   * @throws IllegalArgumentException if a URL is malformed or not having an allowed scheme
```
elharo: not having --> does not have

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -108,20 +115,41 @@ private static String osDetectedArch() {
...
+  public DependencyGraphBuilder(Iterable<String> mavenRepositoryUrls) {
+    ImmutableList.Builder<RemoteRepository> repositoryListBuilder = ImmutableList.builder();
+    for (String mavenRepositoryUrl : mavenRepositoryUrls) {
+      RemoteRepository repository = mavenRepositoryFromUrl(mavenRepositoryUrl);
+      repositoryListBuilder.add(repository);
+    }
+    this.repositories = repositoryListBuilder.build();
+  }
+
+  /** Returns unresolved artifact problems encountered during constructing a dependency graph. */
```
elharo: during constructing a --> while constructing the

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyLister.java
```diff
@@ -33,10 +32,10 @@ public static void main(String[] args) throws RepositoryException {
     
     try {
       DefaultArtifact artifact = new DefaultArtifact(args[0]);
-    
-      DependencyGraph graph =
-          DependencyGraphBuilder.getCompleteDependencies(artifact);
-      
+
+      DependencyGraphBuilder dependencyGraphBuilder = new DependencyGraphBuilder();
```
elharo: This object can only be used once. The instance state makes it not reusable. I suspect what we really need here is a static method that uses private classes and objects to keep track of the invocation state. However none of that should be exposed in the published API. 

And in any case, that's just an implementation detail. Looking at this, what you have is a static method. Everything that's being stored in the fields could be passed along through method arguments from one method call to the next.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/UpdateReport.java
elharo: Implementation detail. This should not change the public API.

suztomo:
I'll need more clarification. I will talk to you later today.

----
## artifactProblems
I like the idea of `artifactProblems` is per method invocation. I guess solution would be

- creating a class that holds both dependency graph and artifactProblems, or
- creating an exception that carries both artifactProblems and a dependency graph

The caller's code would be simpler if we choose the former.

## repositories
For DependencyGraphBuilder's field `ImmutableList<RemoteRepository> repositories`, this is the need for making DependencyGraphBuilder non-static even for public API; otherwise all the static methods would carry `repositories` arguments.

The `repositories` is just one configuration parameter as of now. In future, the caller of DependencyGraphBuilder may want to pass `managedDependencies` or an option to use Gradle's algorithm. DependencyGraphBuilder's constructor taking these parameters seems reasonable abstraction for them.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Let's remove this field.

suztomo:
I'll make DependencyGraphBuilder non static and create a class that holds dependency graph and artifactproblems.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: This field has been removed

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Fixed.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
suztomo: Fixed in DependencyGraphResult.getArtifactProblems.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerMain.java
```diff
@@ -44,12 +47,20 @@ public static void main(String[] arguments)
     LinkageCheckerArguments linkageCheckerArguments =
         LinkageCheckerArguments.readCommandLine(arguments);
 
-    RepositoryUtility.setRepositories(linkageCheckerArguments.getExtraMavenRepositoryUrls(),
-        linkageCheckerArguments.getAddMavenCentral());
+    DependencyGraphBuilder dependencyGraphBuilder =
+        new DependencyGraphBuilder(linkageCheckerArguments.getMavenRepositoryUrls());
```
suztomo: LinkageCheckerMain configures DependencyGraphBuilder's repository URLs.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphResult.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package com.google.cloud.tools.opensource.dependencies;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.collect.ImmutableList;
+
+/** Result of dependency graph building with {@link UnresolvableArtifactProblem}s if any. */
+public final class DependencyGraphResult {
```
suztomo: New class to hold dependency graph and artifact problems.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -229,7 +220,7 @@ private DependencyNode resolveCompileTimeDependencies(
    * @return dependency graph representing the tree of Maven artifacts
    * @throws RepositoryException when there is a problem in resolving or collecting dependency
```
elharo: in resolving or collecting dependency --> resolving or collecting dependencies

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphResult.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package com.google.cloud.tools.opensource.dependencies;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.google.common.collect.ImmutableList;
+
+/** Result of dependency graph building with {@link UnresolvableArtifactProblem}s if any. */
+final public class DependencyGraphResult {
```
elharo: public final (or just final if it doesn't need to be public)

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphResult.java
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+  public DependencyGraph getDependencyGraph() {
+    return dependencyGraph;
+  }
+
+  /**
+   * Returns problems encountered while constructing the dependency graph.
+   */
+  ImmutableList<UnresolvableArtifactProblem> getArtifactProblems() {
```
elharo: We don't seem to use this method except in tests. What am I missing?

PR #1104, opened 2020-01-13
Configuring renovate bot

suztomo:
- Ignore example directory. The usage of "**" is in the default list. https://github.com/renovatebot/renovate/blob/master/renovate-schema.json#L300
- "sematic commit" to follow google-cloud-java https://github.com/googleapis/java-cloud-bom/blob/master/renovate.json#L63 . This is for the commit message.
- separateMajorMinor not to distinguish libraries-bom 2.X and 3.X.
- Guava's version scheme needs to distinguish "-jre" and "-android".
  Using https://github.com/renovatebot/config-help/issues/232#issuecomment-491826549 and https://github.com/googleapis/java-cloud-bom/blob/master/renovate.json#L15
- `ignoreUnstable: false` for google-cloud artifacts because its version carries "-alpha"


suztomo:
@elharo PTAL.

suztomo:
Okay. Closed the PRs by Renovate bot for now. We need renovate bot to confirm the configuration is correct.

@elharo  Which files and dependencies you think Renovate bot should not touch in addition to this PR?

renovate.json
```diff
@@ -1,5 +1,22 @@
 {
   "extends": [
     "config:base"
+  ],
+  "ignorePaths": ["example-problems/**"],
```
suztomo: Memo: do we need to exclude "boms/cloud-oss-bom/pom.xml"?

chingor13: The cloud-oss-bom is probably the most important one for renovate to update

suztomo:
boms/cloud-oss-bom/pom.xml?
There may be a problem in a new version. Elliotte wants two humans to confirm the change in the BOM member. I'll exclude `boms/cloud-oss-bom/pom.xml`.


We don't care boms/convergence-check/pom.xml.

Ignore anything that with "-snapshot" 
```
      <groupId>com.google.cloud.tools</groupId>
      <artifactId>dependencies</artifactId>
```

.circleci/config.yml and docs/Gemfile do not have version. No need to worry.

renovate.json
suztomo: Elliotte prefers to maintain it manually; he wants 2 humans involved before modifying the content of the BOM.

renovate.json
```diff
@@ -0,0 +1,35 @@
...
+      "ignoreUnstable": false
+    },
+    {
+      "packagePatterns": [
+        "^com.google.cloud:libraries-bom",
+        "^com.google.cloud.tools:dependencies",
+        "^com.google.cloud.tools:linkage",
+        "^com.google.cloud.tools.opensource"
+      ],
+      "enabled": false
```
suztomo: They are the ones in this repository. Not enabling them.

renovate.json
```diff
@@ -0,0 +1,35 @@
+{
+  "extends": [
+    "config:base"
+  ],
+  "ignorePaths": [
+    "example-problems/**",
+    "boms/cloud-oss-bom/pom.xml",
+    "boms/convergence-check/pom.xml"
```
elharo: I think you need to exclude boms/upper-bounds-check/ too.

suztomo: Updated this to ignore "boms/**".

PR #1127, opened 2020-01-15
Linkage Checker Enforcer Rule in the example no-such-method-error-signature-mismatch

suztomo:


example-problems/no-such-method-error-signature-mismatch/README.md
```diff
@@ -10,7 +10,7 @@ Config included in a DNS record][1].
 
 ## How to run
 
-Run `mvn exec:java` to reproduce the issue.
+Run `mvn compile exec:java` to reproduce the issue.
```
elharo: Is this now just mvn verify? Does exec:java imply compile?

suztomo: This step is to reproduce the runtime error. Not for the enforcer rule.

No, "exec" does not compile.

PR #1134, opened 2020-01-16
libraries-bom usage docs for end users

### LINKED ISSUE #21: BOM usage docs for End users
i.e. devs who simply import the libraries

elharo:
fixes #21 

Focus on using libraries-bom.

For now this simply publishes the doc on jlbp.dev. We can figure out where to link it from later. 

@garrettjonesgoogle


docs/DECLARING_DEPENDENCIES.md
```diff
@@ -124,55 +120,6 @@ For an example, see [gax-java#690](https://github.com/googleapis/gax-java/pull/6
 
 See the "Choosing BOMs" section below if you need to import more than one BOM.
 
-## Choosing BOMs
```
saturnism: this section has pretty useful information; can we keep?

docs/DECLARING_DEPENDENCIES.md
```diff
@@ -201,10 +148,3 @@ library ecosystem:
 - Each library follows semantic versioning. This means that once a
   library reaches a 1.x version, features and APIs can be added in minor/patch
   releases but not removed within a major version.
-
-This combination of characteristics means that you can generally avoid
```
saturnism: this is also valid, any reason for removal?

docs/DECLARING_DEPENDENCIES.md
elharo: This is from earlier days before we had a working libraries-bom. On of our OKRs is to make libraries-bom the standard means of consuming these libraries. Developers shouldn't need to think about which BOMs to pick when. Just use libraries-bom is a much simpler and easier to grok story. 

It might be worth adding a section about combining BOMs from multiple projects; e.g. libraries-bom and spring BOM, but that's not what this is about and would be a subject for a separate PR.

docs/DECLARING_DEPENDENCIES.md
elharo: On rereading this, I found it hard to follow and perhaps a bit tautological. #2 needs to go because we want devs to simply import libraries-bom, not multiple BOMs, and #1 needs to go because developers shouldn't need to think about the versions of individual dependencies, just the BOM.

docs/DECLARING_DEPENDENCIES.md
saturnism: my thought is that there is a difference btwn best practices for java libraries in terms of how to produce and consume boms, vs. preferred method of consumption for GCP client libraries; The latter is in the c.g.c documentation.

It's still super useful to keep the list of the boms, and also adding libraries bom to the list.  Each individual BOM can be used to override 1 component version as a whole. E.g., in situations where gRPC is causing trouble, simply importing a gRPC BOM would solve the issue.

docs/DECLARING_DEPENDENCIES.md
saturnism: ack; thnx for the info, i feel that makes sense too.

docs/DECLARING_DEPENDENCIES.md
elharo: That's a very special case though. It does happen. There's an issue in java-storage right now where people are doing that with the http-java-client, but I'm not sure as written this correctly describes that situation. I want to give a very simple message here:

Use libraries-bom.

That perhaps sometimes  specifies more versions than the users technically needs, but in that case no harm done. It is so much easier for users to consume a simple message like this thna a long, convoluted message of all the possible special cases.

garrettjonesgoogle: I agree with @saturnism .

We should still include the extended message, but we could also "offload" it to a separate page and provide a link to it, so people who do need those details can get to it.

docs/DECLARING_DEPENDENCIES.md
```diff
@@ -0,0 +1,204 @@
...
+
+`com.google.cloud:libraries-bom` also covers several other libraries
+that client code does not usually depend on directly. However if you do import
+any of these, use  `com.google.cloud:libraries-bom` to specify their versions
+too:
+
+- [GAX Google API Extensions for Java](https://github.com/googleapis/gax-java)
+- [Google Auth Library](https://github.com/googleapis/google-auth-library-java)
+- [Google API Client Library for Java](https://github.com/googleapis/google-api-java-client)
+- [Grafeas Artifact Metadata API](https://github.com/grafeas/grafeas)
```
suztomo:
```suggestion
- [Grafeas Artifact Metadata API](https://github.com/googleapis/java-grafeas)
```

as per https://search.maven.org/artifact/io.grafeas/grafeas/0.24.0/jar

docs/DECLARING_DEPENDENCIES.md
```diff
@@ -0,0 +1,204 @@
...
+dependency tree.
+
+When a library publishes multiple artifacts, using different versions
+for different artifacts may cause dependency conflicts. To fix this
+problem, you can import a BOM to force consistent versions, as long as
+you are using at least Gradle 4.6. To do this:
+
+- Turn on BOM support:
+  - If you are using Gradle 4.x:
+
```
suztomo: Case of Gradle 4.x is missing.

docs/DECLARING_DEPENDENCIES.md
```diff
@@ -0,0 +1,204 @@
...
+- [Grafeas Artifact Metadata API](https://github.com/grafeas/grafeas)
+
+## Ensuring Compatibility
+
+To ensure that your own project uses compatible versions of these
+libraries, follow the guidance below for your build system.
+
+### Maven
+
+#### Use the requireUpperBoundDeps enforcer rule
```
suztomo: Shall we put "Import BOMs to ensure dependencies are consistent" first, because the purpose of this document is to encourage users to import the BOM.

docs/DECLARING_DEPENDENCIES.md
```diff
@@ -0,0 +1,204 @@
...
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+```
+
+When the newest versions are not compatible, you can use BOMs to
+select compatible versions.
+
+#### Import BOMs to ensure dependencies are consistent
```
suztomo:
```suggestion
#### Import GCP Libraries BOM to ensure dependencies are consistent
```

This section "Ensuring Compatibility" is written for generic usage of BOMs (not specific to libraries-bom), while this document is "Declaring Dependencies on Google Cloud Java Libraries". Is this intentional?

docs/DECLARING_DEPENDENCIES.md
elharo: done

docs/DECLARING_DEPENDENCIES.md
elharo: done

docs/DECLARING_DEPENDENCIES.md
elharo: done

docs/DECLARING_DEPENDENCIES.md
elharo: done

docs/DECLARING_DEPENDENCIES.md
elharo: done

PR #983, opened 2019-10-23
Linkage monitor error to include dependency path to problematic artifacts

suztomo:
Toward #981 . Option 2: "Show Path to the Problematic Artifact" in the issue description. Not showing scope (yet).

With this PR, when Linkage Monitor shows error, it tells the location of the problematic Maven artifact in the dependency tree.

Example case to show gwt-user-2.8.2.jar:
```
suztomo@suxtomo24:~/cloud-opensource-java/linkage-monitor$ java -jar target/linkage-monitor-1.0.1-SNAPSHOT-all-deps.jar   com.google.cloud:libraries-bom
BOM Coordinates: com.google.cloud:libraries-bom:2.7.1-SNAPSHOT
Newly introduced problems:
...
Class org.objectweb.asm.commons.Method is not found
  referenced from com.google.web.bindery.requestfactory.server.RequestFactoryJarExtractor (gwt-user-2.8.2.jar)
Class com.google.gwt.dev.util.Name$SourceOrBinaryName is not found
  referenced from com.google.web.bindery.requestfactory.server.RequestFactoryJarExtractor (gwt-user-2.8.2.jar)
Class org.hibernate.validator.engine.ConstraintViolationImpl is not found
  referenced from org.hibernate.validator.engine.ConstraintViolationImpl_CustomFieldSerializer (gwt-user-2.8.2.jar)
Class org.hibernate.validator.engine.PathImpl is not found
  referenced from org.hibernate.validator.engine.PathImpl_CustomFieldSerializer (gwt-user-2.8.2.jar)
gwt-user-2.8.2.jar is at:
  com.google.cloud:google-cloud-firestore:1.28.1-SNAPSHOT / com.google.truth:truth:1.0 / com.google.gwt:gwt-user:2.8.2

Exception in thread "main" com.google.cloud.tools.dependencies.linkagemonitor.LinkageMonitorException: Found 278 new linkage errors
	at com.google.cloud.tools.dependencies.linkagemonitor.LinkageMonitor.run(LinkageMonitor.java:137)
	at com.google.cloud.tools.dependencies.linkagemonitor.LinkageMonitor.main(LinkageMonitor.java:87)

```

PR #1152, opened 2020-01-24
LinkageCheckerMain to report dependency paths

suztomo:
Applying #981 (and #983) for LinkageCheckerMain, which I use for Apache Beam.

Example output:
```
beam-sdks-java-core-2.20.0-SNAPSHOT.jar is at:
  org.apache.beam:beam-sdks-java-io-hadoop-format:2.20.0-SNAPSHOT (compile) / org.apache.beam:beam-sdks-java-core:2.20.0-SNAPSHOT (compile)
  and 1 dependency path.
beam-vendor-bytebuddy-1_9_3-0.1.jar is at:
  org.apache.beam:beam-sdks-java-io-hadoop-format:2.20.0-SNAPSHOT (compile) / org.apache.beam:beam-sdks-java-core:2.20.0-SNAPSHOT (compile) / org.apache.beam:beam-vendor-bytebuddy-1_9_3:0.1 (compile)
  and 1 dependency path.
jackson-xc-1.8.3.jar is at:
  org.apache.beam:beam-sdks-java-io-hadoop-format:2.20.0-SNAPSHOT (compile) / org.apache.hadoop:hadoop-common:2.8.5 (provided) / com.sun.jersey:jersey-json:1.9 (compile) / org.codehaus.jackson:jackson-xc:1.8.3 (compile)
  and 125 other dependency paths.
jackson-mapper-asl-1.9.13.jar is at:
  org.apache.beam:beam-sdks-java-io-hadoop-format:2.20.0-SNAPSHOT (compile) / org.apache.hadoop:hadoop-common:2.8.5 (provided) / org.codehaus.jackson:jackson-mapper-asl:1.9.13 (compile)
  and 177 other dependency paths.
weld-osgi-bundle-1.1.0.Final.jar is at:
  org.apache.beam:beam-sdks-java-io-hadoop-format:2.20.0-SNAPSHOT (compile) / org.apache.hadoop:hadoop-common:2.8.5 (provided) / com.sun.jersey:jersey-server:1.9 (compile) / org.jboss.weld:weld-osgi-bundle:1.1.0.Final (provided)
  and 151 other dependency paths.
guava-11.0.2.jar is at:
  org.apache.beam:beam-sdks-java-io-hadoop-format:2.20.0-SNAPSHOT (compile) / org.apache.hadoop:hadoop-common:2.8.5 (provided) / com.google.guava:guava:11.0.2 (compile)
  and 194 other dependency paths.
apacheds-kerberos-codec-2.0.0-M15.jar is at:
  org.apache.beam:beam-sdks-java-io-hadoop-format:2.20.0-SNAPSHOT (compile) / org.apache.hadoop:hadoop-common:2.8.5 (provided) / org.apache.hadoop:hadoop-auth:2.8.5 (compile) / org.apache.directory.server:apacheds-kerberos-codec:2.0.0-M15 (compile)
  and 100 other dependency paths.

```

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
@@ -86,7 +86,7 @@ public final class ClassPathResult {
       if (otherCount == 1) {
         message.append("  and 1 dependency path.\n");
       } else if (otherCount > 1) {
-        message.append("  and other " + otherCount + " dependency paths.\n");
+        message.append("  and " + otherCount + " other dependency paths.\n");
       }
     }
     return message.toString();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
@@ -16,6 +16,8 @@
 
 package com.google.cloud.tools.opensource.classpath;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 import com.google.cloud.tools.opensource.dependencies.DependencyPath;
 import com.google.cloud.tools.opensource.dependencies.UnresolvableArtifactProblem;
 import com.google.common.collect.ImmutableList;
@@ -68,5 +70,25 @@ public final class ClassPathResult {
   /** Returns problems encountered while constructing the dependency graph. */
   ImmutableList<UnresolvableArtifactProblem> getArtifactProblems() {
     return artifactProblems;
+  }
+
+  /** Returns text describing dependency paths to {@code jars} in the dependency tree. */
+  public String formatDependencyPaths(Iterable<Path> jars) {
+    StringBuilder message = new StringBuilder();
+    for (Path jar : jars) {
+      ImmutableList<DependencyPath> dependencyPaths = getDependencyPaths(jar);
+      checkArgument(dependencyPaths.size() >= 1, "%s is not in the class path", jar);
+
+      message.append(jar.getFileName() + " is at:\n");
+
+      int otherCount = dependencyPaths.size() - 1;
+      message.append("  " + dependencyPaths.get(0) + "\n");
+      if (otherCount == 1) {
+        message.append("  and 1 dependency path.\n");
+      } else if (otherCount > 1) {
+        message.append("  and other " + otherCount + " dependency paths.\n");
+      }
+    }
+    return message.toString();
   }
 }
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerMain.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerMain.java
@@ -61,8 +61,11 @@ class LinkageCheckerMain {
     List<ArtifactProblem> artifactProblems = new ArrayList<>();
     ClassPathResult classPathResult = null;
     if (artifacts.isEmpty()) {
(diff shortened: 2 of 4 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
```diff
@@ -69,4 +71,24 @@ public ClassPathResult(
...
+      checkArgument(dependencyPaths.size() >= 1, "%s is not in the class path", jar);
+
+      message.append(jar.getFileName() + " is at:\n");
+
+      int otherCount = dependencyPaths.size() - 1;
+      message.append("  " + dependencyPaths.get(0) + "\n");
+      if (otherCount == 1) {
+        message.append("  and 1 dependency path.\n");
+      } else if (otherCount > 1) {
+        message.append("  and other " + otherCount + " dependency paths.\n");
```
elharo: and 163 other dependency paths
instead
and other 163 dependency paths

suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
```diff
@@ -69,4 +71,24 @@ public ClassPathResult(
   ImmutableList<UnresolvableArtifactProblem> getArtifactProblems() {
     return artifactProblems;
   }
+
+  /** Returns text describing dependency paths to {@code jars} in the dependency tree. */
+  public String formatDependencyPaths(Iterable<Path> jars) {
+    StringBuilder message = new StringBuilder();
+    for (Path jar : jars) {
```
elharo: This could be simpler if this method only formatted a single jar. Then you'd only need one loop in LinkageChecker.main and you wouldn't have to build an intermediate list.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerMain.java
```diff
@@ -82,6 +86,20 @@ public static void main(String[] arguments)
 
     System.out.println(SymbolProblem.formatSymbolProblems(symbolProblems));
 
+    if (classPathResult != null && !symbolProblems.isEmpty()) {
+      ImmutableSet.Builder<Path> problematicJars = ImmutableSet.builder();
```
elharo: You don't need this set, just print each jar as you find it. That should also make the order more predictable.

suztomo: That would explode the number of lines in the output. We need this de-duplication logic. When there's 100 linkage errors on 100 classes from a JAR file, this should not print the dependency path to the that JAR file 100 times.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
suztomo: The intermediate ImmutableSet is for deduplication logic.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerMain.java
elharo: OK, that makes sense. Does the order of output matter though?

suztomo: No, the order does not matter. 

(Guava’s ImmutableSet guarantees a deterministic iteration order, which is not affected by JVM implementation.)

PR #1160, opened 2020-01-29
DependencyGraphBuilder.levelOrder optimization for Linkage Checker

### LINKED ISSUE #1145: Linkage Checker is slow when analyzing beam-sdks-java-extensions-sql-zetasql
I'm feeling it's doing something unexpected and unnecessary.

image

```
./gradlew -Ppublishing -PjavaLinkageArtifactIds=beam-sdks-java-extensions-sql-zetasql :checkJavaLinkage
```

suztomo:
Fixes #1145 . Still draft. A test case for symbol problem needs to be updated.

- If `DependencyGrpahBuilder.levelOrder` is called from Linkage Checker, then it does not need to populate the dependency tree with duplicate nodes, with regard to `groupId:artifactId`.
- The flag is `FULL_DEPENDENCY_WITH_PROVIDED` only when it's called from Linkage Checker. It's false if it's called from Dashboard's dependency convergence check.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
@@ -26,10 +26,12 @@ import com.google.common.collect.ImmutableMap;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Queue;
+import java.util.Set;
 import java.util.Stack;
 import org.eclipse.aether.RepositoryException;
 import org.eclipse.aether.RepositorySystem;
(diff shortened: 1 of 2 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -361,10 +364,21 @@ private DependencyGraphResult levelOrder(
           }
         }
       }
+
+      // If false, the resulting tree does not have duplicate artifacts with regard to
+      // groupId:artifactId. This behavior is an optimization for Linkage Checker.
+      boolean allowDuplicateNodeInTree =
+          graphTraversalOption != GraphTraversalOption.FULL_DEPENDENCY_WITH_PROVIDED;
```
elharo: This seems error prone. In particular this converts a multivalue to a boolean. But it could easily be the case that there are multiple values that allow duplicate nodes in the tree and multiple other values that don't, if not now then in the future. 

Perhaps this would be more obvious if the different traversal algorithms were different methods instead of being configured by an option.

suztomo:
Side effect found:

image

Closing this for now.

PR #1166, opened 2020-01-30
google-http-client-apache-v2

### LINKED ISSUE #973: Why/how is google-http-client-apache in the BOM?
should only be v2

suztomo:
Fixes #973 .


suztomo:
Updated the version to 4.0.0-SNAPSHOT

PR #1172, opened 2020-01-31
deps: update guava to 28.2-android and google-http-client to 1.34.1

chingor13:


elharo:
We already tried this. It's not going to work until protobuf is updated.

PR #1180, opened 2020-02-04
Fixed the relativize order

suztomo:
Fixing the previous commit on `relativize` and added tst.

My assumption on `MoreFiles.fileTraverser` always returning relative paths, was wrong. `MoreFiles.fileTraverser` returns absolute paths when it takes an absolute path. And my use of relativize was incorrect.

### DIFF
--- a/linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
+++ b/linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
@@ -120,12 +120,9 @@ public class LinkageMonitor {
         continue;
       }
 
-      if (path.isAbsolute()) {
-        // relative path from project directory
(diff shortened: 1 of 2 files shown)


linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
```diff
@@ -120,13 +120,9 @@ public static void main(String[] arguments)
         continue;
       }
 
-      if (path.isAbsolute()) {
-        // As of Guava 28, MoreFiles.fileTraverser returns relative paths. Just in case it changes the behavior,
-        // converting absolute paths to relative paths.
-        path = path.relativize(Paths.get(".").toAbsolutePath());
```
suztomo: The order was wrong. It must be `parentDirectory.relativize(child)`, where our parentDirectory is `projectDirectory`.

PR #1227, opened 2020-02-16
Gradle always has upper version alignment

### LINKED ISSUE #913: JLBP 16 should define upper version alignment directly
Not just as reference to glossary

elharo:
@garrettjonesgoogle fixes #913


docs/JLBP-16.md
```diff
@@ -1,27 +1,37 @@
...
 
 Upper version alignment increases the likelihood that consumers' build systems
 select the right versions of direct and transitive dependencies, reducing the
 number of conflicts.
 
-As noted in the definition, to fix a version misalignment caused by a shorter
-path to a version of a dependency that is not the upper bound in the
-dependency tree, a direct dependency needs to be added so that Maven consumers
-select the correct version.
+Gradle selects the highest version automatically. Maven, however, selects
```
garrettjonesgoogle: There is something not captured in the new version, which is that your own project might be using Gradle, but a Maven consumer of yours could be bit by a characteristic of your dependency tree. Thus, you need to fix it to benefit your Maven consumers.

docs/JLBP-16.md
```diff
@@ -1,27 +1,37 @@
...
+the version it encounters first when traversing the dependency tree,
```
suztomo: traversing in level order or breadth-first search order.

or 

the version that has a shorter path to the root node in the tree.

(The next sentence uses the word "shorter path")

docs/JLBP-16.md
```diff
@@ -1,27 +1,37 @@
...
+path to a lower version of a dependency, you can either:
 
-## Maven
+1. Update the dependency on the lower version to the upper bound.
+   This usually requires submitting a patch to another project 
+   and waiting for it to release. 
+   
+2. Add a direct dependency on the misaligned dependency to your 
+   own project, even though your project does not directly 
+   import any of that dependency's classes.
```
suztomo: `dependencyManagement` section capability is suitable for this purpose. I feel that's better than declaring an unused dependency.

docs/JLBP-16.md
suztomo:
```suggestion
Gradle selects the highest version by default. Maven, however, selects
```

docs/JLBP-16.md
```diff
@@ -1,27 +1,37 @@
...
 
 For multi-module projects, this best practice assumes you have already applied
 [JLBP-15](JLBP-15.md), so that your project has a BOM.
 
-Upper version alignment is defined in the [glossary](glossary.md).
+Upper version alignment means that when more than one version of a dependency
+is present in a module's dependency tree, the build system selects
+the highest version. Upper version alignment ensures that when packages are
+upgraded to higher versions which don’t make any breaking changes,
+there are no new linkage conflicts.
```
suztomo: In my view, upper version alignment does not ensure no new linkage conflicts. I would replace "Upper version alignment ensures ..." with the below:

> In general, selecting the highest version of a library is a good choice. The version has more security fixes and is more compatible with other libraries than the lower versions of it.

docs/JLBP-16.md
garrettjonesgoogle: Not explicitly, but it did implicitly, by recommending the course of action for everyone.

docs/JLBP-0016.md
```diff
@@ -8,25 +8,36 @@ permalink: /JLBP-16
...
 
 Upper version alignment increases the likelihood that consumers' build systems
 select the right versions of direct and transitive dependencies, reducing the
 number of conflicts.
 
-As noted in the definition, to fix a version misalignment caused by a shorter
-path to a version of a dependency that is not the upper bound in the
-dependency tree, a direct dependency needs to be added so that Maven consumers
-select the correct version.
+Gradle always chooses the highest version it finds in the dependency graph,
```
suztomo, line 23:
```suggestion
Gradle by default chooses the highest version it finds in the dependency graph,
```

Gradle can select old versions when it's configured to do so.

elharo: How would it be so configured?

suztomo: Adding constraints https://docs.gradle.org/current/userguide/java_platform_plugin.html#sec:java_platform_consumption or import BOMs with enforcedPlatform.

PR #1230, opened 2020-02-18
JLBP-6: Improving readability with positive wording

garrettjonesgoogle:
Also, adding descriptions to the cases at the end, otherwise it is
not clear what the cases mean.

### DIFF
--- a/docs/JLBP-6.md
+++ b/docs/JLBP-6.md
@@ -70,7 +70,7 @@ Consider the following renaming scenario:
 
 Given this scenario, here are the possible combinations of renamings:
 
-- **Keep Java package**:
+- **Keep Java package name**:
   - **Case 1: Keep Maven ID**. This approach can result in diamond
     dependency conflicts because different branches of a dependency tree can
     depend on different major versions, and the build system (Maven or Gradle)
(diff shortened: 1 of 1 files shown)


docs/JLBP-6.md
```diff
@@ -156,18 +156,18 @@ is to the decision.
 Examples in open source
 -----------------------
 
-**Case 1**
+**Case 1 (Keep same Java package and Maven ID)**
```
elharo: can probably diowithout parentheses here and below

docs/JLBP-6.md
```diff
@@ -70,8 +70,8 @@ Consider the following renaming scenario:
 
 Given this scenario, here are the possible combinations of renamings:
 
-- **Don't rename Java package**:
-  - **Case 1: Don't rename Maven ID**. This approach can result in diamond
+- **Keep Java package**:
```
elharo: Keep java package the same? "Keep" alone doesn't seem to convey the full meaning

docs/JLBP-6.md
garrettjonesgoogle: Removed the parentheses.

docs/JLBP-6.md
garrettjonesgoogle: I changed it to "Keep Java package name", which I hope clarifies it.

garrettjonesgoogle:
Addressed feedback, PTAL

docs/JLBP-6.md
```diff
@@ -176,5 +176,5 @@ Examples in open source
 - JDOM (org.jdom -> org.jdom2)
 - jdeferred (org.jdeferred -> org.jdeferred2)
 
-**Case 5 (Bundle old and new in the existing Maven ID)**
+**Case 5 - Bundle old and new in the existing Maven ID**
```
elharo: Can we add a noun here? e.g Publish old and new packages with the existing Maven ID

garrettjonesgoogle: Good idea, updated.

PR #1232, opened 2020-02-19
Guava version "-jre" and "-android" in libraries-bom documentation

suztomo:
@elharo I saw your earlier comment on your intention to rewrite/delete this document. This is the last item in my hand on updating docs/DECLARING_DEPENDENCIES.md

When I helped Sarah's dependency upgrade, I noticed that it's not obvious to resolve `NoSuchMethodError ImmutableList.toImmutableList()` which was caused by libraries-bom's Guava version.

The same note has been added to https://github.com/GoogleCloudPlatform/cloud-opensource-java/wiki/The-Google-Cloud-Platform-Libraries-BOM#guava-versions--jre-or--android.

### DIFF
--- a/docs/DECLARING_DEPENDENCIES.md
+++ b/docs/DECLARING_DEPENDENCIES.md
@@ -185,6 +185,55 @@ https://docs.gradle.org/current/userguide/platforms.html#sub:bom_import).
 For Gradle 4 example, refer to [Gradle 4.6 Release Notes: BOM import](
 https://docs.gradle.org/4.6/release-notes.html#bom-import).
 
+## Guava Versions "-jre" or "-android"
+
+Google's Guava release contains two flavors of artifacts:
+"-jre" and "-android" versions.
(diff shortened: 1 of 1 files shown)


docs/DECLARING_DEPENDENCIES.md
```diff
@@ -185,6 +185,55 @@ https://docs.gradle.org/current/userguide/platforms.html#sub:bom_import).
 For Gradle 4 example, refer to [Gradle 4.6 Release Notes: BOM import](
 https://docs.gradle.org/4.6/release-notes.html#bom-import).
 
+## Guava Versions "-jre" or "-android"
```
elharo: I need to think about this. I'm not sure this document is the right place for this.

suztomo:
Do we want to publish different BOMs for Java 7 and Java 8?

- Is this only guava difference? Objectify has similar situation. (not part of the BOM)
- gcsio (not part of the BOM) only supports Java 8.

suztomo:
I agree. The explanation is already there. Closing.

PR #1235, opened 2020-02-20
Adding links to introductory articles

garrettjonesgoogle:
Also moving the glossary link to the end because it is reference material.


docs/README.md
```diff
@@ -13,9 +13,10 @@ made. We have found that following these rules results in higher quality
 Java libraries with fewer dependency conflicts and other kinds of problems. The
 list is open-ended, so new ones may be added from time to time.
 
-- [Glossary](glossary.md): Terms used in the best practices and other places in
-  cloud-opensource-java.
+## Introductory material
+- [What is a diamond dependency conflict?](what-is-a-diamond-dependency-conflict.md)
```
elharo: This probably belongs with the glossary below.

garrettjonesgoogle: Why do you think the introductory material should be after the JLBPs?

elharo: The page is for the JLBPs. Diamond dependencies are not the only thing these documents are addressing.

garrettjonesgoogle: Sure, but I'm going to add more content than just this article. In my mind, I think we should put the section first that people are more likely to read first. So, are they more likely to read introductory articles first or JLBPs first, assuming they land at the front page? (I wouldn't count people who land directly on a JLBP for this question, because they went there for a reason, probably following a search result or external link)

elharo: The perhaps we should create a landing page just for that content. This page and this site is specifically for the JLBPs.

garrettjonesgoogle: I know we created the site for the JLBPs, but I think it is myopic to strictly limit the site to JLBPs. The content I would like to add helps people to understand the motivation for the JLBPs better and I believe it will drive more traffic to the site in a synergistic way. The site will be "heavier" and be more interlinked, meaning people will stay longer and search engines will rank it as more important. 

For the time being, I'm fine with putting the links to introductory content below the JLBPs in order to move forward. I don't want to isolate it into its own landing page yet.

garrettjonesgoogle:
I moved the link to the end and added a link to the 2nd article also, PTAL

garrettjonesgoogle:
@elharo the circle ci build failed for some random reason...

1. Could you restart the build or give me permissions to do so?
2. Could we skip the docs/ directory for running tests?

garrettjonesgoogle:
Ahh, that was it, I forgot to log in. Thanks!

PR #1229, opened 2020-02-18
DependencyPath and DependencyTreeFormatter to handle null optional flag.

suztomo:
- DependencyPath was not handling null optional flag.
- DependencyTreeFormatter was ignoring scopes and optional flags.

PR #1237, opened 2020-02-20
DependencyTreeFormatter to respect scopes and optional flags

suztomo:
Revise of https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1229 .

DependencyTreeFormatter was assuming every dependency has "compile" scope and optional false:

```
      parentArtifactPath.forEach(
          parentArtifactPathNode ->
              parentDependencyPath.add(new Dependency(parentArtifactPathNode, "compile", false)));
```

https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1237/files#diff-e825ccadbd98c809005839481296b814L84

This caused incorrect tree output when a dependency is not compile scope or optional=true.

### DIFF
--- a/dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatterTest.java
+++ b/dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatterTest.java
@@ -73,4 +73,24 @@ public class DependencyTreeFormatterTest {
         "The dependency should be output as tree with indentation",
         expectedTreeOutput, actualTreeOutput);
   }
+
+  @Test
+  public void testDependencyTree_scopeAndOptionalFlag() {
+    List<DependencyPath> dependencyPathList = new ArrayList<>();
+
+    DependencyPath path1 = new DependencyPath();
+    path1.add(new Dependency(new DefaultArtifact("io.grpc:grpc-auth:jar:1.15.0"), "compile", true));
+    dependencyPathList.add(path1);
+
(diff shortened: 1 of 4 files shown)


dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyTreeFormatterTest.java
```diff
@@ -73,4 +73,24 @@ public void testDependencyTree() {
...
+    DependencyPath path2 = new DependencyPath();
+    path2.add(new Dependency(new DefaultArtifact("io.grpc:grpc-auth:jar:1.15.0"), "compile", true));
+    path2.add(
+        new Dependency(new DefaultArtifact("io.grpc:grpc-core:jar:1.15.0"), "provided", false));
+    dependencyPathList.add(path2);
+
+    String actualTreeOutput = DependencyTreeFormatter.formatDependencyPaths(dependencyPathList);
+    String expectedTreeOutput =
+        "  io.grpc:grpc-auth:jar:1.15.0\n" + "    io.grpc:grpc-core:jar:1.15.0\n";
+    Assert.assertEquals(expectedTreeOutput, actualTreeOutput);
```
suztomo: This test was failing before this fix. DependencyTreeFormatter was ignoring scopes and optional flags (https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1237/files#diff-e825ccadbd98c809005839481296b814L84), while it was relying on equality of DependencyPath to find parent-child relationship from a Multimap.

PR #1239, opened 2020-02-20
LevelOrder not to call resolveCompileDependencies

### LINKED ISSUE #1188: Can we remove resolveCompileDependencies in DependencyGraphBuilder.levelOrder?
DependencyGraphBuilder.levelOrder calls `resolveCompileDependencies`. This makes things complex.

Can we remove this?

LevelOrder calls resolveCompileDependencies because we want to have duplicate nodes in resulting dependency graph. Normal Maven dependency graph does not allow duplicate on artifactId and groupId. If we can control Maven resolver to skip its dependency mediation, we don't need to call resolveCompileDependencies.

This will solve 1 million method calls in #1145 .

Trying with a new type of session that gives the full dependency graph.
```
  /**
   * Opens a new Maven repository session for full dependency graph resolution.
   *
   * <p>Compared with Maven's normal dependency graph, full dependency graph includes provided-
   * scope dependencies and optional dependencies even if they are transitive dependencies.
   */
  static RepositorySystemSession newSessionWithFullDependency(RepositorySystem system) {
    DefaultRepositorySystemSession session = createDefaultRepositorySystemSession(system);

    // This combination of DependencySelector comes from the default specified in
    // `MavenRepositorySystemUtils.newSession`.
    // LinkageChecker needs to include 'provided' scope and optional scope
    DependencySelector dependencySelector =
        new AndDependencySelector(
            new ScopeDependencySelector("test"),
            // OptionalDependencySelector is omitted
            new FilteringZipDependencySelector());
    session.setDependencySelector(dependencySelector);

    session.setDependencyGraphTransformer(
        new ChainedDependencyGraphTransformer(
            new CycleBreakerGraphTransformer(),
            new JavaDependencyContextRefiner()
        )
    );

    session.setReadOnly();
    return session;
  }
```

suztomo:
Fixes #1188 .

This fixes the bug of DependencyGraphBuilder applying exclusions elements only to its child's dependencies, rather than the entire subtree.

In go/jdd-tree-comparison, this bug is marked in the red rectable below

image


dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -332,10 +332,13 @@ public void testComponent_success() throws IOException, ParsingException {
   public void testComponent_failure() throws IOException, ParsingException {
     Document document = parseOutputFile(
         "com.google.api.grpc_grpc-google-common-protos_1.14.0.html");
+
+    // com.google.api.grpc:grpc-google-common-protos:1.14.0 selects the highest version of all dependencies.
     Nodes greens = document.query("//h3[@style='color: green']");
-    Assert.assertEquals(0, greens.size());
+    Assert.assertEquals(1, greens.size());
```
suztomo: Before this fix [the artifact report for grpc-google-common-protos](https://storage.googleapis.com/cloud-opensource-java-dashboard/com.google.cloud/libraries-bom/1.0.0/com.google.api.grpc_grpc-google-common-protos_1.14.0.html
) was showing Upper Bound Fixes for jsr305 version 3.0.1. But that version should not appear in the tree because grpc-core declares exclusions.

Tree: https://screenshot.googleplex.com/g8cQ3BXq36O.png

suztomo:
@elharo I'll explain tomorrow.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -316,15 +313,11 @@ public static UnresolvableArtifactProblem createUnresolvableArtifactProblem(
    * just follows the given dependency tree starting with firstNode.
    *
    * @param firstNode node to start traversal
-   * @param graphTraversalOption option to recursively resolve the dependency to build complete
-   *     dependency tree, with or without dependencies of provided scope
    */
-  private DependencyGraphResult levelOrder(
-      DependencyNode firstNode, GraphTraversalOption graphTraversalOption) {
+  private DependencyGraph levelOrder(DependencyNode firstNode) {
```
suztomo: No need to wrap DependencyGraph with DependencyGraphResult. Because this function does not encounter `UnresolvedArtifactProblem` any more.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/CycleBreakerGraphTransformerTest.java
```diff
@@ -0,0 +1,56 @@
...
+    DefaultRepositorySystemSession session =
+        RepositoryUtility.createDefaultRepositorySystemSession(system);
+
+    // This dependencySelector selects everything except test scope. This creates a dependency tree
+    // with a cycle of dom4j:dom4j:jar:1.6.1 (optional) and jaxen:jaxen:jar:1.1-beta-6 (optional).
+    session.setDependencySelector(new ScopeDependencySelector("test"));
+
+    session.setDependencyGraphTransformer(
+        new ChainedDependencyGraphTransformer(
+            new CycleBreakerGraphTransformer(), // This prevents StackOverflowError
```
suztomo: Without this class, this test would throw StackOverflowError.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/CycleBreakerGraphTransformer.java
```diff
@@ -0,0 +1,67 @@
...
+  @Override
+  public DependencyNode transformGraph(
+      DependencyNode dependencyNode, DependencyGraphTransformationContext context)
+      throws RepositoryException {
+
+    removeCycle(null, dependencyNode, new HashSet<>());
+    return dependencyNode;
+  }
+
+  private void removeCycle(DependencyNode parent, DependencyNode node, Set<Artifact> parents) {
```
elharo: There's only one parent. Perhaps this is ancestors? e.g. like in XPath

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/CycleBreakerGraphTransformer.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    parents.add(artifact);
+    for (DependencyNode child : node.getChildren()) {
+      removeCycle(node, child, parents);
+    }
+    parents.remove(artifact);
+  }
+
+  private void removeChildFromParent(DependencyNode parent, DependencyNode node) {
```
elharo: Is node the child to remove? If so, rename to `child`.
Also it could be a tad clearer if the argument order match the method name. That is,

`removeChildFromParent( DependencyNode child, DependencyNode parent)`

elharo: I think this method can be static

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/CycleBreakerGraphTransformer.java
```diff
@@ -0,0 +1,67 @@
...
+
+  private void removeCycle(DependencyNode parent, DependencyNode node, Set<Artifact> parents) {
+    Artifact artifact = node.getArtifact();
+    if (parents.contains(artifact)) {
+      // parent is not null when parents is not empty
+      removeChildFromParent(parent, node);
+      return;
+    }
+
+    parents.add(artifact);
```
elharo: Set feels wrong for parents, because these are ordered, even if we're not using the order here.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/CycleBreakerGraphTransformer.java
elharo: static

suztomo: Updated

suztomo: Nice. Updated to ancestor.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/CycleBreakerGraphTransformer.java
suztomo: Set is preferred here. Using a Set for ancestors gives O(1) lookup. At line 47:

```
if (ancestors.contains(artifact)) {
```

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/CycleBreakerGraphTransformer.java
suztomo: Updated.

suztomo: Updated to static.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/CycleBreakerGraphTransformer.java
```diff
@@ -0,0 +1,69 @@
...
+    removeCycle(null, dependencyNode, new HashSet<>());
+    return dependencyNode;
+  }
+
+  private static void removeCycle(
+      DependencyNode parent, DependencyNode node, Set<Artifact> ancestors) {
+    Artifact artifact = node.getArtifact();
+
+    if (ancestors.contains(artifact)) { // Set (rather than List) gives O(1) lookup here
+      // parent is not null when ancestors is not empty
```
elharo: How many elements are we talking about though? 10? Maybe 20 at the outside? Even if it were 1000 the difference is unlikely to be measurable.

suztomo: Sometimes around 1,000,000 nodes in a full dependency tree.
https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1145#issuecomment-578274062

suztomo: The tree height (`ancestors.size`) is not big, but this `ancestors.contains` is called that many times.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
suztomo: The comment above was correct but `grpc-google-common-protos:1.14.0`'s dependency tree still has old protobuf-java dependency.

PR #1243, opened 2020-02-21
Trying to fix link that should be converted to html

garrettjonesgoogle:
These links are not being converted to html links, so I'm removing the wrapping in an attempt to fix them.


suztomo:
Did you try running Jekyll in your local?

https://github.com/GoogleCloudPlatform/cloud-opensource-java/wiki/Tools#jekyll-for-github-pages

garrettjonesgoogle:
@suztomo I didn't try that before, thanks for the suggestion! It appears that undoing the wrapping does indeed fix the issue locally (I tried before & after to see the difference).

PR #1259, opened 2020-02-27
Adding article about version resolution

garrettjonesgoogle:


docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,108 @@
+# How does version resolution work in Maven and Gradle?
+
+The build system that you choose for building your Java code determines which
+version selection algorithm is used to choose the versions of your
```
elharo: usually called the dependency mediation algorithm, at least in Maven

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,108 @@
...
+dependencies. Unfortunately, the two most popular build systems in open source
+(Maven and Gradle) use different version resolution algorithms. No matter which
+of the two you choose, you are likely to have direct or indirect dependencies
```
elharo: delete "of the two"

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,108 @@
...
+that use the other one, so you need to understand how they both work when your
```
elharo: delete "they"

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,108 @@
...
+dependency graph experiences conflicts.
```
elharo: delete "when your dependency graph...conflicts"

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,108 @@
...
+
+Before we look at Maven and Gradle specifically, we need to understand what
+happens in the version resolution process. The input of this process is a
+dependency graph which reflects the dependencies exactly as specified by each
```
elharo: delete "exactly"

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,108 @@
...
+library in the graph, which we call an unresolved dependency graph. In such a
+graph, there can be multiple versions of each library. The version resolution
+process walks the unresolved dependency graph and produces a new graph where
```
elharo: I'm not sure it does, at least not Maven. It produces a linear classpath, not another graph.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: In Gradle it is called "version conflict resolution": https://docs.gradle.org/current/userguide/dependency_resolution.html

I will make a note of the two terms.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: will do.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: will do.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: will do.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: will do.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok, I will detach the actual output of Maven from the treatment of the output as a new graph.

garrettjonesgoogle:
Feedback addressed, PTAL

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,117 @@
+# How does version resolution work in Maven and Gradle?
+
+The build system that you choose for building your Java code determines which
```
elharo: Not sure what to do about this, but there are other build systems that do something completely different. E.g Eclipse and Ant don't necessarily use the Maven repository system at all. SBT and Ant+Ivy do. This article is really about "Version resolution in the Maven Respository System" or some such.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,117 @@
...
+version resolution algorithm is used to choose the versions of your
+dependencies. Unfortunately, the two most popular build systems in open source
+(Maven and Gradle) use different version resolution algorithms. They also use
+different names for the concept: Maven calls the process *dependency mediation*,
+while Gradle calls the process *version conflict resolution*. In this article,
+we will use the term *version resolution*. No matter which you choose, you are
```
elharo: Version conflict resolution is fine after this note. It's less jargony than "dependency mediation." but let's not introduce a thrid synonym.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,117 @@
...
+likely to have direct or indirect dependencies that use the other one, so you
+need to understand how both work.
+
+Before we look at Maven and Gradle specifically, we need to understand what
+happens in the version resolution process. The input of this process is a
+dependency graph which reflects the dependencies as specified by each library in
+the graph, which we call an *unresolved dependency graph*. In such a graph, there
+can be multiple versions of each library. The version resolution process walks
+the unresolved dependency graph and decides which version to use for every
+library encountered. Whether or not the build tool produces a graph as an output
```
elharo: "Whether or not the build tool produces a graph as an output
of this resolution, it is useful to think of this output in terms of a new
graph. We call the graph representation of this output a *resolved dependency
graph*."

I don't buy this. It's clearer to simply go straight to the linear classpath. The intermediate graph is an unnecessary step.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: There's a problem though - it overloads the "conflict" word. In our world (the JLBP project), we generally apply the word "conflict" to situations where something breaks, and we treat version resolution as a neutral process that doesn't have any "conflicts" if the result works. Using the word "conflict" for this part of the process in our documentation will muddy the waters. 

You could also view this differently: "version resolution" is just a simplified version of the term "version conflict resolution", and isn't really a distinct term. 

I will add some additional verbiage about omitting the word "conflict".

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: How do you describe the output of `mvn dependency:tree`?

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Our entire set of best practices ignores the Eclipse/Ant world, right? The practices are all in terms of the Maven repository system. Anyway, I will add a note that Eclipse/Ant are out of scope.

elharo: no, not completely. JLBP-17, 18, 19, and 20 for instance apply equally well to ant. So does JLBP-1 and probably others.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
elharo: debugging text printed on the console

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: They *can* apply incidentally, but our focus is on build systems using the Maven repository system. 

If you want to increase our scope to Eclipse/Ant, we'd probably want to clarify throughout our best practices what the scope of each one is. For example, I have no idea if BOMs are useful (or possible) for Eclipse/ant.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: The text structure is in a tree format (or as best a tree can be represented in text).

When is it used? When debugging dependency conflicts.

Thus, I would argue that thinking of the result of dependency resolution as a tree (or more correctly, a graph) already has precedent, and is the best format. A linear format is harder to take action on in order to fix dependency problems since the input is a tree, and the input is where problems need to be fixed. 

@suztomo do you have thoughts on this?

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,121 @@
...
+
+Next let's consider Maven. Maven chooses the version that is closest to the root
+of the dependency graph, and if there is a tie, then it chooses the version it
+encounters first. Another way to describe the algorithm is that Maven chooses
+the first version encountered when traversing the dependency graph following a
+breadth-first traversal. Looking at the same example we used first for Gradle:
+
+<img src="assets/images/ddc-resolution-06.png">
+
+As you can see if you compare the resolved graphs, Maven makes the opposite
```
elharo: delete " of Gradle in this circumstance"

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,121 @@
...
+choice of Gradle in this circumstance. Let's look at the second example, where
+the root has both a direct and indirect dependency on the same library:
+
+<img src="assets/images/ddc-resolution-07.png">
+
+In this case, Maven chooses A:10.0 because it is only 1 hop away from the root
```
elharo: delete "In this case, Maven also makes the opposite choice of Gradle."

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,121 @@
...
+instead of 2 hops (A:10.1). In this case, Maven also makes the opposite choice
+of Gradle.
+
+So where can the mixture of these two build systems cause a problem? The
```
elharo: delete "So where can the mixture of these two build systems cause a problem?"

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,121 @@
...
+resolved dependency graph can experience conflicts when the two build systems
```
elharo: That's not quite right. There are two different resolved dependency graphs (or classpaths) here, not one. Perhaps rephrase around program behavior. E.g "When the two build systems select different versions of a transitive dependency, a library that works with one might not work with the other."

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,121 @@
...
+come to different decisions for a particular subgraph of a dependency graph.
+
+Let's look at an example, starting with Gradle example 2, where library D
```
elharo: where's "Gradle Example 2"? Generally should the figures have captions?

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,121 @@
...
+depends on A:10.0 and C:30.0 (and thus indirectly A:10.1). Let's also say that C
+depends on a feature added into 10.1. This means that if 10.0 is selected as the
+resolved version, then C will fail at runtime. From library D's perspective,
+this is fine, since Gradle chooses version 10.1.
+
+Let's add a new library (let's call it library E) which uses Maven as its build
```
elharo: "Let's add a new library (let's call it library E)" --> Suppose we add a new library E

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,121 @@
...
+system, and which adds D as a dependency. When E is built, Maven will resolve
```
elharo: will resolve --> resolves for consistent present tense

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,121 @@
...
+the whole dependency graph, including D's subgraph, even though Gradle resolved
+the subgraph of D for itself when D was originally built. When Maven performs
+its version resolution, it chooses a different version of A (10.0) than Gradle
+did (10.1), which breaks C. This happens even though D works perfectly fine
+internally.
+
+<img src="assets/images/ddc-resolution-08.png">
+
+As a consequence, the author of library E might file a bug against library D
+because it seems to have a bug. The author of library D would argue that there
```
elharo: delete "because it seems to have a bug"

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,121 @@
...
+is no bug from their perspective. From their narrow perspectives they are both
```
elharo: delete "from their perspective"

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,121 @@
...
+right. From the ecosystem perspective, the author of library D unfortunately
+needs to adapt their dependencies so that they don't cause problems for Maven
```
elharo: so that --> so

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,121 @@
...
+consumers (even though they may have sworn off Maven and use Gradle
+exclusively). In this case, they have an easy fix - they can upgrade the direct
+dependency of D on A:10.0 to A:10.1, so that both build systems make the same
```
elharo: "ake the same
version resolution decisions" --> select the same versions

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: I have added a note about this.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Why? I want to call it out explicitly - it's not otherwise stated.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: That leaves a jarring shift in topic. I'll make a header instead I suppose.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: It's above - ddc-resolution-05.png. Should the caption / image title be displayed differently?

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: oops I've been trying to avoid this, good catch

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: I decided to completely rewrite the paragraph - I hope the rewritten version conveys the idea better.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
+# How does version resolution work in Maven and Gradle?
+
+The build system that you choose for building your Java code determines which
+version resolution algorithm is used to choose the versions of your
+dependencies. Unfortunately, the two most popular build systems in open source
```
elharo: delete "in open source" since these are used for closed source projects too

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+(Maven and Gradle) use different version resolution algorithms. They also use
+different names for the concept: Maven calls the process *dependency mediation*,
+while Gradle calls the process *version conflict resolution*. In this article,
+we will use the term *version resolution*, omitting the word "conflict" so that
+the process it not confused with program-breaking conflicts that we discuss
```
elharo: it --> is.
and don't we call those "dependency conflicts" elsewhere? If so I'm not sure we need to change it here.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+elsewhere. No matter which build system you choose, you are likely to have
+direct or indirect dependencies that use the other one, so you need to
+understand how both work. Note: Our focus is on build systems based on the Maven
```
elharo: Delete "note:"
This still feels to me like it belongs in the first paragraph rather than a note at the bottom. This is fundamental to what the article discusses.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+repository system, so we ignore other build environments like Eclipse and Ant
+that work differently.
+
+Before we look at Maven and Gradle specifically, we need to understand what
+happens in the version resolution process. The input of this process is a
```
elharo: But there is no version resolution process without Maven or Gradle. You can describe what exists in the repository, and you can describe what Maven and Gradle do with the information they find there. Nothing else exists.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+looks like the following. In this graph, C's dependency on A:10.1 is overridden.
+
+<img src="assets/images/ddc-resolution-02.png">
+
+If A:10.1 is picked, the resolved dependency graph looks like the following. In
+this graph, B's dependency on A:10.0 is overridden.
+
+<img src="assets/images/ddc-resolution-03.png">
+
+So why do we need to pick a single version? The reason is that Java's default
```
elharo: Java's default class loader --> each class loader

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+class loader only loads one version of each fully-qualified class name at
+runtime. So, if you try to put multiple versions on the classpath, you would not
```
elharo: So, if you try to put --> If you put
would not --> do not

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+get both versions at runtime. Java build systems handle this issue upfront by
+picking a single version for each library before the classpath is even
+constructed.
+
+Now let's look at the version resolution algorithms used by Gradle and
```
elharo: Google tech writer style guide discourages "Let's'" and recommends second and third person as appropriate: 

https://developers.google.com/style/person

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+Maven. Let's consider Gradle first because it is the simpler one to
+describe. Simply put, Gradle chooses the highest version encountered in your
+dependency graph. For the example above, Gradle would pick A:10.1.
```
elharo: would pick -->picks
Per style guide, avoid "would"
https://developers.google.com/style/tense

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+
+<img src="assets/images/ddc-resolution-04.png">
+
+Gradle applies the higher-version rule even if it overrides one of your own
+direct dependency declarations:
+
+<img src="assets/images/ddc-resolution-05.png">
+
+Next let's consider Maven. Maven chooses the version that is closest to the root
```
elharo: delete "let's"

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+of the dependency graph, and if there is a tie, then it chooses the version it
```
elharo: ,and,

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+encounters first. Another way to describe the algorithm is that Maven chooses
+the first version encountered when traversing the dependency graph following a
```
elharo: "following a
breadth-first traversal" --> in breadth first order.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+breadth-first traversal. Looking at the same example we used first for Gradle:
```
elharo: delete "first"

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+
+<img src="assets/images/ddc-resolution-06.png">
+
+As you can see if you compare the resolved graphs, Maven makes the opposite
```
elharo: "if you compare" --> by comparing

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+choice. Let's look at the second example, where the root has both a direct and
```
elharo: let's look at --> Now consider

docs/how-does-version-resolution-work-in-maven-and-gradle.md
elharo: Maven won't always make the opposite choice of Gradle. The sentence would work if Gradle chose the furthest an Gradle chose the nearest, or some such weird thing.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+
+<img src="assets/images/ddc-resolution-07.png">
+
+In this case, Maven chooses A:10.0 because it is only 1 hop away from the root
+instead of 2 hops (A:10.1). In this case, Maven also makes the opposite choice
+of Gradle.
+
+## Interaction between Maven and Gradle
+
+Every time a library is built, its build system performs version resolution for
```
elharo: Every time --> When
its build system --> the build system
(some libraries are built with different systems; e.g. blaze)

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+its entire dependency graph. Since the libraries in that dependency graph also
+perform their own version resolution, the versions selected can be different,
```
elharo: libraries don't perform version resolution. Only build systems do.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+especially when a consumer uses a different build system than its dependencies.
+Sometimes, the version selected by the consumer can be incompatible with the
+version selected by the dependency when the dependency is built by itself.
```
elharo: "selected by the dependency when the dependency is built by itself" --> "expected by the dependency"

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+
+Let's look at an example, starting with Gradle example 2, where library D
```
elharo: Let's

docs/how-does-version-resolution-work-in-maven-and-gradle.md
elharo: It should probably be text in the document somehow as that's much better for accessibility. This might require a little embedded HTML. https://stackoverflow.com/questions/19331362/using-an-image-caption-in-markdown-jekyll has some suggestions

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+depends on A:10.0 and C:30.0 (and thus indirectly A:10.1). Let's also say that C
+depends on a feature added into 10.1. This means that if 10.0 is selected as the
```
elharo: " as the
resolved version" is redundant with selected and can be deleted

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+resolved version, then C will fail at runtime. From library D's perspective,
```
elharo: will fail --> fails
https://developers.google.com/style/tense

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+this is fine, since Gradle chooses version 10.1.
+
+Suppose we add a new library E which uses Maven as its build system, and which
```
elharo: no first person

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+adds D as a dependency. When E is built, Maven resolves the whole dependency
+graph, including D's subgraph, even though Gradle resolved the subgraph of D for
+itself when D was originally built. When Maven performs its version resolution,
```
elharo: "itself" is unclear here. Does this refer to Gradle or D?

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+it chooses a different version of A (10.0) than Gradle did (10.1), which breaks
+C. This happens even though D works perfectly fine internally.
+
+<img src="assets/images/ddc-resolution-08.png">
```
elharo: Images need alt text for accessibility

https://developers.google.com/style/images#alt-text

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+
+As a consequence, the author of library E might file a bug against library
+D. The author of library D would argue that there is no bug. From their narrow
```
elharo: narrow --> individual
comma after perspectives

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+perspectives they are both right. From the ecosystem perspective, the author of
```
elharo: they are both right --> both are right

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+library D unfortunately needs to adapt their dependencies so they don't cause
```
elharo: their --> its

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,124 @@
...
+problems for Maven consumers (even though they may have sworn off Maven and use
+Gradle exclusively). In this case, they have an easy fix - they can upgrade the
+direct dependency of D on A:10.0 to A:10.1, so that both build systems select
+the same versions, and everyone can be happy.
```
elharo: versions --> version
can be --> is

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: I will rework this to refer to "the concepts of version resolution in general" - there are logical steps that any build system needs to perform. Denying generalities is like denying that we can talk about circles since no object is a perfect circle.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: ping @suztomo

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Removing "then"... not sure exactly what you were looking for though.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Gradle examples 1 and 2 are different - I wanted to refer to 1. Rewording to try to make that clear.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: I don't get your argument...

I will add one more diagram to show Maven & Gradle doing the same thing. I hope that makes this sentence work better.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: "the build system" also implies there is only one... I don't feel like it clarifies anything. Going into the nuances of multiple build systems per library seems like it would muddy the picture too much and make the article less useful. In the end, only 1 build system is used to build the artifacts that are pushed to Maven, even if a second build system is available to use.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Adding "the build systems of ..."

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: I can understand the motivation to simplify this, but I feel that there is value in being explicit about the separate build processes in the dependency vs the consumer. Just saying "expected by" is somewhat inert.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: That makes sense... could I defer this until after the navigation PR is submitted? (We don't have an _includes/ directory yet, as suggested by one of the recommendations)

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Taken care of

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Done

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: yep yep, deeply ingrained habits die hard...

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ugh, all of my articles use this extensively. Do you really want to force a full rewrite before this article is submitted?

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Removing "itself", it's not necessary

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok, adding alt text

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok

garrettjonesgoogle:
PTAL

docs/how-does-version-resolution-work-in-maven-and-gradle.md
elharo: It's not used that much, and the change is pretty straightforward. E.g. here "Suppose a new library E uses Maven as its build system and has a dependency on D."

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,132 @@
...
+library D's perspective, this is fine, since Gradle chooses version 10.1.
+
+Suppose we add a new library E which uses Maven as its build system, and which
+adds D as a dependency. When E is built, Maven resolves the whole dependency
+graph, including D's subgraph, even though Gradle resolved the subgraph of D for
+when D was originally built. When Maven performs its version resolution,
+it chooses a different version of A (10.0) than Gradle did (10.1), which breaks
+C. This happens even though D works perfectly fine internally.
+
+<img src="assets/images/ddc-resolution-08.png" alt="Gradle and maven disagreement">
```
elharo: Maven

docs/how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,132 @@
...
+
+As a consequence, the author of library E might file a bug against library
+D. The author of library D would argue that there is no bug. From their individual
+perspectives, are both right. From the ecosystem perspective, the author of
+library D unfortunately needs to adapt its dependencies so they don't cause
+problems for Maven consumers (even though they may have sworn off Maven and use
+Gradle exclusively). In this case, they have an easy fix - they can upgrade the
```
elharo: nit: hyphen to em dash, or make this two sentences.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Ok I rewrote this sentence as suggested.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
garrettjonesgoogle: Done.

garrettjonesgoogle:
Merged in the page metadata PR. PTAL.

docs/0003-how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,136 @@
...
+put multiple versions on the classpath, you do not get both versions at
+runtime. Java build systems handle this issue upfront by picking a single
+version for each library before the classpath is even constructed.
+
+Now let's look at the version resolution algorithms used by Gradle and
+Maven. Let's consider Gradle first because it is the simpler one to
+describe. Simply put, Gradle chooses the highest version encountered in your
+dependency graph. For the example above, Gradle picks A:10.1.
+
+<img src="assets/images/ddc-resolution-04.png" alt="Gradle example 1">
```
elharo: alt text should be a description of the content of the image in words sufficient for a non-sighted user. These images are core parts of the article, not simple decorations, so it's worth writing this out in some detail. E.g.

D: 40.0 is the root.
It depends on B:20.0 and C:30.0.
B:20.0 and C:30.0  both depend on A:10.1.
Nothing depends on A:10.0.

docs/how-does-version-resolution-work-in-maven-and-gradle.md
suztomo: I need this paragraph. This input tree and output tree are the fundamental part of comparing Gradle and Maven's behavior.

docs/0003-how-does-version-resolution-work-in-maven-and-gradle.md
```diff
@@ -0,0 +1,136 @@
...
+system than its dependencies. Sometimes, the version selected by the consumer
+can be incompatible with the version selected by the dependency when the
+dependency is built by itself.
+
+Consider an example using Gradle example 2 as its basis, where library D depends
+on A:10.0 and C:30.0 (and thus indirectly A:10.1). Assume C depends on a feature
+added into 10.1. This means that if 10.0 is selected, C fails at runtime. From
+library D's perspective, this is fine, since Gradle chooses version 10.1.
+
+Suppose a new library E uses Maven as its build system and has a dependency on
```
suztomo: I would add the assumption that library D is now published to a Maven repository (potentially local) through Gradle's maven-publish plugin. That makes the two worlds work together.

elharo:
closing to reopen and rerun

PR #1266, opened 2020-03-03
remove deprecated class

elharo:
@suztomo " This interface does not support loading of parent POM(s) from the same backing store, integrators are strongly encouraged to implement {@link ModelSource2} instead of implementing this interface directly." and in fact the superclass here does that.

### DIFF
--- a/linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/VersionSubstitutingModelResolver.java
+++ b/linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/VersionSubstitutingModelResolver.java
@@ -20,7 +20,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import java.util.List;
 import java.util.Map;
-import org.apache.maven.model.building.ModelSource;
+import org.apache.maven.model.building.ModelSource2;
 import org.apache.maven.model.resolution.ModelResolver;
 import org.apache.maven.model.resolution.UnresolvableModelException;
 import org.apache.maven.project.ProjectModelResolver;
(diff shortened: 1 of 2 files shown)


linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/VersionSubstitutingModelResolver.java
```diff
@@ -68,16 +68,16 @@
...
-   * Returns model source for {@code groupId:artifactId:version}. The version is substituted if
-   * {@link #versionSubstitution} contains the versionless coordinates ({@code groupId:artifactId})
-   * in its keys.
+   * Returns a file model source for {@code groupId:artifactId:version}.
+   * The version is substituted if {@link #versionSubstitution} contains the
+   * versionless coordinates ({@code groupId:artifactId}) in its keys.
    */
   @Override
-  public ModelSource resolveModel(String groupId, String artifactId, String version)
+  public ModelSource2 resolveModel(String groupId, String artifactId, String version)
```
suztomo: I feel weird that the function return type is different from the super class.  http://maven.apache.org/ref/3.6.3/maven-model-builder/apidocs/org/apache/maven/model/resolution/ModelResolver.html#resolveModel-java.lang.String-java.lang.String-java.lang.String-

```
ModelSource resolveModel(String groupId,
                         String artifactId,
                         String version)
                  throws UnresolvableModelException
```

Shouldn't this be fixed at `org.apache.maven.model.resolution.ModelResolver`?

elharo: Possibly, though it might need to be that way for backwards compatibility.

This is a covariant return type introduced in Java 5. It still fulfills the contract of the superclass/superinterface .

suztomo: This case was return value type, which is not used in method lookup. But what if there’s an argument of ModelSource.

suztomo: > covariant

Right. That’s the difference from the case of the arguments. Thank you.

PR #1269, opened 2020-03-03
Parser for Linkage Checker exclusion file

suztomo:
Fixes https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1265


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileParser.java
```diff
@@ -0,0 +1,150 @@
...
+  static ImmutableList<LinkageErrorMatcher> parse(Path exclusionFile)
+      throws ParserConfigurationException, SAXException, IOException {
+    SAXParserFactory spf = SAXParserFactory.newInstance();
+    spf.setNamespaceAware(true);
+    SAXParser saxParser = spf.newSAXParser();
+
+    XMLReader xmlReader = saxParser.getXMLReader();
+    LinkageErrorMatcherHandler handler = new LinkageErrorMatcherHandler();
+    xmlReader.setContentHandler(handler);
+    xmlReader.parse("file://" + exclusionFile);
```
suztomo: This part does not work with Windows.

suztomo:
Memo: prefer org.xml.sax to javax.xml

suztomo:
No usage of javax.xml. The test passes in Windows too.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileParser.java
```diff
@@ -0,0 +1,177 @@
...
+import org.xml.sax.helpers.DefaultHandler;
+import org.xml.sax.helpers.XMLReaderFactory;
+
+/**
+ * Parser for Linkage Checker exclusion files.
+ *
+ * <p>The exclusion file for Linkage Checker is an XML file. Its top-level element is
+ * LinkageCheckerFilter. The XML file contains the following structure:
+ *
+ * <ul>
```
elharo: Optional, and maybe not in this PR: writing an XSD, RELAX NG schema, or DTD can be an effective way to document this and si sometimes helpful in runtime validation.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileParser.java
```diff
@@ -0,0 +1,177 @@
...
+    XMLReader xmlReader = XMLReaderFactory.createXMLReader();
+    LinkageErrorMatcherHandler handler = new LinkageErrorMatcherHandler();
+    xmlReader.setContentHandler(handler);
+
+    InputSource inputSource = new InputSource(new FileInputStream(exclusionFile.toFile()));
+    xmlReader.parse(inputSource);
+    return handler.getMatchers();
+  }
+
+  private static class LinkageErrorMatcherHandler extends DefaultHandler {
```
elharo: Outer classes are easier to follow and reason about.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileParser.java
```diff
@@ -0,0 +1,177 @@
...
+        throw new SAXException(
+            "Unexpected parent-child relationship. Parent:" + parent + ", child:" + child);
+      }
+    }
+
+    @Override
+    public void startElement(
+        String namespaceURI, String localName, String qName, Attributes attributes)
+        throws SAXException {
+      switch (localName) {
```
elharo: you can check that the namespace is empty before the switch statement and error or ignore if it isn't

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileParser.java
```diff
@@ -0,0 +1,177 @@
...
+          break;
+        default:
+          throw new SAXException("Unknown tag " + localName);
+      }
+    }
+
+    @Override
+    public void endElement(String uri, String localName, String qName) throws SAXException {
+      SymbolProblemMatcher poppedMatcher;
+      switch (localName) {
```
elharo: watch for namespaces

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/PackageMatcher.java
```diff
@@ -0,0 +1,44 @@
...
+
+  private final String packageName;
+
+  PackageMatcher(String packageName) {
+    this.packageName = checkNotNull(packageName);
+  }
+
+  @Override
+  public boolean match(Symbol problem) {
+    return problem.getClassBinaryName().startsWith(packageName);
```
elharo: Does startsWith work here or do we need a period or forward slash at the end to avoid accidental matches?

dependencies/src/test/resources/exclusion-sample-rules/linkage-checker-sample-exclusion-rule.xml
```diff
@@ -0,0 +1,25 @@
...
+      <Package name="jdk.vm.ci" />
+    </Target>
+    <Source>
+      <Package name="com.oracle.svm" />
+      <Package name="com.oracle.graal" />
+      <Package name="org.graalvm" />
+    </Source>
+  </LinkageError>
+  <LinkageError>
+    <Target>
```
elharo: can target come after source too?

dependencies/src/test/resources/exclusion-sample-rules/target-one-match.xml
```diff
@@ -0,0 +1,12 @@
+<LinkageCheckerFilter>
+  <LinkageError>
+    <Target>
+      <Method className="com.google.Foo" name="methodA" />
```
elharo: There's a combinatorial explosion here: every source class against every target method/field. Perhaps each target should have at most one source and at most one target and each target can have exactly one child.

dependencies/src/test/resources/exclusion-sample-rules/target-package.xml
```diff
@@ -0,0 +1,7 @@
+<LinkageCheckerFilter>
+  <LinkageError>
+    <Target>
+      <Package name="com.google" />
```
elharo: does this include subclasses?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileParser.java
suztomo: Nice idea. Created https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1277

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileParser.java
suztomo: Moved this to an independent class.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileParser.java
suztomo: Added that check to throw an error if namespaceURI is not empty.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileParser.java
suztomo: Added that check.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/PackageMatcher.java
suztomo: Updated the logic. "com.ab" should not match "com.abcd" here. Added test `testParse_targetPackage_unmatch`.

dependencies/src/test/resources/exclusion-sample-rules/linkage-checker-sample-exclusion-rule.xml
suztomo: Yes, the order does not matter.

dependencies/src/test/resources/exclusion-sample-rules/target-one-match.xml
suztomo: Created a followup on the restriction to avoid combinational explosion.

https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1278

dependencies/src/test/resources/exclusion-sample-rules/target-package.xml
suztomo: Yes, it includes subpackages, as we do the special cases in [Linkage Checker](https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java#L208)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
```diff
@@ -0,0 +1,129 @@
...
+
+import com.google.common.collect.ImmutableList;
+import java.util.ArrayDeque;
+import org.xml.sax.Attributes;
+import org.xml.sax.SAXException;
+import org.xml.sax.helpers.DefaultHandler;
+
+class ExclusionFileHandler extends DefaultHandler {
+
+  private ArrayDeque<SymbolProblemMatcher> stack = new ArrayDeque<>();
```
elharo: It is unclear what this stack does or how it works. Some comments would be helpful. 

It might be that stack is not the best data structure here. A custom class might work better.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
```diff
@@ -0,0 +1,129 @@
...
+  private ImmutableList.Builder<LinkageErrorMatcher> matchers;
+
+  ImmutableList<LinkageErrorMatcher> getMatchers() {
+    return matchers.build();
+  }
+
+  @Override
+  public void startDocument() {
+    matchers = ImmutableList.builder();
```
elharo: It looks like the fields are half reusable and half not. Do you intend a single handler to be used to parse more than one document or not? Either way and make sure to document this and either:

Throw an IllegalStateException the second time startDocument is called

or 

reset all fields in startDocument

dependencies/src/test/resources/exclusion-sample-rules/target-one-match.xml
elharo: This is not really followup eligible. We need to get this right in the design now, not fix it later.

suztomo:
@elharo I will revise the target/source combination explosion https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1278 . Closing this for now.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: Added IllegalStateException and Javadoc.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: Added javadoc to the stack. (Let me know if you think a custom class can express this better than stack.)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  @Override
+  public void startElement(
+      String namespaceURI, String localName, String qName, Attributes attributes)
+      throws SAXException {
+    if (!namespaceURI.isEmpty()) {
+      throw new SAXException("The exclusion rule does not support XML namespace");
+    }
+    switch (localName) {
```
suztomo: More validation in XML structure is to be done at https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1277

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * Stack to hold the ancestors of the XML element in {@link #startElement(String, String, String,
+   * Attributes)}.
+   */
+  private ArrayDeque<SymbolProblemMatcher> stack = new ArrayDeque<>();
+
+  private ImmutableList.Builder<LinkageErrorMatcher> matchers;
+
+  ImmutableList<LinkageErrorMatcher> getMatchers() {
```
elharo: Is the order of the matchers significant?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
```diff
@@ -0,0 +1,142 @@
...
+
+  @Override
+  public void startDocument() {
+    if (matchers != null) {
+      throw new IllegalStateException("This handler started reading document already");
+    }
+    matchers = ImmutableList.builder();
+  }
+
+  private void addMatcherToTop(Object child) throws SAXException {
```
elharo: Why is the argument type Object? It seems this could be more specific.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
```diff
@@ -0,0 +1,142 @@
...
+          "Unexpected parent-child relationship. Parent:" + parent + ", child:" + child);
+    }
+  }
+
+  @Override
+  public void startElement(
+      String namespaceURI, String localName, String qName, Attributes attributes)
+      throws SAXException {
+    if (!namespaceURI.isEmpty()) {
+      throw new SAXException("The exclusion rule does not support XML namespace");
```
elharo: Suggest changing message to "unrecognized element: " + qName + " in namespace " + namespaceURI

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
```diff
@@ -0,0 +1,142 @@
...
+      ((TargetMatcher) parent).setMatcher((SymbolProblemTargetMatcher) child);
+    } else {
+      throw new SAXException(
+          "Unexpected parent-child relationship. Parent:" + parent + ", child:" + child);
+    }
+  }
+
+  @Override
+  public void startElement(
+      String namespaceURI, String localName, String qName, Attributes attributes)
```
elharo: namespaceUri
qualifiedName

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
elharo: I'd suggest that as documentation, not as part of the processing chain.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
```diff
@@ -0,0 +1,142 @@
...
+            new MethodMatcher(classNameOnMethod, attributes.getValue("name"));
+        addMatcherToTop(methodMatcher);
+        break;
+      case "Field":
+        String classNameOnField = attributes.getValue("className");
+        FieldMatcher fieldMatcher = new FieldMatcher(classNameOnField, attributes.getValue("name"));
+        addMatcherToTop(fieldMatcher);
+        break;
+      default:
+        throw new SAXException("Unknown tag " + localName);
```
elharo: It's OK and arguably better to simply ignore unrecongized elements.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodMatcher.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package com.google.cloud.tools.opensource.classpath;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/** Matcher on method symbols. */
+class MethodMatcher implements SymbolProblemTargetMatcher {
+
+  private String className;
```
elharo: final

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: The order matters to detect unused exclusion rules.
https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1275

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: Good catch. It's SymbolProblemTargetMatcher interface. Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: Fixed.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: Nice. Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: Started writing https://github.com/GoogleCloudPlatform/cloud-opensource-java/wiki/Linkage-Checker-Exclusion-File

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: I prefer to show errors for unrecognized elements. The errors help to detect typos.

More validation will be done in https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1277

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodMatcher.java
suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
```diff
@@ -0,0 +1,146 @@
...
+    if (matchers != null) {
+      throw new IllegalStateException("This handler started reading document already");
+    }
+    matchers = ImmutableList.builder();
+  }
+
+  /** Adds {@code child} to the parent matcher that is at the top of the {@link #stack}. */
+  private void addMatcherToParent(SymbolProblemTargetMatcher child) throws SAXException {
+    SymbolProblemMatcher parent = stack.peek();
+    if (parent instanceof SourceMatcher && child instanceof SymbolProblemSourceMatcher) {
```
elharo: This is an awful lot of casting and instanceof, and it makes me wonder if we're not taking advantage of polymorphism where we could, or perhaps we don't have quite the right class hierarchy.

Is there perhaps some way SymbolProblemMatcher.setMatcher could simply work when it's passed any instance of SymbolProblemTargetMatcher?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
```diff
@@ -0,0 +1,146 @@
...
+      default:
+        throw new SAXException("Unknown tag " + localName);
+    }
+  }
+
+  @Override
+  public void endElement(String namespaceUri, String localName, String qualifiedName)
+      throws SAXException {
+    SymbolProblemMatcher poppedMatcher;
+    if (!namespaceUri.isEmpty()) {
```
elharo: I don't think this case can happen, given that this case would have been detected already unless the document is malformed, in which case you'll also have errored out before this is called.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
```diff
@@ -0,0 +1,146 @@
...
+      throw new SAXException(
+          "unrecognized element: " + qualifiedName + " in namespace " + namespaceUri);
+    }
+    switch (localName) {
+      case "Source":
+      case "Target":
+        poppedMatcher = stack.pop();
+        if (!(poppedMatcher instanceof TargetMatcher)
```
elharo: I'm not sure this case is reachable either

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
```diff
@@ -0,0 +1,146 @@
...
+        if (!(poppedMatcher instanceof LinkageErrorMatcher)) {
+          throw new SAXException("Unexpected stack status after reading LinkageError element");
+        }
+        matchers.add((LinkageErrorMatcher) poppedMatcher);
+    }
+  }
+
+  @Override
+  public void endDocument() throws SAXException {
+    if (stack.size() != 0) {
```
elharo: `!stack.isEmpty()`

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: > SymbolProblemMatcher.setMatcher could simply work when it's passed any instance of SymbolProblemTargetMatcher?

No, SourceMatcher cannot take FieldMatcher. I replaced the stack and this function with the two variables linkageErrorMatcher and symbolProblemMatcher. Now the logic is much simpler. Does this look good now?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: Removed this check in endElement.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: Removed endElement altogether.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: Good catch. But endDocument is no longer needed.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileHandler.java
suztomo: SymbolProblemMatcher's addChild now may be the one you mentioned above:

```
  default void addChild(SymbolProblemTargetMatcher child) {
    throw new IllegalStateException("This element is not supposed to take child element");
  }
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorMatcher.java
```diff
@@ -0,0 +1,50 @@
...
+  void setSourceMatcher(SourceMatcher sourceMatcher) {
+    this.sourceMatcher = sourceMatcher;
+  }
+
+  void setTargetMatcher(TargetMatcher targetMatcher) {
+    this.targetMatcher = targetMatcher;
+  }
+
+  /**
+   * Returns true if all of non-null matchers of {@link #sourceMatcher} and {@link #targetMatcher}
```
elharo: all of --> all
Can there be null matchers?

suztomo: No `NullMatcher` but `sourceMatcher` or `targetMatcher` field can be null.

suztomo: Removed "of" in "all of"

suztomo:
@elharo PTAL.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageErrorMatcher.java
```diff
@@ -16,26 +16,31 @@
...
 
   void setTargetMatcher(TargetMatcher targetMatcher) {
-    this.targetMatcher = targetMatcher;
+    this.targetMatcher = checkNotNull(targetMatcher);
   }
 
   /**
-   * Returns true if all of non-null matchers of {@link #sourceMatcher} and {@link #targetMatcher}
-   * match.
+   * Returns true if all non-null {@link #sourceMatcher} and {@link #targetMatcher} match.
```
elharo: There can be multiple source matchers and multiple target matchers? Or at most one of each?

suztomo: At most one of each.

elharo: That's not quite clear from the test. Perhaps something like "
   * Returns true if sourceMatcher matches "if present and the target matcher matches if present."

suztomo: Updated.

suztomo:
@elharo PTAL.

PR #1270, opened 2020-03-03
Adding page metadata, making index page dynamic

garrettjonesgoogle:
Also fixing some links.

This is the subset of https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1253 that adds page metadata and makes the index dynamic. I realized it would be easy to pull it out of the navigation PR, and that it would be useful on its own.

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,2 +1,4 @@
 _site/
 Gemfile.lock
+vendor/
+.bundle/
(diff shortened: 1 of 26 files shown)


garrettjonesgoogle:
Screenshot:

Screenshot from 2020-03-03 14-14-10

Note that there is one slight formatting change - the JLBP number is on a line before the title, instead of being part of the title.

PR #1280, opened 2020-03-05
Remove weak adverb

elharo:
@garrettjonesgoogle

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -6,7 +6,7 @@ permalink: /glossary
 - **Class path**: an ordered list of jar files, zip files, and directories, each of which
   contains Java class files.
   A [class loader](https://docs.oracle.com/javase/7/docs/api/java/lang/ClassLoader.html)
-  typically transforms the fully package qualified name of a Java class
+  transforms the fully package qualified name of a Java class
   into a file name and then searches for a class file with that name in a class path.
   
   When more than one entry in a class path contains a class file with the same name,


docs/glossary.md
```diff
@@ -6,7 +6,7 @@ permalink: /glossary
 - **Class path**: an ordered list of jar files, zip files, and directories, each of which
   contains Java class files.
   A [class loader](https://docs.oracle.com/javase/7/docs/api/java/lang/ClassLoader.html)
-  typically transforms the fully package qualified name of a Java class
+  transforms the fully package qualified name of a Java class
```
garrettjonesgoogle: Are there any circumstances where a classloader doesn't do this?

elharo: When it's not invoked, but even if so we don't need that. cf.

"The farmer plows the field"

we don't say, "The farmer typically plows the field" even though there are many other things a farmer does.

garrettjonesgoogle: I meant more in the context of the classloader doing stuff. Assuming it does get invoked, does it do something different than this? If not, "typically" shouldn't have ever been there. If so, I'm curious to know (even if it's not put into the content).

elharo: It's also going to load the class into the virtual machine, and do various error handling if it doesn't find the file, and so forth.

garrettjonesgoogle: Sure, but that's not excluded from the current wording.

Anyway, seems fine to proceed.

Issue #407, opened 2019-02-04
ClasspathChecker cannot find mockito 2's MockMethodDispatcher


suztomo:
While investigating #394, newer mockito also fails the check. Classpath Checker shows following errors for `-a org.mockito:mockito-core:2.23.4,junit:junit:4.12`

```
mockito-core-2.23.4.jar (18 errors):
  ClassSymbolReference{sourceClassName=org.mockito.internal.creation.bytebuddy.MockMethodAdvice$ForReadObject, targetClassName=org.mockito.internal.creation.bytebuddy.MockMethodDispatcher, subclass=false}, reason: CLASS_NOT_FOUND, target class location not found
  ClassSymbolReference{sourceClassName=org.mockito.internal.creation.bytebuddy.MockMethodAdvice$ForEquals, targetClassName=org.mockito.internal.creation.bytebuddy.MockMethodDispatcher, subclass=false}, reason: CLASS_NOT_FOUND, target class location not found
  ...

```

The class `MockMethodDispatcher` is included in mockito code ([github url](https://github.com/mockito/mockito/blob/30d3127ee2c4e41452787a058a31b7a553ba3b00/src/main/java/org/mockito/internal/creation/bytebuddy/MockMethodDispatcher.java)). However, the jar file contains ".raw" file:

```
~/cloud-opensource-java$ jar tf `find ~/.m2 -name 'mockito-core-2.23.4.jar'` |grep MockMethodDispatcher
org/mockito/internal/creation/bytebuddy/MockMethodDispatcher.raw

```

# Investigation

```
/tmp/mockito-2.23.4$ file ./org/mockito/internal/creation/bytebuddy/MockMethodDispatcher.raw
./org/mockito/internal/creation/bytebuddy/MockMethodDispatcher.raw: compiled Java class data, version 50.0 (Java 1.6)
```

The ".raw" file is just a class file renamed to avoid the class loaded by different class loaders than the one in Mockito's. https://github.com/mockito/mockito/pull/847 . Mockito's class loader handles ".raw" files in special manner.

suztomo:
Discussed with Elliotte, this corner case can be avoided by implementing specific handling for this class.

PR #1287, opened 2020-03-12
Linkage Checker to use exclusion XML file for default special cases

### LINKED ISSUE #1272: Exclusion file: use exclusion file to express special cases in Linkage Checker
Replace https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java#L198 with exclusion rule.

### LINKED ISSUE #1288: exclusion files: reason element
As per Elliotte in https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1287/files/a307b484bac633a92458ed8cf2d3aab12956d81c#r391851656

> The benefit of using elements and being flexible in what's allowed is that it enables people to fill needs we didn't plan for. It's the eXtensible part of XML. We don't show these today. Tomorrow, who knows? It didn't occur to me when you designed this format that we might want documentation like this, bit now I can see the need.

I think there are two topics here: reason element (v.s. XML comment), and reason element having HTML elements.

# Reason element

# Allowing "reason" elements to have HTML elements

suztomo:
Fixes #1272 

Fixes #1288

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileParser.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileParser.java
@@ -20,6 +20,7 @@ import com.google.common.collect.ImmutableList;
 import com.thaiopensource.xml.sax.DraconianErrorHandler;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import org.iso_relax.verifier.Schema;
@@ -69,13 +70,29 @@ class ExclusionFileParser {
   static ImmutableList<LinkageErrorMatcher> parse(Path exclusionFile)
       throws SAXException, IOException, VerifierConfigurationException {
 
+    InputSource inputSource = new InputSource(Files.newInputStream(exclusionFile));
+    inputSource.setSystemId(exclusionFile.toUri().toString());
+
+    return parse(inputSource);
+  }
+
+  static ImmutableList<LinkageErrorMatcher> parse(URL exclusionFile)
+      throws SAXException, IOException, VerifierConfigurationException {
+
+    InputSource inputSource = new InputSource(exclusionFile.openStream());
+    inputSource.setSystemId(exclusionFile.toString());
+
+    return parse(inputSource);
+  }
+
+  private static ImmutableList<LinkageErrorMatcher> parse(InputSource inputSource)
+      throws SAXException, IOException, VerifierConfigurationException {
+
     XMLReader reader = createXmlReader();
 
     ExclusionFileHandler handler = new ExclusionFileHandler();
     reader.setContentHandler(handler);
 
-    InputSource inputSource = new InputSource(Files.newInputStream(exclusionFile));
-    inputSource.setSystemId(exclusionFile.toUri().toString());
     reader.parse(inputSource);
 
     return handler.getMatchers();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
@@ -28,7 +28,9 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Multimaps;
 import com.google.common.collect.SetMultimap;
 import java.io.IOException;
+import java.net.URISyntaxException;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayDeque;
 import java.util.Arrays;
 import java.util.HashSet;
@@ -43,6 +45,8 @@ import org.apache.bcel.classfile.FieldOrMethod;
 import org.apache.bcel.classfile.JavaClass;
 import org.apache.bcel.classfile.Method;
 import org.eclipse.aether.artifact.Artifact;
+import org.iso_relax.verifier.VerifierConfigurationException;
+import org.xml.sax.SAXException;
 
 /** A tool to find linkage errors in a class path. */
 public class LinkageChecker {
@@ -58,6 +62,7 @@ public class LinkageChecker {
   private final ImmutableList<Path> jars;
   private final SymbolReferenceMaps classToSymbols;
   private final ClassReferenceGraph classReferenceGraph;
+  private final ImmutableList<LinkageErrorMatcher> exclusionMatchers;
 
   @VisibleForTesting
   SymbolReferenceMaps getClassToSymbols() {
@@ -79,7 +84,10 @@ public class LinkageChecker {
     ClassReferenceGraph classReferenceGraph =
         ClassReferenceGraph.create(symbolReferenceMaps, ImmutableSet.copyOf(entryPoints));
 
-    return new LinkageChecker(dumper, jars, symbolReferenceMaps, classReferenceGraph);
+    ImmutableList<LinkageErrorMatcher> exclusionMatchers = readExclusionMatchers();
+
+    return new LinkageChecker(
+        dumper, jars, symbolReferenceMaps, classReferenceGraph, exclusionMatchers);
   }
 
   public static LinkageChecker create(Bom bom) throws IOException {
@@ -99,18 +107,21 @@ public class LinkageChecker {
 
   @VisibleForTesting
   LinkageChecker cloneWith(SymbolReferenceMaps newSymbolMaps) {
-    return new LinkageChecker(classDumper, jars, newSymbolMaps, classReferenceGraph);
+    return new LinkageChecker(
+        classDumper, jars, newSymbolMaps, classReferenceGraph, exclusionMatchers);
   }
 
   private LinkageChecker(
       ClassDumper classDumper,
       List<Path> jars,
       SymbolReferenceMaps symbolReferenceMaps,
-      ClassReferenceGraph classReferenceGraph) {
+      ClassReferenceGraph classReferenceGraph,
+      ImmutableList<LinkageErrorMatcher> exclusionMatchers) {
     this.classDumper = Preconditions.checkNotNull(classDumper);
     this.jars = ImmutableList.copyOf(jars);
     this.classReferenceGraph = Preconditions.checkNotNull(classReferenceGraph);
     this.classToSymbols = Preconditions.checkNotNull(symbolReferenceMaps);
+    this.exclusionMatchers = Preconditions.checkNotNull(exclusionMatchers);
   }
 
   /**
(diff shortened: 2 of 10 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFileParser.java
```diff
@@ -69,13 +70,29 @@
   static ImmutableList<LinkageErrorMatcher> parse(Path exclusionFile)
       throws SAXException, IOException, VerifierConfigurationException {
 
+    InputSource inputSource = new InputSource(Files.newInputStream(exclusionFile));
+    inputSource.setSystemId(exclusionFile.toUri().toString());
+
+    return parse(inputSource);
+  }
+
+  static ImmutableList<LinkageErrorMatcher> parse(URL exclusionFile)
```
suztomo: `linkage-checker-exclusion-default.xml` is a file in a JAR file when running as enforcer rule. So picking up URL rather than Path.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -186,39 +196,38 @@ private LinkageChecker(
...
+    return exclusionMatchers.build();
+  }
+
   /**
    * Returns true if the linkage error {@code entry} should be reported. False if it should be
    * suppressed.
    */
-  private static boolean problemFilter(Map.Entry<SymbolProblem, ClassFile> entry) {
-    ClassFile classFile = entry.getValue();
+  private boolean problemFilter(Map.Entry<SymbolProblem, ClassFile> entry) {
```
suztomo: This method is non-static because exclusionMatchers field is an instance variable.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -186,39 +196,38 @@ private LinkageChecker(
...
-            || sourceClassName.startsWith("com.oracle.graal")
-            || sourceClassName.startsWith("org.graalvm"))) {
-      return false;
-    }
 
-    // Mockito's MockMethodDispatcher uses special class loader to load MockMethodDispatcher.raw
-    // https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/407
-    if (problematicClassName.equals("org.mockito.internal.creation.bytebuddy.MockMethodDispatcher")
-        && sourceClassName.startsWith("org.mockito.internal.creation.bytebuddy")) {
-      return false;
```
suztomo: These Java logic is replaced with the XML file below.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -99,18 +106,21 @@ public static LinkageChecker create(Bom bom) throws IOException {
...
+        classDumper, jars, newSymbolMaps, classReferenceGraph, exclusionMatchers);
   }
 
   private LinkageChecker(
       ClassDumper classDumper,
       List<Path> jars,
       SymbolReferenceMaps symbolReferenceMaps,
-      ClassReferenceGraph classReferenceGraph) {
+      ClassReferenceGraph classReferenceGraph,
+      ImmutableList<LinkageErrorMatcher> exclusionMatchers) {
```
elharo: List feels wrong here, as if it's exposing implementation details. Would it make sense to have a single object that contains a list, rather than a bunch of matchers? See comment below.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -186,39 +196,41 @@ private LinkageChecker(
...
-            || sourceClassName.startsWith("org.graalvm"))) {
-      return false;
-    }
 
-    // Mockito's MockMethodDispatcher uses special class loader to load MockMethodDispatcher.raw
-    // https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/407
-    if (problematicClassName.equals("org.mockito.internal.creation.bytebuddy.MockMethodDispatcher")
-        && sourceClassName.startsWith("org.mockito.internal.creation.bytebuddy")) {
-      return false;
+    for (LinkageErrorMatcher matcher : exclusionMatchers) {
```
elharo: This list iteration logic could become part of an ExcludedErrors class. Then it becomes simply 

return excludedErrors.contains(symbolProblem, sourceClass)

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
```diff
@@ -0,0 +1,50 @@
...
+    https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/816 -->
+  <LinkageError>
+    <Source>
+      <Package name="reactor.core.publisher.Traces" />
+    </Source>
+  </LinkageError>
+
+  <!-- GraalVM-related libraries depend on Java Compiler Interface (JVMCI) that
+    only exists in special JDK.
+    https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/929 -->
```
elharo: Expand comment to explain why this is not a problem.

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
```diff
@@ -0,0 +1,50 @@
...
+  <LinkageError>
+    <Target>
+      <Package name="jdk.vm.ci" />
+    </Target>
+    <Source>
+      <Package name="org.graalvm" />
+    </Source>
+  </LinkageError>
+
+  <!-- Mockito's MockMethodDispatcher uses special class loader to load
```
elharo: The comment should explain why we don't care about this. In fact this doesn't need to be a comment. It can be another element, and include full HTML markup (if that's useful).

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
suztomo: Added explanation. It's not "why this is not a problem" but "why Linkage Checker reports false positives".

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
suztomo: Added explanation why the errors are false positives.

I just thought about adding `<reason>` element that would enable HTML markup. However it does not help users because the exclusion rules are for errors that won't be shown to users, nor in an HTML renderer. If you still think `<reason>` element helpful, feel free to file an issue. I'll check that later.

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
elharo: The benefit of using elements and being flexible in what's allowed is that it enables people to fill needs we didn't plan for. It's the eXtensible part of XML.  We don't show these today. Tomorrow, who knows? It didn't occur to me when you designed this format that we might want documentation like this, bit now I can see the need.

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
```diff
@@ -0,0 +1,53 @@
...
+    MockMethodDispatcher.raw. As Linkage Checker does not use the class loader,
+    it ignores the missing class errors to avoid false positives.
+    https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/407 -->
```
elharo: There's no reason not to summarize the relevant parts of #407 here. I think there's more content there than in this file.

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
elharo: Also, I'm not sure this is correct. The reason is not that "As Linkage Checker does not use GraalVM, it ignores the errors to avoid false positives." Even if we did use GraalVM, I suspect this would happen. What we need here is an answer to the question of why do we consider this a false positive, and that answer is that the code is only invoked when running in a GraalVM that does have those classes.  In a non-Graal VM it's unreachable.

suztomo: Created an issue: https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1288 (not in this PR)


> Even if we did use GraalVM, I suspect this would happen.

If we use Java runtime that provides JVMCI Java Compiler Interface (JVMCI), this missing classes won't happen.

>  the code is only invoked when running in a GraalVM

Updated so.

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
suztomo: Created https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1288

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
suztomo: Added the raw file is in the mockito JAR.

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
```diff
@@ -0,0 +1,55 @@
...
+      <Package name="jdk.vm.ci" />
+    </Target>
+    <Source>
+      <Package name="org.graalvm" />
+    </Source>
+  </LinkageError>
+
+  <!-- Mockito's MockMethodDispatcher uses special class loader to load
+    MockMethodDispatcher.raw. The class file with "raw" suffix is in mockito-core JAR file.
+    As long as developers use Mockito through a normal usage, there should not be
```
elharo: delete the sentence beginning "As long as"

suztomo: Deleted.

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
elharo: I'm not OK with leaving this for later. This PR needs that information, and that information belongs in an element, not a comment.

suztomo:
@elharo Would you be willing to explain why you think the new element definition, such as 4 items in https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1288#issuecomment-598911102, should be part of this PR "Linkage Checker to use exclusion XML file for default special cases" ?

If you prefer not to have the 3 comments in the XML file, I'll close this PR and we can focus on the discussion in #1288 first.

elharo:
I'm holding the line on this one. #1288 should not be a separate issue. You've uncovered a need for extra content in this document that we did not originally plan for. That's nor a surprise. Including newly discovered information in existing formats is precisely what XML was designed to make possible. Add the new element or elements (probably `reason`) to mark up  newly added information as you discover a need for it. 

This would be slightly easier to manage if the schema validation wasn't part of the processing chain, but it's still not particularly onerous. Using comments like this will make future processing of this information more difficult.

suztomo:
@elharo For adding new element in the Relax NG schema, would you share your thought on the element?

- Element tag name. `<reason>`?
- whether the comment element is to be a child of LinkageError, Source, and/or Target. 
- whether they are optional, one-or-more, or exactly one.
- whether they allow HTML elements in them.

elharo:
`documentation` is another common name for this. I sort of like `reason` here though. That is, the reason the linkage error is excluded. In which case it should be a child of `LinkageError` because that's what it's the reason for.

Don't overthink the schema. Just pick the simplest thing we need right now, a single reason per linkage error that contains plain text. We can expand what's allowed later if we need more.

suztomo:
Thank you for explanation. I'll create `reason` element with optional one element in the schema. It's optional because we don't know the content when linkage error matchers are automatically generated.

Will update this PR shortly.

suztomo:
@elharo PTAL.

PR #1294, opened 2020-03-17
add alt attributes

### LINKED ISSUE #1285: docs/why-doesnt-the-compiler-catch-diamond-dependency-conficts.md needs alt attributes
some of these are well enough described in the surrounding text we can set `alt=""`

elharo:
fixes #1285 
@suztomo In this case, I think the text is complete without the pictures so with slight rewrites, we can use empty `alt` attributes.


suztomo:
Making the text complete without a picture itself is a good thing.

Why do we want to add an empty alt attribute?

PR #1299, opened 2020-03-18
limit information in CLI error printing

elharo:
@suztomo The goal here is not to dump a stack trace or other random content, just print what actually went wrong. Also this uses stderr instead of stdout.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerArguments.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerArguments.java
@@ -27,7 +27,6 @@ import java.util.Arrays;
 import org.apache.commons.cli.CommandLine;
 import org.apache.commons.cli.CommandLineParser;
 import org.apache.commons.cli.DefaultParser;
-import org.apache.commons.cli.HelpFormatter;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.OptionGroup;
 import org.apache.commons.cli.Options;
@@ -53,7 +52,6 @@ import org.eclipse.aether.artifact.DefaultArtifact;
 final class LinkageCheckerArguments {
 
   private static final Options options = configureOptions();
-  private static final HelpFormatter helpFormatter = new HelpFormatter();
 
   private final CommandLine commandLine;
   private ImmutableList<Artifact> cachedArtifacts;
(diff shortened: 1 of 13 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerArguments.java
```diff
@@ -84,9 +82,6 @@ static LinkageCheckerArguments readCommandLine(String... arguments) throws Parse
       return new LinkageCheckerArguments(parser.parse(options, arguments));
     } catch (IllegalArgumentException ex) {
       throw new ParseException("Invalid URL syntax in Maven repository URL");
-    } catch (ParseException ex) {
-      helpFormatter.printHelp("LinkageChecker", options);
```
suztomo: Does the command still shows available options when invoked without any argument?

elharo: It does not. We should distinguish help from an error condition. The most common convention seems to be that -h or --help prints help.

suztomo: Common behaviors print the help when a required argument is missing. Would you update this PR to do so?

elharo: I added a --help/-h option. With that there are no required arguments any more.

Also, I think there's an existing bug where jar files can't be selected.

suztomo: I just ran LinkageCheckerMain without any argument and it existed 0 without showing anything. The empty output does not tell what option I have to add LinkgeCheckerMain.  Would you fix that?

suztomo: > Common behaviors print the help when a required argument is missing.

I'm referring `java` or `javac` command without any argument, for example.

elharo: I'll see your javac and raise you a gcc:

```
$ gcc
gcc: fatal error: no input files
compilation terminated.
```

but sure, whatever.

Issue #1296, opened 2020-03-17
Investigate gRPC 1.28 Linkage Errors


suztomo:
Followup of 
https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1286#issuecomment-598196225

```
Class io.reactivex.internal.schedulers.NonBlockingThread is not found
Class reactor.core.scheduler.NonBlocking is not found
Class com.sun.jna.Native is not found
Class com.sun.jna.Structure is not found
Class com.sun.jna.platform.win32.WinNT$HANDLE is not found
Class com.sun.jna.ptr.IntByReference is not found
Class com.sun.jna.platform.win32.Kernel32 is not found
Class com.sun.jna.platform.win32.Win32Exception is not found
Class com.sun.jna.Platform is not found
...
```

Todo: find piece of  source code that protects users from NoClassDefFoundError. Very likely it's BlockHound's code using reflection or try-catch surrounding to find the existence of classes.

# Dashboard

Errors in grpc-netty

<img width="760" alt="Screen Shot 2020-03-19 at 10 16 07" src="https://user-images.githubusercontent.com/28604/77076816-bccbbb00-69ca-11ea-89a6-aa7718d1c879.png">

<img width="832" alt="Screen Shot 2020-03-19 at 10 19 41" src="https://user-images.githubusercontent.com/28604/77077127-31065e80-69cb-11ea-98a3-4bf385e46fc0.png">

Gist: https://gist.github.com/suztomo/bbfa68b4259e9fbd7484b900db90d09e

```
Dependency path 'io.grpc:grpc-netty > io.netty:netty-common > io.projectreactor.tools:blockhound' exists in all 80 dependency paths. Example path:
 io.grpc:grpc-netty:1.28.0 (compile)
 / io.netty:netty-codec-http2:4.1.45.Final (compile)
 / io.netty:netty-common:4.1.45.Final (compile)
 / io.projectreactor.tools:blockhound:1.0.1.RELEASE (compile, optional)
```

##  Classes to examine

The following 3 classes are the sources of the references.

- `reactor.blockhound.integration.RxJava2Integration`

- `reactor.blockhound.integration.ReactorIntegration`

- `reactor.blockhound.shaded.net.bytebuddy.agent.VirtualMachine`

suztomo:
# reactor.blockhound.integration.RxJava2Integration

This implements ClassNotFoundException guard:

```
[...]

PR #1302, opened 2020-03-19
Suppress Linkage Errors from classes that check ClassNotFoundException

suztomo:
From the investigation of #1296 . Some classes in BlockHound checks whether certain classes are available or not before using them.


suztomo:
Looking into dashboard test failure.

Issue #1179, opened 2020-02-03
Linkage Checker: Annotated JAR file class


suztomo:
As of now, Linkage Checker uses `ImmutableList<Path>` to represent class paths. However, a `java.nio.filePath` instance does not hold Maven artifact information. Because of this missing information, Linkage Checker cannot perform operation based on Maven coordinates.

A class that holds an optional Maven artifact and Path, when used in LinkageChecker, provides better utility:

- SymbolProblem.toString can use the Maven artifact coordinates rather than JAR file.
  As of now, the error is based on jar file, not coordinates (see [dashboard](https://storage.googleapis.com/cloud-opensource-java-dashboard/com.google.cloud/libraries-bom/snapshot/com.google.http-client_google-http-client-appengine_1.34.1.html))
- Linkage Checker can filter problems by Maven artifact coordinates
  This is related to https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1167

Note that Maven coordinates are optional for the class to support LinkageChecker for list of JAR files without Maven coordinates.

Credit: Elliotte.

PR #1303, opened 2020-03-19
AnnotatedJar class to replace java.nio.Path

suztomo:
Towards #1179 . Once this is merged, I'll create another PR to replace `ImmutableList<Path>` with `ImmutableList<AnnotatedJar>` in many places.

Linkage Checker will be able to use the artifact part to filter linkage errors based on Maven artifact coordinates.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.opensource.classpath;
+
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.nio.file.Path;
+import java.util.Objects;
+import javax.annotation.Nullable;
+import org.eclipse.aether.artifact.Artifact;
+
+/**
+ * JAR file annotated with an optional Maven artifact.
+ */
+final class AnnotatedJar {
+  private Path jar;
+
+  private Artifact artifact;
+
+  AnnotatedJar(Path jar, @Nullable Artifact artifact) {
+    this.jar = checkNotNull(jar);
+    this.artifact = artifact;
+  }
+
+  AnnotatedJar(Path jar) {
+    this(jar, null);
+  }
+
+  /**
+   * The Maven artifact that holds the JAR file, if the JAR is downloaded for a Maven artifact.
+   * {@code Null} if the JAR file is not from a Maven artifact.
+   */
+  @Nullable
+  Artifact getArtifact() {
+    return artifact;
+  }
+
+  /**
+   * Path to the JAR file.
+   */
+  Path getJar() {
+    return jar;
+  }
+
+  @Override
+  public boolean equals(Object other) {
+    if (this == other) {
+      return true;
+    }
+    if (other == null || getClass() != other.getClass()) {
+      return false;
+    }
+    AnnotatedJar that = (AnnotatedJar) other;
+    return Objects.equals(jar, that.jar) &&
+        Objects.equals(artifact, that.artifact);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(jar, artifact);
+  }
+}
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
@@ -39,15 +39,15 @@ final class AnnotatedJar {
   }
 
   /**
-   * The Maven artifact that holds the JAR file, if the JAR is downloaded for a Maven artifact.
-   * {@code Null} if the JAR file is not from a Maven artifact.
+   * Returns the Maven artifact that holds the JAR file, if the JAR is downloaded for a Maven
+   * artifact. {@code Null} if the JAR file is not from a Maven artifact.
    */
   @Nullable
   Artifact getArtifact() {
     return artifact;
   }
 
-  /** Path to the JAR file. */
+  /** Returns path to the JAR file. */
   Path getJar() {
     return jar;
   }
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
@@ -16,7 +16,6 @@
 
 package com.google.cloud.tools.opensource.classpath;
 
-
 import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
```diff
@@ -0,0 +1,71 @@
...
+    this.artifact = artifact;
+  }
+
+  AnnotatedJar(Path jar) {
+    this(jar, null);
+  }
+
+  /**
+   * The Maven artifact that holds the JAR file, if the JAR is downloaded for a Maven artifact.
+   * {@code Null} if the JAR file is not from a Maven artifact.
```
elharo: null

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
```diff
@@ -0,0 +1,71 @@
...
+   */
+  @Nullable
+  Artifact getArtifact() {
```
elharo: The getters aren't tested. Writing tests first avoids unnecessary code.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
suztomo: I think it's capital N because this is not middle of a sentence.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
suztomo: Added test.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
```diff
@@ -0,0 +1,71 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import java.nio.file.Path;
+import java.util.Objects;
+import javax.annotation.Nullable;
+import org.eclipse.aether.artifact.Artifact;
+
+/** JAR file annotated with an optional Maven artifact. */
+final class AnnotatedJar {
```
elharo: What do you think about renaming this `CachedArtifact` or `LocalArtifact` instead? That is, it's an artifact that has a JAR rather than a JAR that has coordinates?

Though now that I look deeper, `org.eclipse.aether.artifact.Artifact` has a `File` property that's serving the same purpose as the path here. Can we just use that? Then we don't need to add this class at all.

suztomo: The idea of _not_ using `org.eclipse.aether.artifact.Artifact` is to support Linkage Checker invocation without Maven artifacts. That is `LinkageCheckerMain --jar A.jar,B.jar`.

elharo: In that case we could use an Artifact with group ID, artifact ID, etc. st to null:

https://maven.apache.org/resolver/apidocs/org/eclipse/aether/artifact/DefaultArtifact.html

suztomo: That would technically work. However, I don't think it's good idea to use Artifact class to represent something that is not a Maven artifact.

elharo: I suppose it's more of a generalized thing that includes the class of Maven and non-Maven artifacts. Or it's a special kind of artifact that has a file and null coordinates. It does feel like this class is redundant.

suztomo:
Memo:

This `AnnotatedJar` does not know the location (or a Maven dependency node) in a dependency graph. Class `org.eclipse.aether.artifact.Artifact` does not carry information such as dependencies or its parent in a tree.

If we happen to need it, we can add `DependencyPath` field to this `AnnotatedJar` later.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
suztomo: @elharo What do you think about dropping bare JAR support? I've implemented that but did not encounter any need to perform operations for bare JAR files; when I use Linkage Checker, it's always through Maven artifacts.

suztomo:
When designing the class, it's nice to think about how the usage looks like:

Current:
```
    List<Path> paths = ImmutableList.of(guavaPath);
    LinkageChecker linkageChecker = LinkageChecker.create(paths,...);
```

With AnnotatedJar

```
    List<AnnotatedJar> paths = ImmutableList.of(new AnnotatedJar(guavaPath));
    LinkageChecker linkageChecker = LinkageChecker.create(paths,...);
```

With [DefaultArtifact](https://maven.apache.org/resolver/apidocs/org/eclipse/aether/artifact/DefaultArtifact.html#DefaultArtifact(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20java.util.Map,%20java.io.File))

```
    List<Artifact> paths = ImmutableList.of(new DefaultArtifact(null, null, null, null, null, null, guavaPath.toFile()));
    LinkageChecker linkageChecker = LinkageChecker.create(paths,...);
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
suztomo: I checked [DefaultArtifact constructor that takes a file](https://maven.apache.org/resolver/apidocs/org/eclipse/aether/artifact/DefaultArtifact.html#DefaultArtifact(java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20java.util.Map,%20java.io.File)). 6 out of 7 arguments would null if we only store a file there. And it's `java.io.File`, not `java.nio.file.Path`.  I added an example usages of AnnotatedJar and DefaultArtifact in [a comment](https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1303#issuecomment-601869672). 

With the reason above, I prefer this AnnotatedJar to DefaultArtifact.

elharo:
The constructor's a little ugly for DefaultArtifact, but that could be hidden in a helper method or subclass, easily enough; and most of the time that's not how we'll create this anyway. Of the three return types `List<Artifact>` seems by far the most obvious.

If you want to proceed with AnnotatedJar, then let's not stop at this PR. We should see it in context. That is try replacing the current List<Path> with List<AnnotatedJar>. I predict what you'll see is that List<Artifact> is even cleaner and more obvious.

suztomo:
`org.eclipse.aether.artifact.DefalutArtifact` is final. No subclasses. Instead, we can create this:

```
/** JAR file without a Maven artifact **/
final class JarArtifact implements Artifact {
...
```

>  try replacing the current `List<Path>` with `List<AnnotatedJar>`

I did some already. Let me raise a PR to check.

suztomo:
@elharo Created a draft PR to show how AnnotatedJar would look like: https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1305

suztomo:
I don't like `AnnotatedJar`'s name. Maybe `ClassPathItem` or `ClassPathElement`? Because we always use the class in context of a class path.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AnnotatedJar.java
elharo: Think of it this way. This is inside out and redundant. All the information we need is in Artifact. If you add a third field for Dependency or DependencyNode, then all the information is in there. But this type doesn't add any new information. It's not like we're adding things that aren't in the artifact like download time or something like that. Does that make sense?

suztomo: That's a great explanation. Now I feel the redundancy of AnnotatedJar clearly when we use Linkage Checker for Maven artifacts.

How about cases without Maven artifacts (bare JAR files)? Do you have rationale for using Maven's Artifact class for JAR files that do not carry `groupId`, `artifactId`, or `version`?

If we had algebraic data types, I could just write:

```
data ClassPathElement = Path | Artifact  
```

suztomo:
Closing this in favor of #1308

PR #1309, opened 2020-03-24
docs: remove table

elharo:
@suztomo I removed a section that is obsolete and not actionable. No one needs to know which now very old version first appeared in a BOM.


suztomo:
I still think the BOM reference section valuable to know available BOMs for Google Cloud.

As you commented in this PR description, I agree that the readers might not care "First available version".

elharo (commented):
Ping. This content has gotten even more out of date in the last year. We should remove it.

suztomo (approved):
I agree with the removal now. The table explained our engagement with the owners but it doesn't help the readers.

PR #1314, opened 2020-03-27
push class file names into ClasspathEntry

elharo:
some initial work on pushing more info into classes instead of parallel data structures

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -190,7 +190,7 @@ public class DashboardMain {
     List<ClassPathEntry> artifactJarsInBom = classpath.subList(0, managedDependencies.size());
     ImmutableSet<ClassPathEntry> entryPoints = ImmutableSet.copyOf(artifactJarsInBom);
     for (ClassPathEntry entry : entryPoints) {
-      entry.listClassFileNames();
+      entry.loadClassFileNames();
     }
     
     LinkageChecker linkageChecker = LinkageChecker.create(classpath, entryPoints);
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -189,7 +189,10 @@ public class DashboardMain {
     // When checking a BOM, entry point classes are the ones in the artifacts listed in the BOM
     List<ClassPathEntry> artifactJarsInBom = classpath.subList(0, managedDependencies.size());
     ImmutableSet<ClassPathEntry> entryPoints = ImmutableSet.copyOf(artifactJarsInBom);
-
+    for (ClassPathEntry entry : entryPoints) {
+      entry.listClassFileNames();
+    }
+    
     LinkageChecker linkageChecker = LinkageChecker.create(classpath, entryPoints);
 
     ImmutableSetMultimap<SymbolProblem, ClassFile> symbolProblems =
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -186,13 +186,6 @@ public class DashboardMain {
     // LinkedListMultimap preserves the key order
     ImmutableList<ClassPathEntry> classpath = classPathResult.getClassPath();
 
-    // When checking a BOM, entry point classes are the ones in the artifacts listed in the BOM
-    List<ClassPathEntry> artifactJarsInBom = classpath.subList(0, managedDependencies.size());
-    ImmutableSet<ClassPathEntry> entryPoints = ImmutableSet.copyOf(artifactJarsInBom);
-    for (ClassPathEntry entry : entryPoints) {
-      entry.loadClassFileNames();
-    }
-
     LinkageChecker linkageChecker = LinkageChecker.create(classpath);
 
     ImmutableSetMultimap<SymbolProblem, ClassFile> symbolProblems =
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -87,10 +87,6 @@ class ClassDumper {
             .collect(toImmutableList());
     checkArgument(
         unreadableFiles.isEmpty(), "Some jar files are not readable: %s", unreadableFiles);
-
-    for (ClassPathEntry entry : entries) {
-      entry.loadClassFileNames();
-    }
     
     return new ClassDumper(entries, extensionClassLoader);
   }
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -341,7 +341,7 @@ class ClassDumper {
       List<ClassPathEntry> classPath) throws IOException {
     Builder<ClassPathEntry, String> pathToClasses = ImmutableSetMultimap.builder();
     for (ClassPathEntry jar : classPath) {
-      for (String classFileName : jar.listClassFileNames()) {
+      for (String classFileName : jar.getClassFileNames()) {
         pathToClasses.put(jar, classFileName);
       }
     }
@@ -357,7 +357,7 @@ class ClassDumper {
 
     ImmutableList.Builder<String> corruptedClassFileNames = ImmutableList.builder();
 
-    for (String classFileName : entry.listClassFileNames()) {
+    for (String classFileName : entry.getClassFileNames()) {
       if (classFileName.startsWith("META-INF.versions.")) {
         // Linkage Checker does not support multi-release JAR (for Java 9+) yet
         // https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/897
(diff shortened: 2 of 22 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathEntry.java
```diff
@@ -48,22 +48,18 @@
 
   /** An entry for a Maven artifact. */
   public ClassPathEntry(Artifact artifact) {
-    checkNotNull(artifact.getFile());
+    this(artifact.getFile().toPath());
```
suztomo: This defies line 39 "Either jar or artifact is non-null".

elharo: removed

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -189,7 +189,10 @@ private static Path generate(Bom bom) throws IOException, TemplateException, URI
     // When checking a BOM, entry point classes are the ones in the artifacts listed in the BOM
     List<ClassPathEntry> artifactJarsInBom = classpath.subList(0, managedDependencies.size());
     ImmutableSet<ClassPathEntry> entryPoints = ImmutableSet.copyOf(artifactJarsInBom);
-
+    for (ClassPathEntry entry : entryPoints) {
+      entry.loadClassFileNames();
+    }
+
```
suztomo, line 195: I feel this piece of code is wrong.  It is as if `classPathResult.getClassPath()` has a contract that its caller needs to call `loadClassFileNames()` for each elements.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathEntry.java
```diff
@@ -96,27 +103,31 @@ public String toString() {
...
 
-    return classPath.getAllClasses().stream().map(ClassInfo::getName).collect(toImmutableSet());
+    classFileNames =
+        classPath.getAllClasses().stream().map(ClassInfo::getName).collect(toImmutableSet());
+  }
+  
+  /**
+   * @return the fully qualified names of the classes in this entry's jar file
+   */
+  public ImmutableSet<String> getClassNames() {
```
suztomo, line 130: How about calling `loadClassFileNames` and cache the result?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
elharo: I don't like this either I haven't found something I like better yet though. I was hoping you might have an idea.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathEntry.java
elharo: I tried this. The problem is it introduces an IOException on this method, which both feels wrong and makes a lot of calling code very ugly. 

I'm not sure how to proceed. Suggestions appreciated.

elharo: Maybe the right place to do this is in the constructor. It's called much less frequently than `getClassNames` so adding an IOException there would be less invasive.

suztomo:
For making the JAR-reading at instance creation,
- factory methods are better choice as in  "Avoid doing work in constructors" go/java-practices/constructors#avoid-doing-work-in-constructors . (Same 
- It makes tests harder. I asssume `ClassPathEntry.of("foo:bar:1.2.3", "dummy.jar")` starts throwing IOException.


I'm feeling getClassNames should do the IO work.

>  The problem is it introduces an IOException on this method, which both feels wrong and makes a lot of calling code very ugly.

As the original method `mapJarToClassFileNames` (being removed) has IOException, it shouldn't create new problems with IOExceptions. I'm checking how it looks like.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -73,7 +69,6 @@
   private final ImmutableList<ClassPathEntry> inputClassPath;
   private final FixedSizeClassPathRepository classRepository;
   private final ClassLoader extensionClassLoader;
-  private final ImmutableSetMultimap<ClassPathEntry, String> classPathEntryToClassFileNames;
   private final ImmutableListMultimap<String, ClassPathEntry> classFileNameToClassPathEntry;
```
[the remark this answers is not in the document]
elharo: As much as possible needs to be pushed into ClasspathEntry, and we need to start replacing all these parallel multimaps with object types. The existing structure is brittle and hard to follow.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -92,19 +87,25 @@ static ClassDumper create(List<ClassPathEntry> entries) throws IOException {
             .collect(toImmutableList());
     checkArgument(
         unreadableFiles.isEmpty(), "Some jar files are not readable: %s", unreadableFiles);
-
-    return new ClassDumper(entries, extensionClassLoader, mapJarToClassFileNames(entries));
```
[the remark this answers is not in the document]
elharo: Let me think about that. There might be another way to do this, but it can't involve maintaining multiple multimaps.

I do like the principle of keeping work out of constructors. However I don't believe one should contort the natural design of a class to support testing and mocking. That leads to both bad design and tests that don't actually test the real code. Most of the time more effort spent in test setup avoid the need for mocking.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathEntry.java
```diff
@@ -118,4 +121,11 @@ public static ClassPathEntry of(String coordinates, String filePath) {
 
     return classPath.getAllClasses().stream().map(ClassInfo::getName).collect(toImmutableSet());
```
suztomo, line 122: Memo: this `classPath.getAllClasses()` returns empty if `jar` does not exist.
Screen Shot 2020-04-02 at 11 42 39

suztomo: If we start reading this JAR files without Guava's ClassPath (https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1320 ). I think appropriate behavior is to throw IOException upon nonexistent JAR file.

The existing tests didn't fail for this PR because Guava's ClassPath happened to ignore nonexistent JAR file.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -73,7 +69,6 @@
   private final ImmutableList<ClassPathEntry> inputClassPath;
   private final FixedSizeClassPathRepository classRepository;
   private final ClassLoader extensionClassLoader;
-  private final ImmutableSetMultimap<ClassPathEntry, String> classPathEntryToClassFileNames;
```
suztomo, line 76: Memo: where does LinkageChecker use this ?

It's used by `classDumper.classNamesInJar()`. This method is used by Linkage Checker:

```
    classToFieldSymbols.forEach(
        (classFile, fieldSymbol) -> {
          if (!classDumper
              .classNamesInJar(classFile.getClassPathEntry())
              .contains(fieldSymbol.getClassBinaryName())) {
            findSymbolProblem(classFile, fieldSymbol)
                .ifPresent(problem -> problemToClass.put(problem, classFile.topLevelClassFile()));
          }
        });
```

This if-statement checks whether the source class and target class of a symbol reference are in the same JAR file or not. If they are in the same JAR file, Linkage Checker skips the references because such javac would not compile to generate such JAR files.

suztomo: Therefore the result of class file name reading should be cached (this PR does that)

suztomo:
For 1-3, they make sense. Although I liked symmetry of many-to-many relationship, I also see the benefit of pushing class file names to ClassPathEntry. Now the discussion between reading JAR in a constructor or a factory method.

For 4,

> We should not contort the design of a class to make testing easier.

We should aim to avoid making tests less testable. Reading JAR files in the ClassPathEntry constructor makes the class less testable (below):

> This PR does not make this class any less testable. All existing tests pass.

The existing tests didn't fail for this PR because Guava's ClassPath _happened to_ not raise an IOException upon nonexistent JAR files.

If we start reading this JAR files without Guava's ClassPath (#1320 ), then the constructor starts throwing IOException (I strongly believe this is correct behavior of the `readClassFileNames `). In tests with "dummy.jar", the class would require mockito or some special handling to avoid raising IOException. ["Avoid doing work in constructors
"](https://engdoc.corp.google.com/eng/doc/devguide/java/practices/constructors.md?cl=head#avoid-doing-work-in-constructors ) explains simple constructors let us write tests without mockito.

For 5,

the discussion in "Avoid doing work in constructors
"](https://engdoc.corp.google.com/eng/doc/devguide/java/practices/constructors.md?cl=head#avoid-doing-work-in-constructors ) does not require subclasses. It's fine to use static factory method without subclasses.

suztomo:
Surveyed the design of other classes:

- No IO at constructor: `java.nio.file.Path`, `java.io.File`, `DefaultArtifact`
- Potential IOException at constructor: `java.util.jar.JarFile`.

I recognize ClassPathEntry is similar to the former group. The creation of the class does not mean reading the content. The former group may have methods (or helper class) to read their content but not at the constructor.

elharo:
I don;t think you can compare java.io.File to this class. Java.io.File has always been a bit of a weird duck, going back to Java 1.0, and was so badly designed Oracle had to go back to the drawing board and do it again with Path years later.

I think we want ClasspathEntry to actually exist, not simply be a pointer to something that might or might not exist. 

Ultimately there will be a point at which we find out a file can't be read and need to handle the failure. That's unescapable. The question is when is it most convenient to recognize that? About the earliest we can do that is when the ClasspathEntry is constructed, and indeed we already fail fast at that point if the jar is missing, just with a NullPointerException instead of an IOException. The latest we can do that is when we're trying to list out the contents of the jar, which is essentially what we're doing now. 

Failing early is usually easier to fix but failing late might avoid some cases where we don't need to read the JAR at all.

PR #1318, opened 2020-03-28
remove @VisibleForTesting method

elharo:
@suztomo

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathEntry.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathEntry.java
@@ -20,19 +20,15 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 
 import com.google.cloud.tools.opensource.dependencies.Artifacts;
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.reflect.ClassPath.ClassInfo;
-import java.io.File;
 import java.io.IOException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.Objects;
 import org.apache.bcel.classfile.JavaClass;
 import org.eclipse.aether.artifact.Artifact;
-import org.eclipse.aether.artifact.DefaultArtifact;
 
 /** An entry in a class path. */
 public final class ClassPathEntry {
@@ -95,12 +91,6 @@ public final class ClassPathEntry {
     } else {
       return jar.toString();
     }
-  }
-
-  @VisibleForTesting
-  public static ClassPathEntry of(String coordinates, String filePath) {
-    Artifact artifact = new DefaultArtifact(coordinates);
-    return new ClassPathEntry(artifact.setFile(new File(filePath)));
   }
 
   /**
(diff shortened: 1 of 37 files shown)


linkage-monitor/src/test/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitorTest.java
```diff
@@ -96,23 +115,17 @@ public void testBomSnapshot()
...
-    ClassPathEntry jarA = ClassPathEntry.of("foo:a:1.2.3", "foo/a-1.2.3.jar");
-    ClassPathEntry jarB = ClassPathEntry.of("foo:b:1.0.0", "foo/b-1.0.0.jar");
+    
+    Artifact artifactA = new DefaultArtifact("foo:a:1.2.3")
+        .setFile(new File("foo/a-1.2.3.jar"));
+    Artifact artifactB = new DefaultArtifact("foo:b:1.0.0")
+        .setFile(new File("foo/b-1.0.0.jar"));
+    
+    ClassPathEntry jarA = new ClassPathEntry(artifactA);
+    ClassPathEntry jarB = new ClassPathEntry(artifactB);
```
suztomo: Before: 2 lines with 2 variables
After: 6 lines with 4 variables

I see the before is better.

elharo: Arguably we don't have the right constructors yet. However the model code seems to be OK with this, and it's nice to keep the tests as close to the model code as reasonable. But my main goal is to remove a method that's only needed for testing. That's always a good thing. 

Could this become a test utility? Perhaps. Or we could use test fixtures to avoid duplicating code for object creation.

suztomo: I agree that the removal of the test-only method is a good thing if the lost benefit is minimal.

> Could this become a test utility? Perhaps.

Would you implement this `of()` as a test utility?

I haven't thought about setting up test fixtures for the test cases. If you think that's even better, I'm interested in what it looks like.

suztomo:
Would you implement the factory method as a test utility?

suztomo:
I think pushing back to model code or somewhere close to the model is better. Other test cases in this PR also benefit from the method. If it's removed, I'll soon need to create a utility method to create one.
https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1324/files#diff-bcf4f5a626b8d855bf134df7e002e437R331

How about moving `com.google.cloud.tools.opensource.classpath.TestHelper` in test to main?

elharo:
I prefer to avoid pushing test code into the src/java tree unless absolutely necessary, so I prefer not to move TestHelper. I'm not fundamentally opposed to a new constructor for this except that we don't seem to need it except for tests.

suztomo:
Given the observations below, I think we should keep `ClassPathEntry.of` at the class. The benefit of having the method (concise instance creation in tests) is bigger than the goodness of removing a test-only method.

- `ClassPathEntry.of` makes test code concise.
- `ClassPathEntry.of` is used by different modules (`dependencies`, `linkage-monitor` and `dashboard`). We cannot place a utility method in `dependencies`'s TestHelper class.
- We don't want to place test-specific code to main.
- `ClassPathEntry.of` is explicitly marked as `@VisibleForTesting`.

suztomo:
I now find it's not good use of `@VisibleForTesting`. That method is actually a test utility in main directory. The visibility is public because of it, not to test the `of(...)` method.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathEntry.java
```diff
@@ -95,12 +92,6 @@ public String toString() {
     }
   }
 
-  @VisibleForTesting
-  public static ClassPathEntry of(String coordinates, String filePath) {
```
suztomo, line 99: This is a normal factory method. Nothing special for tests.

suztomo:
Created a PR https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1331 for my idea.

PR #1322, opened 2020-03-30
Add test for listFileNames

elharo:
@suztomo This helps us explore alternate implementations of this functionality

### DIFF
--- a/dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassPathEntryTest.java
+++ b/dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassPathEntryTest.java
@@ -24,6 +24,7 @@ import com.google.common.collect.ImmutableSet;
 import com.google.common.testing.EqualsTester;
 import com.google.common.truth.Truth;
 import java.io.IOException;
+import java.net.URISyntaxException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import org.eclipse.aether.RepositorySystem;
@@ -88,7 +89,7 @@ public class ClassPathEntryTest {
 
   @Test
   public void testListFileNames() throws IOException, ArtifactResolutionException {
-    
+ 
     // copy into the local repository so we can read the jar file
     Artifact artifact = resolveArtifact("com.google.truth.extensions:truth-java8-extension:1.0.1");
     
@@ -104,6 +105,24 @@ public class ClassPathEntryTest {
         "com.google.common.truth.PathSubject",
         "com.google.common.truth.Truth8",
         "com.google.common.truth.StreamSubject");
+  }
+  
+  @Test
+  public void testListFileNames_innerClasses()
+      throws IOException, ArtifactResolutionException, URISyntaxException {
+
+    ClassPathEntry entry = TestHelper.classPathEntryOfResource(
+        "testdata/conscrypt-openjdk-uber-1.4.2.jar");
+    
+    ImmutableSet<String> classFileNames = entry.listClassFileNames();
+    Truth.assertThat(classFileNames).containsAtLeast(
+        "org.conscrypt.OpenSSLSignature$1",
+        "org.conscrypt.OpenSSLContextImpl$TLSv1",
+        "org.conscrypt.TrustManagerImpl$1",
+        "org.conscrypt.PeerInfoProvider",
+        "org.conscrypt.PeerInfoProvider$1",
+        "org.conscrypt.ExternalSession$Provider",
+        "org.conscrypt.OpenSSLMac");
   }
 
   private static Artifact resolveArtifact(String coords) throws ArtifactResolutionException {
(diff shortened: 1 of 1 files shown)


dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassPathEntryTest.java
```diff
@@ -78,6 +86,38 @@ public void testToStringArtifact() {
     assertEquals("com.google:foo:0.0.1", entry.toString());
   }
 
+  @Test
+  public void testListFileNames() throws IOException, ArtifactResolutionException {
+    
+    // copy into the local repository so we can read the jar file
+    Artifact artifact = resolveArtifact("com.google.truth.extensions:truth-java8-extension:1.0.1");
```
suztomo:
```suggestion
    Artifact artifact = LinkageCheckerTest.resolvePaths("com.google.truth.extensions:truth-java8-extension:1.0.1").get(0).getArtifact();
```

This does almost same thing, although `resolveArtifact` does not fetch (unnecessary) dependencies.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassPathEntryTest.java
```diff
@@ -78,6 +86,38 @@ public void testToStringArtifact() {
...
+    Truth.assertThat(classFileNames).containsExactly(
+        "com.google.common.truth.IntStreamSubject",
+        "com.google.common.truth.LongStreamSubject",
+        "com.google.common.truth.OptionalDoubleSubject",
+        "com.google.common.truth.OptionalSubject",
+        "com.google.common.truth.OptionalIntSubject",
+        "com.google.common.truth.OptionalLongSubject",
+        "com.google.common.truth.PathSubject",
+        "com.google.common.truth.Truth8",
+        "com.google.common.truth.StreamSubject");
```
suztomo: You may want to add some inner classes.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassPathEntryTest.java
suztomo: Or, you can just use one of the JAR files in test/resources/testdata. This idea does not require (unnecessary) network connections.

elharo: This test definitely shouldn't pull an entire dependency tree to only look at one jar. 

We could bundle it. I might yet do that if we need to contrive a specific jar with specific characteristic for testing. For now this one was small and simple enough that it could easily be characterized.

suztomo: Truth-java-8-extension is small JAR file and thus a good case to observe the completeness of listing. 

Apart from that, I found conscrypt-openjdk-uber-1.4.2.jar in `test/resources/testdata` has nice list of contents to assert.

```
suztomo-macbookpro44% jar tf dependencies/src/test/resources/testdata/conscrypt-openjdk-uber-1.4.2.jar
META-INF/MANIFEST.MF
META-INF/SIGNINGC.SF
META-INF/SIGNINGC.RSA
META-INF/
META-INF/native/
META-INF/native/libconscrypt_openjdk_jni-linux-x86_64.so
META-INF/native/conscrypt_openjdk_jni-windows-x86.dll
META-INF/native/libconscrypt_openjdk_jni-osx-x86_64.dylib
META-INF/native/conscrypt_openjdk_jni-windows-x86_64.dll
org/
org/conscrypt/
org/conscrypt/NativeCrypto.class
org/conscrypt/OpenSSLSignature$1.class
org/conscrypt/TrustManagerImpl$1.class
org/conscrypt/OpenSSLContextImpl$TLSv1.class
org/conscrypt/OpenSSLX509CertPath.class
org/conscrypt/OpenSSLECPrivateKey.class
org/conscrypt/OpenSSLCipherRSA$OAEP$SHA512.class
```

elharo: That does have some nice inner classes and is not too large.

As to network connections in tests, yes, I am nervous about that. Though I do wonder, is this any different than pulling in our own dependencies from Maven central. In fact, this is a Maven dependency we're loading from the same serverlocal cache we'd load it from if we depended on it directly from pom.xml.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ClassPathEntryTest.java
elharo: yes, looking at that now.

suztomo: Also verify there's no such class that has package name `META-INF`?

PR #1324, opened 2020-03-31
ArtifactMatcher

### LINKED ISSUE #1271: Exclusion file: support Artifact matcher
Exclusion file: support Artifact matcher.

Annotated JAR (https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1179) is needed before starting implementing this.

suztomo:
Fixes #1271


suztomo:
As discussed, we don't use it for now. Even appengine-api-sdk, we'd like to find new errors that are not previously there. Therefore not suppressing artifact-level.

When we need this, I'll reopen this.

PR #1331, opened 2020-04-03
Removing ClassPathEntry.of in favor of new constructor

suztomo:
The previous usage of `@VisibleForTesting` was wrong. This PR treats ClassPathEntry instance creation with coordinates and JAR location as a normal constructor. Nothing special for tests.


suztomo:
This is not YAGNI. We do use the constructors. This is the same manner as DefaultArtifact having many constructors.

PR #1335, opened 2020-04-06
avoid polluting local repository

### LINKED ISSUE #1301: DependencyGraphBuilderTest.testConfigureAdditionalMavenRepositories_notToUseMavenCentral is flaky
Normal usage installs the test dependency into the local repo which causes the test to fail. Perhaps use a different, temporary local repo for the test?

[ERROR]   DependencyGraphBuilderTest.testConfigureAdditionalMavenRepositories_notToUseMavenCentral:224 expected to contain: com.google.guava:guava:15.0-rc1
testing whether    : actual element has artifact expected element
but was            : []

elharo:
fixes #1301

This injects a local repository into the graph builder from the test, so we avoid installing any artifacts into `~.m2`

Suggestions for making this less hacky appreciated.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
@@ -122,9 +122,8 @@ public final class RepositoryUtility {
    * customary ~/.m2 directory. If not found, it creates an initially empty repository in
    * a temporary location.
    */
-  public static RepositorySystemSession newSession(RepositorySystem system) {
+  public static DefaultRepositorySystemSession newSession(RepositorySystem system) {
     DefaultRepositorySystemSession session = createDefaultRepositorySystemSession(system);
-    session.setReadOnly();
     return session;
   }
 
@@ -133,7 +132,7 @@ public final class RepositoryUtility {
    *
    * @see {@link DependencyGraphBuilder}
    */
-  static RepositorySystemSession newSessionForFullDependency(RepositorySystem system) {
+  static DefaultRepositorySystemSession newSessionForFullDependency(RepositorySystem system) {
     DefaultRepositorySystemSession session = createDefaultRepositorySystemSession(system);
 
     // This combination of DependencySelector comes from the default specified in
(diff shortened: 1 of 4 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -122,9 +122,8 @@ static DefaultRepositorySystemSession createDefaultRepositorySystemSession(
    * customary ~/.m2 directory. If not found, it creates an initially empty repository in
    * a temporary location.
    */
-  public static RepositorySystemSession newSession(RepositorySystem system) {
+  public static DefaultRepositorySystemSession newSession(RepositorySystem system) {
     DefaultRepositorySystemSession session = createDefaultRepositorySystemSession(system);
-    session.setReadOnly();
```
suztomo, line 127: I think setting we want to set the maven directory in createDefaultRepositorySystemSession to keep the returned session read-only.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilderIntegrationTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class DependencyGraphBuilderIntegrationTest {
+
+  private Correspondence<UnresolvableArtifactProblem, String> problemOnArtifact =
+      Correspondence.transforming(
+          (UnresolvableArtifactProblem problem) -> Artifacts.toCoordinates(problem.getArtifact()),
+          "has artifact");
+
+  @Test
+  public void testConfigureAdditionalMavenRepositories_notToUseMavenCentral() {
```
suztomo, line 36: Now this test passes. Thank you.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
elharo: I'm not sure that's necessary. Is there some reason it has to be read-only?

I did try setting it here, but the extra code to configure it from the test was even uglier since DependencyGraphBuilderIntegrationTest doesn't directly access RepositoryUtility.

suztomo: OK. Then this current code is good.

PR #1362, opened 2020-04-20
Linkage Checker 1.3.0 release

suztomo:
Minor version bump as It has a new functionality to output linkage errors as a file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 1.3.0
 * LinkageCheckerMain has an option (`-o`) to output linkage errors into a file ([document](
   https://github.com/GoogleCloudPlatform/cloud-opensource-java/wiki/LinkageCheckerMain#exclusion-files
-  )). This feature is currently alpha; we may change the behavior/format in later release.
+  )). This feature is currently alpha; we may change the behavior/format in later releases.
 * LinkageCheckerMain throws LinkageCheckResultException if it finds linkage errors.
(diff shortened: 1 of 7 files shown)


suztomo:
No, this doesn't. Maven 3.6.0 is required only for running the integration tests with maven-invoker-plugin.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Linkage Checker Enforcer Rule Change Log
 
+## 1.3.0
+* LinkageCheckerMain has an option (`-o`) to output linkage errors into a file.
```
elharo, line 4: How stable is this? I.e. how committed are we to the exact format and feature set? I'm wondering if this should be marketed as an alpha release. 

I was going to suggest we should add doc improvements in this PR, but maybe it would be better to have a separate PR ready to go with the doc changes before we release it. Writing the docs up front will also give us viewpoint on whether we have the design right for this feature.

suztomo: I will create a new Wiki page for LinkageCheckerMain class indicating this is alpha and will update this changelog in this PR.

suztomo:
@elharo PTAL. Added a new document https://github.com/GoogleCloudPlatform/cloud-opensource-java/wiki/LinkageCheckerMain and linked it from the changelog, mentioning it's alpha.

suztomo:
Added " we may change the behavior/format in later releases."

PR #1366, opened 2020-04-23
gRPC 1.29 in GCP Libraries BOM

suztomo:
The release note (https://github.com/grpc/grpc-java/releases/tag/v1.29.0) does not have any API change.


boms/cloud-oss-bom/pom.xml
```diff
@@ -46,7 +46,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <guava.version>29.0-android</guava.version>
     <google.cloud.java.version>0.124.0</google.cloud.java.version>
```
elharo, line 48: this needs to be updated too.

suztomo: Updated.

PR #1380, opened 2020-04-29
The Enforcer Rule to Print Dependency Paths

### LINKED ISSUE #1369: Show dependency path to a problematic artifact in Linkage Checker Enforcer Rule
Towards #1368 , which suggests 2 paths. This issue is to show one path first.

suztomo:
Fixes #1369.

### DIFF
--- a/enforcer-rules/src/it/bom-project-error/verify.groovy
+++ b/enforcer-rules/src/it/bom-project-error/verify.groovy
@@ -1,4 +1,4 @@
-def buildLog = new File(basedir, "build.log").text.replaceAll("\\r\\n", "\\n")
+def buildLog = new File(basedir, "build.log").text.replaceAll("\\r\\n", "\n")
 
 assert buildLog.contains('''\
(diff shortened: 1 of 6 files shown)


enforcer-rules/src/it/bom-project-error/verify.groovy
```diff
@@ -1,8 +1,16 @@
-def buildLog = new File(basedir, "build.log")
+def buildLog = new File(basedir, "build.log").text.replaceAll("\\r", "")
```
elharo, line 1: why do we need to do this?
what if it's only \r we're removing and there aren't any \n?
at the least this should convert \r\n to \n rather than stripping \r

suztomo: Updated to replace "\r\n".

suztomo: Now Windows build fails. Investigating.

suztomo:
@elharo PTAL.

PR #1389, opened 2020-05-06
remove dependency on multimap

elharo:
@suztomo I noticed it was pretty easy to remove this multimap from our public API.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
@@ -94,9 +94,8 @@ public final class ClassPathResult {
   }
   
   /**
-   * Returns the classpath entries for the specified coordinates.
-   * 
-   * TODO: Explain why this is plural. Shouldn't there be exactly one?
+   * Returns the classpath entries whose transitive dependencies contain
+   * the artifact with the specified coordinates.
    */
   public ImmutableSet<ClassPathEntry> getClassPathEntries(String coordinates) {
     ImmutableSet.Builder<ClassPathEntry> builder = ImmutableSet.builder();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
@@ -24,8 +24,6 @@ import com.google.cloud.tools.opensource.dependencies.UnresolvableArtifactProble
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.ImmutableSetMultimap;
-import com.google.common.collect.ImmutableSetMultimap.Builder;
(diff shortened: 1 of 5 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
```diff
@@ -93,21 +105,14 @@ public String formatDependencyPaths(Iterable<ClassPathEntry> entries) {
...
   }
-
+  
   /**
-   * Returns mapping from the Maven coordinates to {@link ClassPathEntry}. The keys are the
-   * coordinates of the direct dependencies of the root nodes in {@link #dependencyPaths}. The
-   * values are all {@link ClassPathEntry}s in the subtree of the key.
+   * Returns the classpath entries for the specified coordinates.
+   * 
+   * TODO: Explain why this is plural. Shouldn't there be exactly one?
```
suztomo, line 112: As in the previous javadoc, the values are all {@link ClassPathEntry}s in the subtree of the key.

elharo: Probably not though we can revisit. The focus is on not passing a bunch of weird data structures around. Instead I'd like public classes that represent our entities (ClassFile, ClassPathEntry, ClassPath, Artifact, Dependency, etc.) and operations on those.

suztomo: That's good. Would you revert the javadoc update? The previous javadoc explains why the values are plural.

elharo: I'm not sure it did. Or rather what it said didn't match the code. Notice that this returns class path entries, **not** dependency paths.

suztomo: How about "Returns the class path entries in the subtree of the specified coordinates"?

elharo: done

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
```diff
@@ -93,21 +92,24 @@ public String formatDependencyPaths(Iterable<ClassPathEntry> entries) {
...
     return message.toString();
   }
-
+  
   /**
-   * Returns mapping from the Maven coordinates to {@link ClassPathEntry}. The keys are the
-   * coordinates of the direct dependencies of the root nodes in {@link #dependencyPaths}. The
-   * values are all {@link ClassPathEntry}s in the subtree of the key.
+   * Returns the classpath entries whose transitive dependencies contain
+   * the artifact with the specified coordinates.
```
suztomo, line 98: It's the other way around. The artifact with the specified coordinates contains the class path entries as (transitive and direct) dependencies.

elharo: fixed

PR #1390, opened 2020-05-07
make methods static

### LINKED ISSUE #1391: Multiple executions of enforcer rule do not reset artifact problems
so one can leak into another

elharo:
@suztomo I wanted your opinion on this refactoring. If this works, then it would prove we could make the ClassPathResult constructor private (or visible for testing) by moving the build methods into ClassPathResult itself.

It does change when we log unavailable artifacts, but I think that's OK.

I do need to check whether execute can be called twice in normal evaluation. Hmm, looks like it can be so this also fixes #1391.

### DIFF
--- a/enforcer-rules/src/main/java/com/google/cloud/tools/dependencies/enforcer/LinkageCheckerRule.java
+++ b/enforcer-rules/src/main/java/com/google/cloud/tools/dependencies/enforcer/LinkageCheckerRule.java
@@ -83,7 +83,7 @@ public class LinkageCheckerRule extends AbstractNonCacheableEnforcerRule {
   /**
    * Maven packaging values known to be irrelevant to Linkage Check for non-BOM project.
    *
-   * @see <a href="https://maven.apache.org/ref/3.6.1/maven-core/artifact-handlers.html">Maven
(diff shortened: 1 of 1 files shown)


enforcer-rules/src/main/java/com/google/cloud/tools/dependencies/enforcer/LinkageCheckerRule.java
```diff
@@ -313,8 +303,9 @@ private ClassPathResult buildClasspathFromException(
       if (cause instanceof ArtifactTransferException) {
         ArtifactTransferException artifactException = (ArtifactTransferException) cause;
         Artifact artifact = artifactException.getArtifact();
-        artifactProblems.add(
-            DependencyGraphBuilder.createUnresolvableArtifactProblem(dependencyGraph, artifact));
```
suztomo, line 317: I thought that the artifactProblems should be be passed to the second argument of ClassPathResult. But the field seems unused in this LinkageCheckerRule. Filed https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1392.

PR #1394, opened 2020-05-07
don't load classes when we only need the names

### LINKED ISSUE #1320: avoid Guava ClassPath.from in ClassPathEntry
Here's what we do:

```
    // Leveraging Google Guava reflection as BCEL doesn't have API to list classes in a jar file
    com.google.common.reflect.ClassPath classPath =
```

Could we read the jar file directly instead without loading the classes? It's just a zip file. That might miss inner classes though.

elharo:
@suztomo it might save us a little time to not try to load classes when we only need the names. This also avoids some problems with malformed .class files. If we want to detect such malformed files this is a bad approach. If we want to allow such malformed files this is a good approach. 

fixes #1320


elharo:
Thanks. I thought we had an issue for this somewhere but I couldn't put my finger on it. 

We do still need to solve that OutOfMemoryError. I don't expect this to do it. but there are some other things I've yet to try.

PR #1395, opened 2020-05-09
introduce dependency mediation class

elharo:
@suztomo I think this makes the code for removing excess versions from a classpath somewhat more straightforward. Let me know what you think.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilder.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilder.java
@@ -21,8 +21,6 @@ import com.google.cloud.tools.opensource.dependencies.DependencyGraphResult;
 import com.google.cloud.tools.opensource.dependencies.DependencyPath;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.LinkedListMultimap;
-import java.io.File;
-import java.nio.file.Path;
 import java.util.List;
 import org.eclipse.aether.artifact.Artifact;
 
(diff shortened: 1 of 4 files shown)


suztomo:
As you commented in TODO there might be even better solution to design the dependency mediation logic. Yet, this PR clarifies the current code nicely.

Would you consider renaming the class to be more specific? Also add a test for the DependencyMediation class?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathBuilder.java
```diff
@@ -70,39 +65,19 @@ public ClassPathResult resolve(List<Artifact> artifacts) {
...
+    DependencyMediation mediation = new DependencyMediation();
 
     for (DependencyPath dependencyPath : dependencyPaths) {
       Artifact artifact = dependencyPath.getLeaf();
-      File file = artifact.getFile();
-      if (file == null) {
-        // When artifact was not downloaded, it's recorded in result.getArtifactProblems().
-        continue;
+      mediation.put(dependencyPath);
+      if (mediation.selects(artifact)) {
```
suztomo, line 76: It's a bit odd that DependencyMediation class does not have logic to traverse a tree. Because of that, its `put` and `selects` interface cannot implement Gradle's highest-win dependency mediation.

(I don't expect this PR to address that concern; just writing down what I felt)

elharo: renamed "MavenDependencyMediation"

PR #1403, opened 2020-05-13
Linkage Checker to use class file names when checking existence of a class in ClassPathEntry

### LINKED ISSUE #1401: Linkage Checker rule reporting not accessible field for project of packaging `war`
I'm troubled with Linkage Checker rule reporting not accessible field for project of packaging `war`.
[Example project](https://github.com/pzrep/LinkageChecker-and-servlet/) (with rule enabled in [enable_LC_rule](https://github.com/pzrep/LinkageChecker-and-servlet/tree/enable_LC_rule) branch) has two almost identical modules, with the only difference in packaging (and artifactId):
```
$ diff --recursive webapp asjar
diff --recursive webapp/pom.xml asjar/pom.xml
9,10c9,10
<   <artifactId>webapp</artifactId>
<   <packaging>war</packaging>
---
>   <artifactId>asjar</artifactId>
>   <packaging>jar</packaging>
```
Having the same enforcer configuration the `asjar` module verifies successfully, while the `webapp` reports following:
```
[INFO] --- maven-enforcer-plugin:3.0.0-M3:enforce (enforce-on-verify) @ webapp ---
[ERROR] Linkage Checker rule found 1 error. Linkage error report:
(com.github.pzrep.LinkageChecker-and-servlet:webapp:1-SNAPSHOT) pzrep.ReportServlet's field inv is not accessible;
  referenced by 1 class file
    pzrep.ReportServlet (com.github.pzrep.LinkageChecker-and-servlet:webapp:1-SNAPSHOT)

[ERROR] Problematic artifacts in the dependency tree:
com.github.pzrep.LinkageChecker-and-servlet:webapp:1-SNAPSHOT is at:
  com.github.pzrep.LinkageChecker-and-servlet:webapp:war:1-SNAPSHOT

[WARNING] Rule 0: com.google.cloud.tools.dependencies.enforcer.LinkageCheckerRule failed with message:
Failed while checking class path. See above error report.
```
Interestingly - when fields are [made public](https://github.com/pzrep/LinkageChecker-and-servlet/tree/public), LC no longer complains.

suztomo:
Fixes #1401.

This snapshot version successfully builds the project in #1401:
https://gist.github.com/suztomo/1979c34ec9e35cb0291fa2ebb96bf58b

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -321,6 +321,7 @@ class ClassDumper {
     // However, it required the superclass of a target class to be loadable too; otherwise
     // ClassNotFoundException was raised. It was inconvenient because we only wanted to know the
     // location of the target class, and sometimes the superclass is unavailable.
+
     String filename = classRepository.getFileName(className);
     return Iterables.getFirst(fileNameToClassPathEntry.get(filename), null);
   }
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -327,9 +327,8 @@ class ClassDumper {
   }
 
   /**
-   * Converts a binary name to a class file name. Usually they are the same, however it may be
-   * different if the JAR file has framework-specific prefix such as "WEB-INF.classes." in class
-   * file names.
+   * Converts a binary name to the file name of a class. Read {@link FixedSizeClassPathRepository}
+   * for the difference.
    */
   String getFileName(String classBinaryName) {
     return classRepository.getFileName(classBinaryName);
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -69,7 +69,7 @@ class ClassDumper {
   private final ImmutableList<ClassPathEntry> inputClassPath;
   private final FixedSizeClassPathRepository classRepository;
   private final ClassLoader extensionClassLoader;
-  private final ImmutableListMultimap<String, ClassPathEntry> classFileNameToClassPathEntry;
+  private final ImmutableListMultimap<String, ClassPathEntry> fileNameToClassPathEntry;
 
   private static FixedSizeClassPathRepository createClassRepository(List<ClassPathEntry> entries) {
     ClassPath classPath = new LinkageCheckClassPath(entries);
@@ -90,7 +90,7 @@ class ClassDumper {
     
     ImmutableListMultimap.Builder<String, ClassPathEntry> builder = ImmutableListMultimap.builder();
     for (ClassPathEntry entry : entries) {
-      for (String className : entry.getClassFileNames()) {
+      for (String className : entry.getFileNames()) {
         builder.put(className, entry);
       }
     }
@@ -108,7 +108,7 @@ class ClassDumper {
     this.inputClassPath = ImmutableList.copyOf(inputClassPath);
     this.classRepository = createClassRepository(inputClassPath);
     this.extensionClassLoader = extensionClassLoader;
-    this.classFileNameToClassPathEntry = map;
+    this.fileNameToClassPathEntry = map;
   }
 
   /**
@@ -320,7 +320,7 @@ class ClassDumper {
     // However, it required the superclass of a target class to be loadable too; otherwise
     // ClassNotFoundException was raised. It was inconvenient because we only wanted to know the
     // location of the target class, and sometimes the superclass is unavailable.
-    ClassPathEntry entry = Iterables.getFirst(classFileNameToClassPathEntry.get(className), null);
+    ClassPathEntry entry = Iterables.getFirst(fileNameToClassPathEntry.get(className), null);
     if (entry != null) {
       return entry;
     }
@@ -331,7 +331,7 @@ class ClassDumper {
     if (specialLocation == null) {
       return null;
     }
-    return Iterables.getFirst(classFileNameToClassPathEntry.get(specialLocation), null);
+    return Iterables.getFirst(fileNameToClassPathEntry.get(specialLocation), null);
   }
 
   /**
@@ -356,7 +356,7 @@ class ClassDumper {
 
     ImmutableList.Builder<String> corruptedClassFileNames = ImmutableList.builder();
 
-    for (String classFileName : entry.getClassFileNames()) {
+    for (String classFileName : entry.getFileNames()) {
       if (classFileName.startsWith("META-INF.versions.")) {
         // Linkage Checker does not support multi-release JAR (for Java 9+) yet
         // https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/897
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -90,7 +90,7 @@ class ClassDumper {
     
     ImmutableListMultimap.Builder<String, ClassPathEntry> builder = ImmutableListMultimap.builder();
     for (ClassPathEntry entry : entries) {
-      for (String className : entry.getClassNames()) {
+      for (String className : entry.getClassFileNames()) {
         builder.put(className, entry);
       }
     }
@@ -335,6 +335,19 @@ class ClassDumper {
   }
 
   /**
+   * Returns class file name which is a location of a class file within a JAR file. Usually it is
+   * the same as {@code classBinaryName}, however it may be different if the JAR file has framework-
+   * specific prefix such as "WEB-INF.classes".
+   */
+  String getClassFileName(String classBinaryName) {
+    String specialLocation = classRepository.getSpecialLocation(classBinaryName);
+    if (specialLocation == null) {
+      return classBinaryName;
+    }
+    return specialLocation;
+  }
+
+  /**
    * Returns a set of {@link JavaClass}es which have entries in the {@code entry} through {@link
    * #classRepository}.
    */
@@ -343,7 +356,7 @@ class ClassDumper {
 
     ImmutableList.Builder<String> corruptedClassFileNames = ImmutableList.builder();
 
-    for (String classFileName : entry.getClassNames()) {
+    for (String classFileName : entry.getClassFileNames()) {
       if (classFileName.startsWith("META-INF.versions.")) {
         // Linkage Checker does not support multi-release JAR (for Java 9+) yet
         // https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/897
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -335,9 +335,9 @@ class ClassDumper {
   }
 
   /**
-   * Returns class file name which is a location of a class file within a JAR file. Usually it is
-   * the same as {@code classBinaryName}, however it may be different if the JAR file has framework-
-   * specific prefix such as "WEB-INF.classes".
+   * Converts a binary name to a class file name. Usually they are the same, however it may be
+   * different if the JAR file has framework-specific prefix such as "WEB-INF.classes." in class
+   * file names.
    */
   String getClassFileName(String classBinaryName) {
     String specialLocation = classRepository.getSpecialLocation(classBinaryName);
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -103,12 +103,12 @@ class ClassDumper {
   private ClassDumper(
       List<ClassPathEntry> inputClassPath,
       ClassLoader extensionClassLoader,
-      ImmutableListMultimap<String, ClassPathEntry> map)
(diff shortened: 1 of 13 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -202,9 +209,12 @@ private LinkageChecker(
         classToSymbols.getClassToFieldSymbols();
     for (ClassFile classFile : classToFieldSymbols.keySet()) {
       ImmutableSet<FieldSymbol> fieldSymbols = classToFieldSymbols.get(classFile);
+      ImmutableSet<String> classFileNamesInSource =
+          classFile.getClassPathEntry().getClassFileNames();
       for (FieldSymbol fieldSymbol : fieldSymbols) {
-        if (!classFile.getClassPathEntry().getClassNames()
-            .contains(fieldSymbol.getClassBinaryName())) {
+        String classBinaryName = fieldSymbol.getClassBinaryName();
```
suztomo, line 215: This is the root cause of the problem. LinkageChecker was checking the existence of a class by class *binary* name (e.g., `pzrep.ReportServlet`), while the `ClassPathEntry.getClassNames()` returns a set of class *file* names (e.g., `WEB-INF.classes.pzrep.ReportServlet`).

enforcer-rules/src/it/war-project-private-modifier/src/main/java/com.google.example/Dummy.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.example;
+
+class Dummy {
+  private int privateField = 0;
+
+  private int privateMethod() {
```
suztomo, line 22: This integration test case was failing with the latest enforcer rule 1.4.0:
https://gist.github.com/suztomo/859543af569e53c7ba06fde6ec1f68ba

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -334,6 +334,19 @@ ClassPathEntry findClassLocation(String className) {
     return Iterables.getFirst(classFileNameToClassPathEntry.get(specialLocation), null);
   }
 
+  /**
+   * Returns class file name which is a location of a class file within a JAR file. Usually it is
```
elharo, line 338: "Returns class file name" --> Converts a binary name to a class file name.

It is **not** a location. It is a name.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -334,6 +334,19 @@ ClassPathEntry findClassLocation(String className) {
...
+   * the same as {@code classBinaryName}, however it may be different if the JAR file has framework-
```
elharo, line 339: a framework-specific

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -334,6 +334,19 @@ ClassPathEntry findClassLocation(String className) {
...
+   * specific prefix such as "WEB-INF.classes".
+   */
+  String getClassFileName(String classBinaryName) {
```
elharo, line 342: This logic belongs in ClassPathEntry and ClassFile, not here.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathEntry.java
```diff
@@ -112,13 +113,18 @@ private void readClassFileNames() throws IOException {
...
-  
+
   /**
-   * Returns the fully qualified names of the classes in this entry's jar file.
-   * 
+   * Returns a list of class file names in {@link #jar} as in {@link JavaClass#getFileName()}. This
+   * class file name is usually a fully qualified class name. However a class file name may have a
+   * framework-specific prefix. Example: {@code BOOT-INF.classes.com.google.Foo}.
+   *
+   * <p>This function caches the result. Therefore the caller does not need to worry about the
```
elharo, line 122: I'm hesitant to expose implementation details in API doc

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -163,9 +166,10 @@ private LinkageChecker(
             }
           }
         }
-        if (!classFile.getClassPathEntry().getClassNames()
-            .contains(classSymbol.getClassBinaryName())) {
 
+        String classBinaryName = classSymbol.getClassBinaryName();
```
elharo, line 170: Avoid moving data between classes. Instead ClassSymbol should be able to return the classFileName if that's what we need.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -189,9 +193,12 @@ private LinkageChecker(
         classToSymbols.getClassToMethodSymbols();
     for (ClassFile classFile : classToMethodSymbols.keySet()) {
       ImmutableSet<MethodSymbol> methodSymbols = classToMethodSymbols.get(classFile);
+      ImmutableSet<String> classFileNamesInSource =
+          classFile.getClassPathEntry().getClassFileNames();
       for (MethodSymbol methodSymbol : methodSymbols) {
-        if (!classFile.getClassPathEntry().getClassNames()
-            .contains(methodSymbol.getClassBinaryName())) {
+        String classBinaryName = methodSymbol.getClassBinaryName();
```
elharo, line 199: again, just `methodSymbol.getClassFileName()`

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: ClassPathEntry cannot convert class binary names to class file names, because it does not read the content of .class files. On the other hand, ClassDumper (through classRepository) already reads the content of class files and thus can convert class binary names to class file names.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathEntry.java
suztomo: Removed the mention about the caching.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
suztomo: A ClassSymbol cannot know the class file name. A symbol is an abstract entity apart from concrete class files. For example, when a class file's constant pool references a class symbol of "com.google.Foo", it does not know the class symbol is materialized by a class file "BOOT-INF/classes/com/google/Foo.class", another class file "WEB-INF/classes/com/google/Foo.class", or just "com/google/Foo.class".

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathEntry.java
```diff
@@ -112,13 +114,15 @@ private void readClassFileNames() throws IOException {
       this.classFileNames = classNames.build();
     }
   }
-  
+
   /**
-   * Returns the fully qualified names of the classes in this entry's jar file.
-   *
```
suztomo, line 118: The previous javadoc was incorrect. It is not fully qualified class names. It returns class file names in a JAR.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
elharo: How is BCEL getting these names? Is it reading the individual .class files? Or does it know something we don't know about the format of JAR files?

suztomo: BCEL's JavaClass reads class name from class file's constant pool: https://github.com/apache/commons-bcel/blob/master/src/main/java/org/apache/bcel/classfile/JavaClass.java#L157

elharo: It feels right that readClassFileNames be the place this happens, if necessary by reading the byte code, though perhaps there's a simpler way to do this. For instance we could strip off known prefixes.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -163,9 +165,10 @@ private LinkageChecker(
             }
           }
         }
-        if (!classFile.getClassPathEntry().getFileNames()
-            .contains(classSymbol.getClassBinaryName())) {
 
+        String classBinaryName = classSymbol.getClassBinaryName();
+        String classFileName = classDumper.getFileName(classBinaryName);
```
suztomo, line 170: I was thinking to rename this variable to `fileName` or `fileNameOfClass`. But `classFileName` fits best.

suztomo:
@elharo PTAL. I merged your change.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
suztomo: As the classRepository maintains the mappings, we do not need to read the class files again in ClassPathEntry.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathEntry.java
```diff
@@ -92,11 +93,11 @@ public String toString() {
   }
 
   /**
-   * Returns a list of class file names in {@link #jar} as in {@link JavaClass#getFileName()}. This
-   * class file name is usually a fully qualified class name. However a class file name may have a
-   * framework-specific prefix. Example: {@code BOOT-INF.classes.com.google.Foo}.
+   * Populates {@link #fileNames} through the classes in {@link #jar}. These file names are usually
+   * a fully qualified class name. However a class file name may have a framework-specific prefix.
```
elharo, line 97: a fully qualified class name --> fully qualified class names

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -150,6 +150,8 @@ private LinkageChecker(
     
     for (ClassFile classFile : classToClassSymbols.keySet()) {
       ImmutableSet<ClassSymbol> classSymbols = classToClassSymbols.get(classFile);
```
elharo, line 152: Not new in this PR, but I notice this is another case where the OO API is turned inside out. This should be `classFile.getClassSymbols`.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -150,6 +150,8 @@ private LinkageChecker(
...
+      ImmutableSet<String> classFileNamesInSource = classFile.getClassPathEntry().getFileNames();
```
elharo, line 153: why "InSource"? Would just classFileNames work?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
suztomo: The mismatch from OO API might came from my mental design of Linkage Checker:

- Get class path (JAR files) from input artifact
- Get symbol references in class files (such as this `classToClassSymbols`) from the class path
- Get linkage errors by validating the symbol references

I was feeling that this fits declarative/functional-programming style, not object-oriented:

```
linkageCheck = 
  artifact <- input
  classPath <- resolve(artifact)
  symbolReferences <- read(classPath)
  linkageErrors <- validate(symbolReferences)
  return linkageErrors
```

With that, I think I wanted to keep the clear separation of the value classes (e.g., `ClassFile`) from implementing IO operations. LinkageChecker and ClassDumper are responsible for the IO operations and logic to generate the value classes.

(This is not saying I'm against the object-oriented API, but sharing my finding.)

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
suztomo: classFileNames works. Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -326,6 +326,14 @@ ClassPathEntry findClassLocation(String className) {
     return Iterables.getFirst(fileNameToClassPathEntry.get(filename), null);
   }
 
+  /**
+   * Converts a binary name to the file name of a class. Read {@link FixedSizeClassPathRepository}
+   * for the difference.
+   */
+  String getFileName(String classBinaryName) {
```
elharo, line 333: just "className" might be enough here.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
elharo: Writing functional code in a non-functional language such as Java is a recipe for confusion. Case in point: I am constantly confused while reading our code. I usually have to review several classes to remind myself what the code does and how in each PR.

Separating I/O makes sense, but we shouldn't be storing the results of those I/O operations outside of the relevant classes.

PR #1404, opened 2020-05-14
clarify distinction between file names and class names

elharo:
@suztomo class path --> class path entry

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FixedSizeClassPathRepository.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FixedSizeClassPathRepository.java
@@ -52,14 +52,15 @@ final class FixedSizeClassPathRepository extends ClassPathRepository {
    *
    * <p>While class name and class file name are the same in most cases, sometimes classes are not
    * placed in the root of a JAR file to support a framework-specific JAR structure. For example,
-   * Spring Boot Gradle Java plugin places class files under "BOOT-INF/classes". To load such
+   * the Spring Boot Gradle Java plugin places class files under "BOOT-INF/classes". To load such
    * classes by class name, this mapping keeps track of the special location once they are loaded.
    *
    * <ul>
    *   <li>Key: class name (value from {@link JavaClass#getClassName()}) which has a special class
    *       file name different from its class name. Example: {@code com.google.Foo}
    *   <li>Value: the special class file name as in {@link JavaClass#getFileName()}, a path that
-   *       locates a class file in a class path. Example: {@code BOOT-INF.classes.com.google.Foo}
+   *       locates a class file in a class path entry.
+   *       Example: {@code BOOT-INF.classes.com.google.Foo}
    * </ul>
    *
    * @see <a
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FixedSizeClassPathRepository.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FixedSizeClassPathRepository.java
@@ -48,19 +48,23 @@ final class FixedSizeClassPathRepository extends ClassPathRepository {
   private final Cache<String, JavaClass> loadedClass;
 
   /**
-   * Mapping from class names to special class file locations.
+   * Mapping from class names to file names.
    *
-   * <p>While class name and class file name are the same in most cases, sometimes classes are not
-   * placed in the root of a JAR file to support a framework-specific JAR structure. For example,
-   * the Spring Boot Gradle Java plugin places class files under "BOOT-INF/classes". To load such
-   * classes by class name, this mapping keeps track of the special location once they are loaded.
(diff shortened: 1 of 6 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FixedSizeClassPathRepository.java
```diff
@@ -52,14 +52,15 @@
...
-   * Spring Boot Gradle Java plugin places class files under "BOOT-INF/classes". To load such
+   * the Spring Boot Gradle Java plugin places class files under "BOOT-INF/classes". To load such
    * classes by class name, this mapping keeps track of the special location once they are loaded.
    *
    * <ul>
    *   <li>Key: class name (value from {@link JavaClass#getClassName()}) which has a special class
    *       file name different from its class name. Example: {@code com.google.Foo}
    *   <li>Value: the special class file name as in {@link JavaClass#getFileName()}, a path that
-   *       locates a class file in a class path. Example: {@code BOOT-INF.classes.com.google.Foo}
+   *       locates a class file in a class path entry.
```
suztomo, line 62: This locates a class file in a class path. Not in a class path entry. This FixedSizeClassPathRepository operates on a class path.

Do you see this description is inaccurate?

elharo: A class file name such as BOOT-INF.classes.com.google.Foo does not tell us which jar the file is in. It only tells us where in this jar the file will be. Or am I misunderstanding the special class file name?

suztomo: > A class file name such as BOOT-INF.classes.com.google.Foo does not tell us which jar the file is in

Correct. No specific JAR file. That's why this is a path to locate a class file in a class path, not a JAR file.

elharo: I think I see where you're coming from, but I don't think as written this is correct. For instance, there can be multiple files with the same name in the class path. There can be only one of these in a class path entry. Perhaps we need to get rid of the notion of this as "locating" something. That's not what it does.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FixedSizeClassPathRepository.java
```diff
@@ -114,11 +119,9 @@ public void clear() {
...
   /**
-   * Returns the special location for {@code className}. Null if no special location is known.
-   *
-   * @see #classFileNames
+   * Returns the file name for the class.
    */
-  String getSpecialLocation(String className) {
-    return classFileNames.get(className);
+  String getFileName(String className) {
+    return classFileNames.getOrDefault(className, className);
```
suztomo, line 125: This is nice. Placing getOrDefault here is simplifying other places.

suztomo:
I like the refactoring, although naming "com.google.Foo" as "filename" is strange.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/FixedSizeClassPathRepository.java
```diff
@@ -48,18 +48,23 @@
...
    *
-   * <p>While class name and class file name are the same in most cases, sometimes classes are not
-   * placed in the root of a JAR file to support a framework-specific JAR structure. For example,
-   * Spring Boot Gradle Java plugin places class files under "BOOT-INF/classes". To load such
-   * classes by class name, this mapping keeps track of the special location once they are loaded.
+   * <p>Class names are the fully package qualified names found in Java source code such as
+   * {@code com.google.Foo}. The byte code of this class is normally found in a JAR at the path
+   * com/google/Foo.class. In this case {@code com.google.Foo} is also the file name.
+   * 
+   * A few tools relocate classes into different directories within the JAR and use a special
```
suztomo, line 57:
```suggestion
   * <p>A few tools relocate classes into different directories within the JAR and use a special
```

I like the clear explanation of the difference.

suztomo:
We can define "class file name" as our own term which, we say, corresponds to BCEL's`JavaClass.getFileName()`.

Even java command line tools does not have clear term for it. For example javap's help says it expects "classes", but it actually expects "class file name":

```
suztomo-macbookpro44% jar tf dummy-boot-inf-prefix.jar
META-INF/
META-INF/MANIFEST.MF
BOOT-INF/classes/com/google/firestore/v1beta1/FirestoreGrpc.class
suztomo-macbookpro44% javap -cp dummy-boot-inf-prefix.jar com.google.firestore.v1beta1.FirestoreGrpc
Error: class not found: com.google.firestore.v1beta1.FirestoreGrpc
suztomo-macbookpro44% javap -help
Usage: javap <options> <classes>
where possible options include:
...
```

PR #1406, opened 2020-05-18
update to new proto release

elharo:
@suztomo


elharo:
No, release version is 5.3.0. This is at 5.4.0-SNAPSHOT which is correct.

PR #1407, opened 2020-05-18
abstract equals and hashCode

elharo:
@suztomo I noticed a subtle issue here where two instances of different subclasses could theoretically compare equal. This prevents that by making equals and hashCode abstract in the superclass.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSymbol.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSymbol.java
@@ -65,18 +65,16 @@ public final class MethodSymbol extends Symbol {
     if (other == null || getClass() != other.getClass()) {
       return false;
     }
-    if (!super.equals(other)) {
-      return false;
-    }
     MethodSymbol that = (MethodSymbol) other;
     return isInterfaceMethod == that.isInterfaceMethod
         && name.equals(that.name)
-        && descriptor.equals(that.descriptor);
+        && descriptor.equals(that.descriptor)
+        && this.getClassBinaryName().equals(that.getClassBinaryName());
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(super.hashCode(), name, descriptor, isInterfaceMethod);
+    return Objects.hash(this.getClassBinaryName(), name, descriptor, isInterfaceMethod);
   }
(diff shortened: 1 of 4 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSymbol.java
```diff
@@ -65,18 +65,16 @@ public boolean equals(Object other) {
...
     }
-    if (!super.equals(other)) {
-      return false;
-    }
     MethodSymbol that = (MethodSymbol) other;
     return isInterfaceMethod == that.isInterfaceMethod
         && name.equals(that.name)
-        && descriptor.equals(that.descriptor);
+        && descriptor.equals(that.descriptor)
+        && this.getClassBinaryName().equals(that.getClassBinaryName());
```
suztomo, line 72: Checking the equality of class name here is more clear than `super.equals`. Nice.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/Symbol.java
```diff
@@ -47,21 +46,10 @@ String getClassBinaryName() {
   }
 
   @Override
-  public boolean equals(Object other) {
-    if (this == other) {
-      return true;
-    }
-    if (other == null || getClass() != other.getClass()) {
-      return false;
-    }
```
suztomo, line 56: > two instances of different subclasses could theoretically compare equal.

Doesn't this if statement prevent that?

elharo: You're right. It does.

It still means subclasses can be created with additional data that is not considered in an equality comparison. Mostly it's a little disconcerting to have an equals method that should always be overridden, especially in an abstract class.

suztomo: > It still means subclasses can be created with additional data that is not considered in an equality comparison. 

I'm not sure this change is an improvement on that. Even with this PR, a subclass can declare an additional data that's not considered in equality comparison.

suztomo:
One benefit of this PR is equality for the classes is in within the files. Readers do not need to check superclass's equality.

PR #1411, opened 2020-05-18
use a simple map

### LINKED ISSUE #1409: ClassDumper.fileNameToClassPathEntry should be a Map
Or maybe an ImmutablemMap, but not an ImmutableListMultiMap
 since we always call get first this doesn't have to be a multimap

elharo:
@suztomo fixes #1409


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -69,7 +69,7 @@
   private final ImmutableList<ClassPathEntry> inputClassPath;
   private final FixedSizeClassPathRepository classRepository;
   private final ClassLoader extensionClassLoader;
-  private final ImmutableListMultimap<String, ClassPathEntry> fileNameToClassPathEntry;
+  private final Map<String, ClassPathEntry> fileNameToClassPathEntry;
```
suztomo, line 72: Can you make this ImmutableMap?

Issue #1405, opened 2020-05-18
Replace DependencyPath with Guava Graph


suztomo:
Is it feasible to replace current tree representation of `List<DependencyPath>` with Guava's Graph?

# Considerations

In Maven dependency graph edges are ordered.

Dependency tree represented by `DependencyPath`s are easy to get the path from the root to the leaf (`DependencyPath` itself). Maybe the node can have edges path to root (or parents). This is the biggest difference from the Maven-resolver's native dependency tree, which consists of `DependnecyNode`s. A DependencyNode does not know its parent.

For BOM Dashboard, some existing logic for dependency convergence (irrelevant to Linkage Checker) columns is based on DependencyPath.  Excerpt from `testJaxen`
```
    List<Update> updates = graph.findUpdates();
    Truth.assertThat(updates).hasSize(6);

    List<DependencyPath> conflicts = graph.findConflicts();
    Truth.assertThat(conflicts).hasSize(10);
```

Should we migrate that logic too? Or limit the scope of this PR to the graph used by Linkage Checker.

The OutOfMemoryError in #1256 drives this change. We have to ensure the resulting design would pass Beam’s hcatalog artifact.

# Design

Maven-resolver's `DependencyNode` has Maven artifact, scope, and optional flag.

elharo:
1. We definitely don't want two graph representations.
 
2. We do need to track and store the paths to the dependencies. That's the biggest reason we built this unusual representation. That's also probably why we have the OOMEs. Possibly we can figure out how to stream the processing so we don't have to hold it all in memory at once.

3. This should be a pure refactoring that does not change any tests. 

4. We shouldn't have to change the public API to change the graph representation. However, currently we expose a lot of internal implementation details in multimaps and lists and the like we pass from one class to another so we might have to improve our data encapsulation first before we can adequately explore this.

5. If we work on this, we should time box it. We don't want to burn all our developer cycles on a speculative fix for one bug. 

6. We did consider these classes a couple of years ago, and decided against them. Perhaps it's time to revisit that decision, but it wasn't simply a case of not being familiar with them.

suztomo:
@elharo Thank you for input. "4." is the interesting point; we might be able to keep `DependencyPath` as return types while replacing the internal graph (which is `List<DependencyPath>`) with Guava's Graph.

elharo:
This won't work due to 
b/142723300 and  go/incident-edge-order
[...]

PR #1415, opened 2020-05-20
clean up public API doc

elharo:
@suztomo Trying to focus this on behavior rather than implementation in case we change the implementation for #1405

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
@@ -19,6 +19,7 @@ package com.google.cloud.tools.opensource.dependencies;
 import static com.google.cloud.tools.opensource.dependencies.RepositoryUtility.CENTRAL;
 import static com.google.cloud.tools.opensource.dependencies.RepositoryUtility.mavenRepositoryFromUrl;
 import static com.google.common.collect.ImmutableList.toImmutableList;
+
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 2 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -154,7 +155,7 @@ private DependencyNode resolveCompileTimeDependencies(
 
   /**
    * Finds the full compile time, transitive dependency graph including duplicates, conflicting
-   * versions, and dependencies with 'provided' scope.
+   * versions, and dependencies with provided and optional scope.
```
suztomo, line 158: optional is not a scope. I call "optional flag".

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -19,7 +19,7 @@
 import static com.google.cloud.tools.opensource.dependencies.RepositoryUtility.CENTRAL;
 import static com.google.cloud.tools.opensource.dependencies.RepositoryUtility.mavenRepositoryFromUrl;
 import static com.google.common.collect.ImmutableList.toImmutableList;
-
+import com.google.common.annotations.VisibleForTesting;
```
suztomo, line 22: Add one empty line between static import and non-static import.

elharo: done

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
elharo: true, rewrote

PR #1416, opened 2020-05-20
replace a bunch of continues

elharo:
@suztomo We can simplify the control flow by flipping some boolean conditions and avoiding continue. FYI, in testing it seems it's not just the zeroth entry in a constant pool that might be null. I'm not sure what's going on there.


suztomo:
Thank you for info. One “continue” node has two edges and thus adds 1 complexity point. It makes sense.

PR #1421, opened 2020-05-27
TODOs for discussion

elharo:
@suztomo @netdpb 

To consider and possibly migrate to an alternate graph representation, it could be useful to have a higher layer of abstraction that doesn't expose the implementation as lists of dependency paths and multimaps.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
@@ -34,7 +34,7 @@ public final class ClassPathResult {
   private final ImmutableList<ClassPathEntry> classPath;
 
   /**
-   * An ordered map from class path elements to one or more Maven dependency paths.
+   * An ordered map from class path entries to one or more Maven dependency paths.
    *
    * <p>The keys of the returned map represent Maven artifacts in the resolved class path,
    * including transitive dependencies. The return value of {@link LinkedListMultimap#keySet()}
@@ -43,10 +43,13 @@ public final class ClassPathResult {
    * <p>The values of the returned map for a key (class path entry) represent the different Maven
    * dependency paths from {@code artifacts} to the Maven artifact.
    */
+  // TODO rather than a multimap this should be a DependencyGraph or a DependencyGraphResult to
+  // support alternate representations
   private final ImmutableListMultimap<ClassPathEntry, DependencyPath> dependencyPaths;
 
   private final ImmutableList<UnresolvableArtifactProblem> artifactProblems;
 
+  
   public ClassPathResult(
       ListMultimap<ClassPathEntry, DependencyPath> dependencyPaths,
       Iterable<UnresolvableArtifactProblem> artifactProblems) {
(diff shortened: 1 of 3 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
```diff
@@ -99,6 +99,13 @@ void addPath(DependencyPath path) {
   public List<DependencyPath> list() {
     return new ArrayList<>(graph);
   }
+  
+  /**
+   * @return a mutable copy of the artifacts in this graph in breadth first order
```
suztomo, line 104:
```suggestion
   * Returns a the artifacts in this graph in breadth first order
```

Is mutability important? If not, let's use ImmutableList.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
```diff
@@ -43,10 +43,13 @@
    * <p>The values of the returned map for a key (class path entry) represent the different Maven
    * dependency paths from {@code artifacts} to the Maven artifact.
    */
+  // TODO rather than a multimap this should be a DependencyGraph or a DependencyGraphResult to
+  // support alternate representations
   private final ImmutableListMultimap<ClassPathEntry, DependencyPath> dependencyPaths;
```
suztomo, line 48: Right. This representation is bad, so is the ones in DependencyGraph and DependencyGraphResult. Memo: this exists to support `ImmutableList<DependencyPath> getDependencyPaths(ClassPathEntry entry)`. I think we can implement the method differently without changing the API.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
suztomo: Clarify that this list may contain duplicate; the caller is responsible to perform dependency mediation.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassPathResult.java
```diff
@@ -43,10 +43,13 @@
...
 
   private final ImmutableList<UnresolvableArtifactProblem> artifactProblems;
 
+
```
netdpb, line 52: Revert

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
netdpb: The current implementation returns a list that explicitly has no guarantees on mutability.

netdpb: It currently may return duplicates, but eventually we wouldn't want it to, right?

elharo: I suppose that depends on where the dependency mediation is being performed. If we can move that completely into this class and have this class return only a classpath, then yes. If dependency mediation happens outside of this class, then no.

Issue #1418, opened 2020-05-21
Linkage checker taking too much time for spring-cloud-gcp-dependencies


suztomo:
Follow-up of https://github.com/spring-cloud/spring-cloud-gcp/pull/1658#issuecomment-631578103. I also observe the same thing. 2 hours and still running. No errors except #1417 .

```
suztomo@suztomo:~/spring-cloud-gcp$ git status
On branch linkage-checker
Your branch is up to date with 'origin/linkage-checker'.

nothing to commit, working tree clean
suztomo@suztomo:~/spring-cloud-gcp$ ./mvnw verify -Dmaven.test.skip
...
[INFO] --- maven-enforcer-plugin:3.0.0-M3:enforce (enforce-versions) @ spring-cloud-gcp-dependencies ---
[INFO] 
[INFO] --- maven-enforcer-plugin:3.0.0-M3:enforce (enforce-linkage-checker) @ spring-cloud-gcp-dependencies ---
[WARNING] Could not validate integrity of download from https://repo1.maven.org/maven2/org/slf4j/slf4j-jcl/1.0.1/slf4j-jcl-1.0.1.pom: Checksum validation failed, expected 7035ae7774a9a082a316a6943bbad9dfab6319b3 but is c5c0a3fff6071a4c720f1b7aa1b66cb9d0b26a21
[WARNING] Could not validate integrity of download from https://repo1.maven.org/maven2/org/opensaml/opensaml/maven-metadata.xml: Checksum validation failed, expected b50e385d27683179ca2de5043f96c1252301aca5 but is 7576ac94d82160710ee133d885e0ce71339e0ef1
[WARNING] Could not validate integrity of download from https://repo.maven.apache.org/maven2/org/opensaml/opensaml/maven-metadata.xml: Checksum validation failed, expected b50e385d27683179ca2de5043f96c1252301aca5 but is 7576ac94d82160710ee133d885e0ce71339e0ef1

```

suztomo:
Linkage Checker gets stuck at DependencyGraphBuilder.findArtifactPaths. I guess it's visiting the graph unnecessarily.
 
https://gist.github.com/suztomo/c711510ac56012eab9f32cae7101d154

suztomo:
Does `org.eclipse.aether.util.graph.visitor.PathRecordingDependencyVisitor` avoid revisiting same node twice?

suztomo:
CycleBreakingTransformer is also slow.
[...]

PR #1424, opened 2020-05-28
CycleBreakerGraphTransformer to avoid checking children of a node twice

suztomo:
Towards https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1418#issuecomment-634749982 . Before this PR, CycleBreakerGraphTransformer visits the children of a node multiple times if there are multiple path to the node.

I've added a package-private method to confirm the behavior. This makes the test clear. From caller's perspective, CycleBreakerGraphTransformer's behavior does not change. It just became faster when dealing with non-tree graph. Such as spring-cloud-gcp-dependencies's graph with 645,977 unique nodes.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/CycleBreakerGraphTransformer.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/CycleBreakerGraphTransformer.java
@@ -55,7 +55,7 @@ final class CycleBreakerGraphTransformer implements DependencyGraphTransformer {
       return;
     }
 
-    if (shouldCheckChildren(node)) {
+    if (shouldVisitChildren(node)) {
       ancestors.add(artifact);
       for (DependencyNode child : node.getChildren()) {
         removeCycle(node, child, ancestors);
@@ -66,7 +66,7 @@ final class CycleBreakerGraphTransformer implements DependencyGraphTransformer {
 
   /** Returns true if {@code node} is not visited yet and marks the node as visited. */
   @VisibleForTesting
-  boolean shouldCheckChildren(DependencyNode node) {
+  boolean shouldVisitChildren(DependencyNode node) {
     return visitedNodes.add(node);
   }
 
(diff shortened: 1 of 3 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/CycleBreakerGraphTransformer.java
```diff
@@ -52,6 +55,11 @@ private static void removeCycle(
       return;
     }
 
+    if (visited.put(node, true) != null) {
```
elharo, line 58: you only call put and never get, so this is probably not the simplest data structure we could use here. Maybe a Set?

suztomo: I wish I have Set for object identity. JRE does not provide IdentityHashSet.

suztomo: https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#newSetFromMap-java.util.Map- may work. Let me try this.

suztomo:
@elharo PTAL. It turned out I don't need IdentityHashMap.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/CycleBreakerGraphTransformer.java
```diff
@@ -33,6 +34,9 @@
  */
 final class CycleBreakerGraphTransformer implements DependencyGraphTransformer {
 
+  // DependencyNode inherits Object's equality. No two instances are equal to each other.
```
elharo, line 37: True for now but this is risky. DependencyNode is an interface, not a class. This could change.

suztomo: Good point. Updated to use a set based on IdentityHashMap.

PR #1425, opened 2020-05-28
Linkage Checker to distinguish missing-class and missing-method errors for try-catch

suztomo:
Fixes #1379 

The problem is false negative where method-missing linkage error was suppressed because the source class has try-catch clause for ClassNotFoundException. This is false negative because the combination ("io.grpc:grpc-core:1.17.0", "com.google.guava:guava:20.0") does generate NoSuchMethodError in https://github.com/GoogleCloudPlatform/cloud-opensource-java/tree/master/example-problems/no-such-method-error-signature-mismatch.

### DIFF
--- a/enforcer-rules/src/it/bom-project-error/verify.groovy
+++ b/enforcer-rules/src/it/bom-project-error/verify.groovy
@@ -3,8 +3,10 @@ def buildLog = new File(basedir, "build.log").text.replaceAll("\\r\\n", "\n")
 assert buildLog.contains('''\
 [ERROR] Linkage Checker rule found 1 error. Linkage error report:
 (com.google.guava:guava:20.0) com.google.common.base.Verify's method verify(boolean arg1, String arg2, Object arg3) is not found;
-  referenced by 1 class file
+  referenced by 3 class files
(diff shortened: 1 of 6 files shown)


enforcer-rules/src/it/bom-project-error/verify.groovy
```diff
@@ -3,8 +3,10 @@ def buildLog = new File(basedir, "build.log").text.replaceAll("\\r\\n", "\n")
 assert buildLog.contains('''\
 [ERROR] Linkage Checker rule found 1 error. Linkage error report:
 (com.google.guava:guava:20.0) com.google.common.base.Verify's method verify(boolean arg1, String arg2, Object arg3) is not found;
-  referenced by 1 class file
+  referenced by 3 class files
     io.grpc.internal.ServiceConfigInterceptor (io.grpc:grpc-core:1.17.1)
+    io.grpc.internal.JndiResourceResolverFactory (io.grpc:grpc-core:1.17.1)
+    io.grpc.internal.DnsNameResolver (io.grpc:grpc-core:1.17.1)
```
suztomo, line 9: This is the class that raises NoSuchMethodError in the example project https://github.com/GoogleCloudPlatform/cloud-opensource-java/tree/master/example-problems/no-such-method-error-signature-mismatch

PR #1430, opened 2020-06-01
Omit some redundant words

elharo:
@suztomo a small change I threw out there to see if `gh pr create` would work

- omit redundant words
- 5.6.0-SNAPSHOT
- preparing release 5.5.0-bom


suztomo:
The "gh" command seemed to create the unexpected description for this PR. I assume it's from previous commit log.

elharo:
Yes, I used the `-f` option to create the PR description from the commit messages but we can write our own description with the `-b` option instead.

PR #1442, opened 2020-06-04
Remove DependencyGraphResult

elharo:
@suztomo This is a major refactoring to remove the DependencyGraphResult class so that DependencyGraph holds everything and does more of the work of building that was previously done by DependencyGraphBuilder. In other words, this much more completely encapsulates DependencyGraph logic inside DependencyGraph, which is essential to migrate to a different implementation based on Guava graphs, JGraphT, or something else. Indeed we might now be able to support multiple, parallel DependencyGraph implementations so we can compare them to each other.


elharo:
With this change, DependencyGraphBuilder only knows how Maven builds and represents dependency graphs. All details about our internal representation are encapsulated inside the DependencyGraph class.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -177,68 +173,37 @@ public DependencyGraphResult buildFullDependencyGraph(List<Artifact> artifacts)
...
-      Set<Artifact> checkedArtifacts = new HashSet<>();
       for (ArtifactResult artifactResult : result.getArtifactResults()) {
         Artifact resolvedArtifact = artifactResult.getArtifact();
 
         if (resolvedArtifact == null) {
           Artifact requestedArtifact = artifactResult.getRequest().getArtifact();
-          if (checkedArtifacts.add(requestedArtifact)) {
-            artifactProblems.add(createUnresolvableArtifactProblem(node, requestedArtifact));
-          }
+          graph.addUnresolvableArtifactProblem(requestedArtifact);
```
suztomo, line 200: This method checks whether the artifact is already checked or not. Good.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
```diff
@@ -59,9 +63,14 @@
   
   // map of groupId:artifactId:version to paths
   private SetMultimap<String, DependencyPath> paths = HashMultimap.create();
+
+  private DependencyNode root;
   
-  @VisibleForTesting
-  public DependencyGraph() {
+  DependencyGraph() {
+  }
```
suztomo, line 70: Is this constructor still needed? I see tests uses `new DependencyGraph(null)`.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilder.java
```diff
@@ -271,9 +236,12 @@ public static UnresolvableArtifactProblem createUnresolvableArtifactProblem(
    * @param firstNode node to start traversal
    */
   public static DependencyGraph levelOrder(DependencyNode firstNode) {
-
     DependencyGraph graph = new DependencyGraph();
 
+    return levelOrder(firstNode, graph);
+  }
+
+  private static DependencyGraph levelOrder(DependencyNode firstNode, DependencyGraph graph) {
```
suztomo, line 244: Thinking about other implementation of `DependencyGraph`, this `levelOrder` method of converting DependencyNode to `DependencyGraph` is specific to `DependencyNode`. It's because `graph.addPath(path);` in this method comes from the fact that current `DependencyGraph` consists of `DependnecyPath`s.

Do you think moving this method inside DependencyGraph, as `DependencyGraph.from(DependencyNode root)` is a good idea?

elharo: Good idea. Done, and that simplified some more code too.

elharo:
Artifact problems are part of the dependency graph. They are things that are referenced in the dependency graph and aren't found. 

After this change switching to a different implementation can be done in one class, or a subclass of one class: DependencyGraph.

PR #1445, opened 2020-06-08
Release 5.6.0-bom

elharo:
@suztomo Release 5.6.0-bom


boms/integration-tests/pom.xml
```diff
@@ -18,7 +18,7 @@
     <dependency>
       <groupId>com.google.cloud.tools</groupId>
       <artifactId>dependencies</artifactId>
-      <version>1.4.3-SNAPSHOT</version>
+      <version>5.7.0-SNAPSHOT</version>
```
suztomo, line 21: This is is not the version of the linkage checker.

PR #1453, opened 2020-06-09
Automate a couple more steps when releasing

elharo:
@suztomo


scripts/prepare_release.sh
```diff
@@ -93,7 +98,7 @@ gh pr create -t "Release ${VERSION}-${SUFFIX}" -b "Release ${VERSION}-${SUFFIX}"
 EchoGreen 'Ask someone to approve this PR:'
 echo https://github.com/GoogleCloudPlatform/cloud-opensource-java/compare/${VERSION}-${SUFFIX}
```
suztomo, line 99: Last week, I noticed that "gh pr" shows PR URL. This line can be removed.

PR #1480, opened 2020-06-16
Serializer for dependency graph

ian-lavallee:


verboseTree-maven-plugin/pom.xml
```diff
@@ -70,6 +70,13 @@
       <version>3.6.3</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>2.2</version>
```
elharo, line 76: 2.6

verboseTree-maven-plugin/src/main/java/org/apache/maven/dependency/graph/SerializeGraph.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.IdentityHashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class SerializeGraph
+{
+    // will be injected eventually
+    private String outputType;
+
+    private Map<DependencyNode, Boolean> visitedNodes;
```
elharo, line 33: can these be final?

verboseTree-maven-plugin/src/main/java/org/apache/maven/dependency/graph/SerializeGraph.java
```diff
@@ -0,0 +1,118 @@
...
+        String scope = node.getDependency().getScope();
+        builder.append( getArtifactString( node.getArtifact() ) );
+
+        if ( scope != null && !scope.isEmpty() )
+        {
+            builder.append( ":" ).append( scope );
+        }
+    }
+
+    private static String getArtifactString( Artifact artifact )
```
elharo, line 60: ArtifactString --> CoordinateString

verboseTree-maven-plugin/src/test/java/org/apache/maven/dependency/graph/SerializeGraphTest.java
```diff
@@ -0,0 +1,220 @@
...
+import org.eclipse.aether.graph.Dependency;
+import org.eclipse.aether.graph.DependencyNode;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Arrays;
+
+import static org.apache.commons.io.FileUtils.readFileToString;
```
elharo, line 32: I would tend not to statically import this, but up to you.

verboseTree-maven-plugin/tree.txt
```diff
@@ -0,0 +1,61 @@
+[INFO] Scanning for projects...
```
elharo, line 1: where is this file used?

verboseTree-maven-plugin/src/main/java/org/apache/maven/dependency/graph/SerializeGraph.java
```diff
@@ -0,0 +1,118 @@
...
+package org.apache.maven.dependency.graph;
+
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.graph.DependencyNode;
+
+import java.util.HashSet;
+import java.util.IdentityHashMap;
+import java.util.Map;
+import java.util.Set;
+
```
bimargulies-google, line 27: Please add Javadoc comment explaining purpose of the overall class.

verboseTree-maven-plugin/src/main/java/org/apache/maven/dependency/graph/SerializeGraph.java
```diff
@@ -0,0 +1,118 @@
...
+        if ( scope != null && !scope.isEmpty() )
+        {
+            builder.append( ":" ).append( scope );
+        }
+    }
+
+    private static String getCoordinateString( Artifact artifact )
+    {
+        return artifact.getGroupId() + ":" + artifact.getArtifactId() + ":" +
+                artifact.getExtension() + ":" + artifact.getVersion();
```
bimargulies-google, line 63: Do you need classifier in here?

verboseTree-maven-plugin/src/main/java/org/apache/maven/dependency/graph/SerializeGraph.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    public StringBuilder dfs( DependencyNode node, String start )
+    {
+        builder.append( start );
+        appendDependency( builder, node );
+
+        if ( visitedNodes.containsKey( node ) )
+        {
+            builder.append( " (Omitted due to cycle)" ).append( System.lineSeparator() );
```
bimargulies-google, line 79: Nit. put period after cycle and at the end of the other sentences in this section.

verboseTree-maven-plugin/src/main/java/org/apache/maven/dependency/graph/SerializeGraph.java
stephaniewang526: Also Scope? Optional? How do you determine if this comparison logic is thorough?

ian-lavallee: Based on the documentation https://maven.apache.org/plugins/maven-dependency-plugin/examples/resolving-conflicts-using-the-dependency-tree.html I think they just use GroupID:ArtifactID:Extension:Version:Scope so I don't think I need classifier.

ian-lavallee:
PR is not ready to be merged. Realized I need to add conflicting version and scope logic still.

verboseTree-maven-plugin/src/main/java/org/apache/maven/dependency/graph/SerializeGraph.java
ian-lavallee: @stephaniewang526 thanks for pointing this out, the comparison logic needs some additions before it's ready.

verboseTree-maven-plugin/src/main/java/org/apache/maven/dependency/graph/SerializeGraph.java
```diff
@@ -0,0 +1,122 @@
...
+        builder.append( start );
+        appendDependency( builder, node );
+
+        if ( visitedNodes.containsKey( node ) )
+        {
+            builder.append( " (Omitted due to cycle.)" ).append( System.lineSeparator() );
+        }
+        else if ( isDuplicateArtifact( node.getArtifact() ) )
+        {
+            builder.append( " (Omitted due to duplicate artifact.)" ).append( System.lineSeparator() );
```
stephaniewang526, line 120: nit: Maven -Dverbose syntax looks like `(com.google.auto.value:auto-value-annotations:jar:1.7.2:compile - omitted for duplicate)`. We probably want to be consistent?

verboseTree-maven-plugin/src/main/java/org/apache/maven/dependency/graph/SerializeGraph.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+        else
+        {
+            artifactSet.add( getCoordinateString( node.getArtifact() ) );
+            builder.append( System.lineSeparator() );
+            visitedNodes.put( node, true );
+
+            for ( int i = 0; i < node.getChildren().size(); i++ )
+            {
+                if ( start.endsWith( "+- " ) )
```
stephaniewang526, line 143: Checking hardcoded string tends to be not future-proof -- as soon as those string changes, your logic will break. Is there any other way for you to tell where the dependency is? If not, can you make this logic more dynamic?

verboseTree-maven-plugin/src/main/java/org/apache/maven/dependency/graph/SerializeGraph.java
stephaniewang526: `classifier` should be needed. Here's an example I saw recently:
```
[WARNING] Some problems were encountered while building the effective model for com.google.cloud:google-cloud-bigquerystorage:jar:0.134.0-beta
[WARNING] 'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: com.google.protobuf:protobuf-java:jar -> duplicate declaration of version (?) @ line 159, column 17
```

[Here](https://github.com/googleapis/java-shared-dependencies/blob/772c2a2edc19e708838190e40b8df7eefb1764dc/pom.xml#L234) is a real example where we are using `test` classifier dependency as a test dependency:
```
      <dependency>
        <groupId>com.google.cloud</groupId>
        <artifactId>google-cloud-core</artifactId>
        <version>${google.core.version}</version>
        <classifier>tests</classifier>
      </dependency>
```
Note that `google-cloud-core` appears several times in the pom of java-shared-dependencies repo (in the BOM, as a test-jar, and as a test classified dep).

You may wish to figure out what Maven considers a duplicate dependency.

verboseTree-maven-plugin/src/main/java/org/apache/maven/dependency/graph/SerializeGraph.java
ian-lavallee: Is this an issue since it's that methods responsibility to decide what strings to use anyways, they can't be changed externally? I will change the method to private to reflect that and I'll move the common strings into a variable to make maintaining it easier. Let me know if this isn't good enough because it should be fairly easy to get rid of the string comparison completely.

PR #1493, opened 2020-06-22
Don't follow transitive optional dependencies

elharo:
Frankly astonished this doesn't seem to break any tests.

@suztomo

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -312,7 +312,7 @@ public class DashboardMain {
 
     for (Artifact artifact : artifacts) {
       DependencyGraph completeDependencies =
-          dependencyGraphBuilder.buildFullDependencyGraph(ImmutableList.of(artifact));
+          dependencyGraphBuilder.buildVerboseDependencyGraph(artifact);
       globalDependencies.add(completeDependencies);
 
       // picks versions according to Maven rules
--- a/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
+++ b/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
@@ -444,7 +444,8 @@ public class DashboardTest {
     Document document = parseOutputFile("dependency_trees.html");
     Nodes dependencyTreeParagraph = document.query("//p[@class='dependency-tree-node']");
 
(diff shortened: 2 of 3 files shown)


dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -312,7 +312,7 @@ private static ArtifactCache loadArtifactInfo(List<Artifact> artifacts) {
 
     for (Artifact artifact : artifacts) {
       DependencyGraph completeDependencies =
-          dependencyGraphBuilder.buildFullDependencyGraph(ImmutableList.of(artifact));
+          dependencyGraphBuilder.buildVerboseDependencyGraph(artifact);
```
suztomo, line 315: This change does not affect Linkage Check column in the dashboard. If you want to change the Linkage Checker's class path, you need to update code around line 184:

```
  private static Path generate(Bom bom) throws IOException, TemplateException, URISyntaxException {

    ImmutableList<Artifact> managedDependencies = bom.getManagedDependencies();

    ClassPathResult classPathResult = classPathBuilder.resolve(managedDependencies);
    ImmutableList<ClassPathEntry> classpath = classPathResult.getClassPath();

    LinkageChecker linkageChecker = LinkageChecker.create(classpath);
...
```

suztomo: I just checked buildVerboseDependencyGraph for com.google.http-client:google-http-client:jar:1.35.0. It generates additional linkage errors for the missing classes in the optional dependencies.

Screen Shot 2020-06-23 at 14 44 03

elharo: This is interesting. Two of those might be true positives but I need to investigate. I wonder how this class is guarding those usages?

elharo: After looking into commons-logging I think:

1. Those are unlikely to come up in practice but are true positives.
2.  No one should depend on Commons-logging for multiple reasons, including that it does not properly guard usage of its optional dependencies.

This does add weight to your idea about distinguishing linkage errors caused by dependency conflicts from linkage errors caused by missing optional dependencies.

elharo:
[ERROR] Failures: 
[ERROR]   DashboardTest.testDependencyTrees:447 expected:<53144> but was:<38392>
[INFO] 
[ERROR] Tests run: 43, Failures: 1, Errors: 0, Skipped: 0

Issue #1472, opened 2020-06-15
Gradle plugin: setup release steps


suztomo:
What needs to be done to release the Gradle plugin?

# Learn jib's setup

Kokoro job config: jib/gradle/gcp_ubuntu/release. This points to its build config. The build config specifies the release script "jib-gradle-plugin/scripts/release.sh"

https://github.com/GoogleContainerTools/jib/blob/master/jib-gradle-plugin/scripts/release.sh#L24

```
./gradlew :jib-gradle-plugin:publishPlugins \
  -Pgradle.publish.key="${PUBLISH_KEY}" \
  -Pgradle.publish.secret="${PUBLISH_SECRET}" \
  --info --stacktrace
```

## How does the output look like?

https://plugins.gradle.org/plugin/com.google.cloud.tools.jib

```
classpath "gradle.plugin.com.google.cloud.tools:jib-gradle-plugin:2.4.0"
```

https://plugins.gradle.org/docs/publish-plugin

> By default, the publishing plugin will use the project group, prefixed with gradle.plugin as the maven coordinate group for your plugin.

# Rapid

"Cloud Java Tools Jib Gradle Kokoro Release"

# Questions

## Is it ok to have dependencies outside jcenter?

Yes for Maven Central. JCenter holds artifacts published to Maven Central. For example "com.google.cloud.tools:dependencies:1.4.3" is published to Maven Central and I was able to use it through "jcenter()" next day.

It seems that jcenter's web UI has outdated information but the repository works fine: https://bintray.com/bintray/jcenter/com.google.cloud.tools%3Adependencies (showing 1.3.0 as latest).

# Releasing

```
$ pwd
/Users/suztomo/cloud-opensource-java
$ ./scripts/prepare_release.sh gradle 0.0.1
...
Creating pull request for 0.0.1-gradle into master in GoogleCloudPlatform/cloud-opensource-java

https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1517
...
```

suztomo:
For example, guice is not in jcenter.

```
suztomo-macbookpro44% ../../gradle-plugin/gradlew linkageCheck

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'gradle-project'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not find com.google.inject:guice:4.2.1.
     Searched in the following locations:
[...]

PR #1499, opened 2020-06-25
Release 1.4.3-dependencies

suztomo:
Towards #1472 

The Gradle plugin uses ClassPathEntry.getArtifact. That method needs to be public and the updated Maven artifact "com.google.cloud.tools:dependencies" must be in Maven Central, before releasing the Gradle plugin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Linkage Checker Enforcer Rule Change Log
 
+## 1.4.3
+* Made ClassPathEntry.getArtifact() public for the Linkage Checker Gradle plugin
+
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Linkage Checker Enforcer Rule Change Log
 
+## 1.4.3
+* Made ClassPathEntry.getArtifact() public for the Linkage Checker Gradle plugin
```
elharo, line 4: Fair warning: nothing about ClassPathEntry, including its existence, should be considered stable at this point.

PR #1500, opened 2020-06-25
add method to get verbose graph for list of artifacts

elharo:
@suztomo @ian-lavallee

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/BanOptionalDependencySelector.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/BanOptionalDependencySelector.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.opensource.dependencies;
+
+import org.eclipse.aether.collection.DependencyCollectionContext;
+import org.eclipse.aether.collection.DependencySelector;
+import org.eclipse.aether.graph.Dependency;
+
+/**
+ * A dependency selector that disallows all optional dependencies regardless of depth.
+ */
+class BanOptionalDependencySelector implements DependencySelector {
+
+  @Override
+  public boolean selectDependency(Dependency dependency) {
+      return !dependency.isOptional();
+  }
+
+  @Override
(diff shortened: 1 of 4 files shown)


dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilderTest.java
```diff
@@ -199,6 +201,45 @@ public void testBuildLinkageCheckDependencyGraph_multipleArtifacts() {
...
+  public void testBuildVerboseDependencyGraph_singleArtifact() {
+    DependencyGraph graph =
+        dependencyGraphBuilder.buildVerboseDependencyGraph(logging);
+
+    List<DependencyPath> list = graph.list();
+    DependencyPath firstElement = list.get(0);
+    Assert.assertEquals(
+        "commons-logging",
+        firstElement.getLeaf().getArtifactId());
+    Assert.assertEquals(6, list.size()); // optional dependencies are included
```
ian-lavallee, line 214: Shouldn't the tests assert that the whole graph is what we expect not just the first node and the amount of nodes?

elharo: Ideally yes, assuming that's fixed (and I think it is) so you caught me taking a shortcut. That said, I generally write the simplest test I can until something fails (and this did fail at one point) the fix the code until it passes. Then I write a test for something else until the test fails. Here I don't expect adding that test would cause a failure though the only way to be sure of that would be to actually write it. :-)

elharo: In any case, I added several more asserts here to cover more of the graph. I'm still not testing the exact expected value since that would be quite onerous to setup even for this simple an example. Instead I'm sampling a few know properties of particular parts of the graph.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/BanOptionalDependencySelector.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package com.google.cloud.tools.opensource.dependencies;
+
+import org.eclipse.aether.collection.DependencyCollectionContext;
+import org.eclipse.aether.collection.DependencySelector;
+import org.eclipse.aether.graph.Dependency;
+
+/**
+ * A dependency selector that disallows all optional dependencies regardless of depth.
```
suztomo, line 24: To follow Maven's default behavior, we need to include "direct optional dependencies" (depth 1).

`org.eclipse.aether.util.graph.selector.OptionalDependencySelector` does that.

elharo: That would be only for a specific artifact. This is for a lost of artifacts so all optional dependencies are effectively transitive.

elharo: Per docs,  OptionalDependencySelector "excludes optional dependencies which occur beyond level one of the dependency graph." This need to exclude all optional dependencies, at level one and beyond.

elharo: Indeed using that instead of this breaks the tests.

PR #1501, opened 2020-06-26
Dependency Trees in BOM Dashboard

suztomo:
This "dependency tree" page shows all dependency trees of the artifacts in the BOM. With browser's search function, this page helps to locate the Maven artifacts that depends on certain artifacts, such as `com.google.auto:auto-common:0.10` and `net.ltgt.gradle.incap:incap:0.2`.

# Showing dependency tree for each of the artifacts in the BOM

Screen Shot 2020-06-26 at 14 16 45

# Search for artifacts

The mouse-hover tells the node's parent.

Screen Shot 2020-06-26 at 14 17 16

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/ArtifactResults.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/ArtifactResults.java
@@ -16,6 +16,10 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
+import com.google.cloud.tools.opensource.dependencies.DependencyPath;
+import com.google.common.collect.ImmutableListMultimap;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.ListMultimap;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -33,6 +37,7 @@ public final class ArtifactResults {
   private final Map<String, Integer> results = new HashMap<>();
   private final Artifact artifact;
   private String exceptionMessage;
+  private ImmutableListMultimap<DependencyPath, DependencyPath> dependencyTree;
 
   public ArtifactResults(Artifact artifact) {
     this.artifact = artifact;
@@ -44,6 +49,18 @@ public final class ArtifactResults {
 
   void addResult(String testName, int failures) {
     results.put(testName, failures);
+  }
+
+  void setDependencyTree(ListMultimap<DependencyPath, DependencyPath> dependencyTree) {
+    this.dependencyTree = ImmutableListMultimap.copyOf(dependencyTree);
+  }
+
+  public ImmutableListMultimap<DependencyPath, DependencyPath> getDependencyTree() {
+    return dependencyTree;
+  }
+
+  public DependencyPath getDependencyRoot() {
+    return Iterables.getFirst(dependencyTree.values(), null);
   }
 
   /**
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/ArtifactResults.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/ArtifactResults.java
@@ -16,10 +16,6 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
-import com.google.cloud.tools.opensource.dependencies.DependencyPath;
-import com.google.common.collect.ImmutableListMultimap;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.ListMultimap;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -37,7 +33,6 @@ public final class ArtifactResults {
   private final Map<String, Integer> results = new HashMap<>();
   private final Artifact artifact;
   private String exceptionMessage;
-  private ImmutableListMultimap<DependencyPath, DependencyPath> dependencyTree;
 
   public ArtifactResults(Artifact artifact) {
     this.artifact = artifact;
@@ -49,18 +44,6 @@ public final class ArtifactResults {
 
   void addResult(String testName, int failures) {
     results.put(testName, failures);
-  }
-
-  void setDependencyTree(ListMultimap<DependencyPath, DependencyPath> dependencyTree) {
-    this.dependencyTree = ImmutableListMultimap.copyOf(dependencyTree);
-  }
-
-  public ImmutableListMultimap<DependencyPath, DependencyPath> getDependencyTree() {
-    return dependencyTree;
-  }
-
-  public DependencyPath getDependencyRoot() {
-    return Iterables.getFirst(dependencyTree.values(), null);
   }
 
   /**
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -30,7 +30,6 @@ import com.google.cloud.tools.opensource.dependencies.Bom;
 import com.google.cloud.tools.opensource.dependencies.DependencyGraph;
 import com.google.cloud.tools.opensource.dependencies.DependencyGraphBuilder;
 import com.google.cloud.tools.opensource.dependencies.DependencyPath;
-import com.google.cloud.tools.opensource.dependencies.DependencyTreeFormatter;
 import com.google.cloud.tools.opensource.dependencies.MavenRepositoryException;
 import com.google.cloud.tools.opensource.dependencies.RepositoryUtility;
 import com.google.cloud.tools.opensource.dependencies.Update;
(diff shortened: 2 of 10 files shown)


dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/ArtifactResults.java
```diff
@@ -33,6 +37,7 @@
   private final Map<String, Integer> results = new HashMap<>();
   private final Artifact artifact;
   private String exceptionMessage;
+  private ImmutableListMultimap<DependencyPath, DependencyPath> dependencyTree;
```
elharo, line 40: can we use a more specific type here? i.e. a semantic class instead of a data structure?

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardUnavailableArtifactTest.java
```diff
@@ -118,12 +120,17 @@ public void testDashboardWithRepositoryException()
     validArtifactResult.addResult(DashboardMain.TEST_NAME_UPPER_BOUND, 0);
     validArtifactResult.addResult(DashboardMain.TEST_NAME_DEPENDENCY_CONVERGENCE, 0);
     validArtifactResult.addResult(DashboardMain.TEST_NAME_GLOBAL_UPPER_BOUND, 0);
+    DependencyPath rootValid = new DependencyPath(validArtifact);
+    validArtifactResult.setDependencyTree(ImmutableListMultimap.of(rootValid, rootValid));
```
elharo, line 124: why are we passing the same object as key and value?

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/ArtifactResults.java
suztomo: Will do.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardUnavailableArtifactTest.java
suztomo: It's a convention of the multipmap<DependencyPath, dependencyPath>. I'll try to use `DependencyGraph` and this will be gone.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -348,40 +345,33 @@ private static ArtifactResults generateArtifactReport(
...
           symbolProblemTable.values().stream()
               .flatMap(problemToClasses -> problemToClasses.keySet().stream())
               .distinct()
               .count();
 
-      ListMultimap<DependencyPath, DependencyPath> dependencyTree =
-          DependencyTreeFormatter.buildDependencyPathTree(dependencyPaths);
       Template report = configuration.getTemplate("/templates/component.ftl");
 
-      DependencyPath rootNode = Iterables.getFirst(dependencyTree.values(), null);
```
suztomo, line 376: Now we don't need to depend on `ListMultimap<DependencyPath, DependencyPath>`, because the parent-children relationship is inside `DependencyGraph`.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardUnavailableArtifactTest.java
suztomo: This piece of code is removed.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -348,40 +345,33 @@ private static ArtifactResults generateArtifactReport(
...
       Map<String, Object> templateData = new HashMap<>();
       templateData.put("artifact", artifact);
       templateData.put("updates", convergenceIssues);
       templateData.put("upperBoundFailures", upperBoundFailures);
       templateData.put("globalUpperBoundFailures", globalUpperBoundFailures);
       templateData.put("lastUpdated", LocalDateTime.now());
-      templateData.put("dependencyTree", dependencyTree);
-      templateData.put("dependencyRootNode", rootNode);
+      templateData.put("dependencyGraph", graph);
```
suztomo, line 374: It used to pass parent-children relationship and the root node. Now `DependencyGraph` knows both.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
```diff
@@ -439,6 +439,16 @@ public void testBomCoordinatesInUnstableArtifacts() throws IOException, ParsingE
         "BOM: com.google.cloud:libraries-bom:1.0.0", bomCoordinatesNodes.get(0).getValue());
   }
 
+  @Test
+  public void testDependencyTrees() throws IOException, ParsingException {
+    Document document = parseOutputFile("dependency_trees.html");
+    Nodes dependencyTreeParagraph = document.query("//p[@class='dependency-tree-node']");
+
+    Assert.assertEquals(53144, dependencyTreeParagraph.size());
```
suztomo, line 447: This test is set to use libraries-bom:1.0.0. Therefore this number 53144 does not change when we make changes to libraries-bom.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/ArtifactResults.java
suztomo: This is reverted in favor of `DependencyGraph` class.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
```diff
@@ -79,6 +81,9 @@
   // map of groupId:artifactId:version to paths
   private SetMultimap<String, DependencyPath> paths = HashMultimap.create();
 
+  private final ListMultimap<DependencyPath, DependencyPath> parentToChildren =
+      LinkedListMultimap.create();
```
elharo, line 85: do we need to preserve insertion order here? If so, the type should be LinkedListMultimap

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
```diff
@@ -115,16 +120,29 @@ void addPath(DependencyPath path) {
...
+   * @throws IllegalStateException if the graph is empty
    */
+  public DependencyPath getRootPath() {
+    if (graph.isEmpty()) {
+      throw new IllegalStateException("The graph is empty");
+    }
+    return graph.get(0);
+  }
+
+  /** Returns the dependency paths of the children of the node at {@code dependencyPath}. */
```
elharo, line 140: This comment could be clearer. Is this the paths from the root to what exactly? What are the child paths of a dependency path?

suztomo: Updated as `Returns dependency paths from the root to the dependencies of the node at {@code  dependencyPath}.`.

I thought renaming this to `getDependencyDependencyPaths` but this verbosity does not give clarity. Keeping `getChildPaths`.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
```diff
@@ -115,16 +119,32 @@ void addPath(DependencyPath path) {
...
+   */
+  public DependencyPath getRootPath() {
+    if (graph.isEmpty()) {
+      throw new IllegalStateException("The graph is empty");
+    }
+    return graph.get(0);
+  }
+
+  /**
+   * Returns dependency paths from the root to the dependencies of the node at {@code
```
elharo, line 140: dependencies of the node --> children of the node

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
```diff
@@ -115,16 +119,32 @@ void addPath(DependencyPath path) {
...
+   * dependencyPath}.
    */
+  public List<DependencyPath> getChildPaths(DependencyPath dependencyPath) {
```
elharo, line 143: possibly rename dependencyPath --> node or perhaps dependencyPath --> parent

and now that I think about it, getChildren might be a better method name than getChildPaths

suztomo: Updated to getChildren. Thank you for suggestion.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/DependencyGraph.java
suztomo: Updated.

suztomo:
@elharo The tests pass now.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/DashboardTest.java
elharo: However it appears it does change when we drop transitive optional dependencies. Not unexpected.

suztomo: Yes, that's correct.

PR #1503, opened 2020-06-29
Give examples of adding methods to interfaces as incompatible

elharo:
@suztomo since this is one of the cases we see most frequently


docs/JLBP-0003.md
```diff
@@ -37,6 +37,8 @@ that lays out a number of rules designed to align with common practice.
     when a dependency is bumped to a higher major version.
   - Changing a method signature
   - Removing a method (deprecated or not)
+  - Adding a method to an interface unless you can provide a default implementation
```
suztomo, line 40:
```suggestion
  - Adding a method to an interface without a default implementation
```

I don't think "can" is needed.

PR #1530, opened 2020-07-13
drop optional dependencies when running enforcer rule

### LINKED ISSUE #1464: Optional handling
Use [OptionalDependencySelector](https://maven.apache.org/components/resolver-archives/resolver-LATEST/maven-resolver-util/apidocs/org/eclipse/aether/util/graph/selector/OptionalDependencySelector.html) to more closely match Maven's classpath.

elharo:
@suztomo I am wondering whether this ought to be an option configurable in pom.xml.

fixes #1464


suztomo:
I think we should have an option to show/suppress linkage errors that are caused by optional dependencies, once we implement the `MissingSymbolCause` class associated to `LinkageProblem`s. (I'm inclined to suppress them by default)

PR #1538, opened 2020-07-16
Drop "provided" scope from verbose graphs

### LINKED ISSUE #1537: Dependency Conflict on unimplemented method is not showing up as a cause
https://gist.github.com/suztomo/e40eb79989d20317eb61784a9c8de654

```
(org.apache.hadoop:hadoop-core:0.20.2) org.apache.hadoop.hdfs.server.namenode.FSEditLog$EditLogFileInputStream's method isInProgress() is not implemented in the class;
  referenced by 1 class file
    org.apache.hadoop.hdfs.server.namenode.EditLogInputStream (org.apache.hadoop:hadoop-hdfs:2.6.0)
  Cause:
    Unknown
```

suztomo:
Fixes #1537.

Including provided scope in the dashboard confuses users. For example in #1537, the printed linkage error was referencing a Maven artifact "org.apache.hadoop:hadoop-core:0.20.2", which will not appear the users' class path when they depend on "org.apache.beam:beam-sdks-java-io-hcatalog:jar:2.19.0".

As a side effect, the dashboard will stop showing the linkage errors in appengine-api-sdk.

## Original provide usage such as servlet-api

We may have version conflict with the runtime version (but we do not know that version).

## Provided is used as "compile-only" such as incap-processor

Memo

We're including provided dependencies because their absence causes linkage errors.

Optional and provided are different. Optional are usually guarded.


suztomo:
@elharo I observe including provided-scope dependencies is giving more harm than good.

As provided-scope dependencies of a library do not appear in the class path when users depend on the library. For example a user declaring `org.apache.beam:beam-sdks-java-io-hcatalog:jar:2.19.0` dependency does not see `org.apache.hive.hcatalog:hive-hcatalog-core:2.1.0` or `org.apache.hadoop:hadoop-core:0.20.2` in the project's class path (hadoop-core's case in #1537). The linkage error is about a class that won't exist when depending on beam-sdks-java-io-hcatalog.

Many use the provided scope just to express "compile-only" dependencies to exclude it from users' class paths. The incap-processor's case in #1537 is a good example. The declaration is just for help IDE integration.

suztomo:
@netdpb @elharo Here is my observation that Maven does not include transitive provided-scope dependencies. https://gist.github.com/suztomo/108f44cba9b3562c9fd82d4bdaafd4c6 .

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilderTest.java
```diff
@@ -208,6 +208,8 @@ public void testBuildVerboseDependencyGraph_singleArtifact() {
         dependencyGraphBuilder.buildVerboseDependencyGraph(logging);
 
     List<DependencyPath> list = graph.list();
+    // The list should not include provided dependencies
+    Truth.assertThat(list).hasSize(4);
```
netdpb, line 212: Instead of asserting the size is 4 and then each one individually, you could use fuzzy truth:

```
assertThat(list)
    .comparingElementsUsing(
        Correspondence.transforming(dp -> dp.getLeaf().getArtifactId(), "is path to artifact with ID")
    .containsExactly("commons-logging", "log4j", "logkit", "avalon-framework")
    .inOrder();
```

Then failure would show a good message with all the problems.

suztomo: That's nice. Updated.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilderTest.java
```diff
@@ -208,25 +208,15 @@ public void testBuildVerboseDependencyGraph_singleArtifact() {
...
-        list.get(4).getLeaf().getArtifactId());
-    Assert.assertEquals(
-        "javaee-api",
-        list.get(5).getLeaf().getArtifactId());
-    Assert.assertEquals(6, list.size()); // optional dependencies are included
+
+    // The list should not include provided dependencies
+    Truth.assertThat(list)
+        .comparingElementsUsing(
+            Correspondence.transforming(
```
netdpb, line 215: The Correspondence is a good candidate for extracting into a constant, so this can read

```
assertThat(list)
    .comparingElementsUsing(LEAF_ARTIFACT_ID)
    .containsExactly(...)
    .inOrder();
```

suztomo: Updated to use LEAF_ARTIFACT_ID.

dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/DependencyGraphBuilderTest.java
```diff
@@ -208,25 +213,12 @@ public void testBuildVerboseDependencyGraph_singleArtifact() {
...
-        list.get(3).getLeaf().getArtifactId());
-    Assert.assertEquals(
-        "servlet-api",
-        list.get(4).getLeaf().getArtifactId());
-    Assert.assertEquals(
-        "javaee-api",
-        list.get(5).getLeaf().getArtifactId());
-    Assert.assertEquals(6, list.size()); // optional dependencies are included
+
+    // The list should not include provided dependencies
```
elharo, line 217: direct provided

PR #1539, opened 2020-07-16
Guava's dependency tree should not have srczip in dashboard

suztomo:
Still in progress (I cannot reproduce the dependency graph in my mac)

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
@@ -163,7 +163,7 @@ public final class RepositoryUtility {
         new AndDependencySelector(
             // ScopeDependencySelector takes exclusions. 'Provided' scope is not here to avoid
             // false positive in LinkageChecker.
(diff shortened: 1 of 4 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -177,7 +177,7 @@ static DefaultRepositorySystemSession newSessionForVerboseListDependency(
         new AndDependencySelector(
             // ScopeDependencySelector takes exclusions. 'Provided' scope is not here to avoid
             // false positive in LinkageChecker.
-            new ScopeDependencySelector("test"),
+            new ScopeDependencySelector("test", "system"),
```
elharo, line 180: system scope is a different thing. I'm not sure we want to exclude everything with system scope.

suztomo: Good. I'll exclude jdk groupId for now.

suztomo:
@elharo PTAL. Now special-casing "jdk" groupId.

PR #1550, opened 2020-07-24
Dashboard to show the cause

suztomo:
Screen Shot 2020-07-31 at 10 58 05

The dashboard now shows the message.


suztomo:
Still problem in the dashboard.

Screen Shot 2020-07-31 at 11 05 03

suztomo:
Improvement: mention that all of the paths are either optional or excluded.

suztomo:
If we know they don't cause problem, we can exclude the errors through exclusion file, for the dashboard.

suztomo:
We can run reachability check for the linkage errors in the dashboard. Let's compare the result.

PR #1553, opened 2020-07-27
ExcludedDependency to represent a dependency is missing because of exclusions

### LINKED ISSUE #1551: MissingDependency's description should take exclusions into account
Screen Shot 2020-07-27 at 15 55 35

The MissingDependency's description:

> The valid symbol is in com.google.auto.service:auto-service-annotations:jar:1.0-rc6 at com.google.cloud:google-cloud-nio:0.121.2 (compile) / com.google.auto.value:auto-value:1.7.3 (provided) / com.google.auto.service:auto-service:1.0-rc6 (provided) / com.google.auto.service:auto-service-annotations:1.0-rc6 (compile) but it was not selected because the path contains a provided-scope dependency

This "because" clause is incorrect.

> because the path contains a provided-scope dependency

The reason why auto-service-annotations was not in the dependency graph was auto-value declares auto-service with exclusion element:

```
  <dependencies>
    <dependency>
      <groupId>com.google.auto.service</groupId>
      <artifactId>auto-service</artifactId>
      <version>1.0-rc6</version>
      <scope>provided</scope>
      <exclusions>
        <exclusion>
          <artifactId>auto-service-annotations</artifactId>
          <groupId>com.google.auto.service</groupId>
        </exclusion>
      </exclusions>
    </dependency>
```

suztomo:
Fixes #1551 

ExludedDependency is a special case of MissingDependency where an artifact in the dependency path declares a dependency with an exclusion element, which results in the missing dependency.

In case of https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1551, it was auto-value that declares the exclusion:

```
    <dependency>
      <groupId>com.google.auto.service</groupId>
      <artifactId>auto-service</artifactId>
      <version>1.0-rc6</version>
      <scope>provided</scope>
      <exclusions>
        <exclusion>
          <artifactId>auto-service-annotations</artifactId>
          <groupId>com.google.auto.service</groupId>
        </exclusion>
      </exclusions>
    </dependency>
```

Because of this exclusion, the `auto-service` has linkage error for the class in auto-service-annotations.

### DIFF
--- a/dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ExcludedDependencyTest.java
+++ b/dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ExcludedDependencyTest.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.tools.opensource.classpath;
+
+import static org.junit.Assert.assertEquals;
+
+import com.google.cloud.tools.opensource.dependencies.DependencyPath;
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.artifact.DefaultArtifact;
+import org.eclipse.aether.graph.Dependency;
+import org.junit.Test;
+
+public class ExcludedDependencyTest {
+
+  @Test
+  public void testToString() {
+    Artifact root = new DefaultArtifact("a:b:1");
+    Artifact foo = new DefaultArtifact("com.google:foo:1");
+    Artifact bar = new DefaultArtifact("com.google:bar:1");
+
+    DependencyPath pathRootFooBar =
+        new DependencyPath(root)
+            .append(new Dependency(foo, "test", false))
+            .append(new Dependency(bar, "compile", true));
+
+    ExcludedDependency cause = new ExcludedDependency(pathRootFooBar, foo);
+
+    assertEquals(
+        "The valid symbol is in com.google:bar:jar:1 at a:b:jar:1 / "
+            + "com.google:foo:1 (test) / com.google:bar:1 (compile, optional) "
+            + "but it was not selected because com.google:foo:1 excludes com.google:bar.",
+        cause.toString());
+  }
+}
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExcludedDependency.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExcludedDependency.java
@@ -16,15 +16,14 @@
 
 package com.google.cloud.tools.opensource.classpath;
 
-import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.cloud.tools.opensource.dependencies.Artifacts;
 import com.google.cloud.tools.opensource.dependencies.DependencyPath;
 import java.util.Objects;
 import org.eclipse.aether.artifact.Artifact;
-import org.eclipse.aether.graph.Dependency;
 
-/** An optional or provided-scope dependency is not supplied. */
-class MissingDependency extends LinkageProblemCause {
+/** An dependency is not supplied because {@code excludingArtifact} excludes the dependency. */
+class ExcludedDependency extends MissingDependency {
 
   /**
    * The path from {@code LinkageProblem.sourceClass}'s artifact to an artifact that contains a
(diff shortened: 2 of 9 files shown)


dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/ExcludedDependencyTest.java
```diff
@@ -0,0 +1,48 @@
...
+        new DependencyPath(root)
+            .append(new Dependency(foo, "test", false))
+            .append(new Dependency(bar, "compile", true));
+
+    ExcludedDependency cause = new ExcludedDependency(pathRootFooBar, foo);
+
+    assertEquals(
+        "The valid symbol is in com.google:bar:jar:1 at a:b:jar:1 / "
+            + "com.google:foo:1 (test) / com.google:bar:1 (compile, optional) "
+            + "but it was not selected because com.google:foo:1 excludes com.google:bar.",
```
suztomo, line 45: This is the enhancement of this PR. With ExcludedDependency class, we can represent a case where a dependency is missing because of exclusion element in pom.xml

suztomo:
Because the other artifact is also unselected (for some reason: an optional flag, a scope, or excluded), Maven does not include the missing dependency. It does not show the dependency in `dependency:tree`.

suztomo:
It's better to have a bit concrete example:

Suppose the following dependency path from P to C:1 and P to C:2 in a graph, which is excluded. All dependencies are compile-scope and non-optional, unless noted otherwise.

```
P -> A  (depends on B with exclusion C) -> B -> C:1 (excluded by A; depth: 3)
```

Suppose another version of C at this dependency path.

```
P -> D -> F -> G -> C:2 (depth: 4)
```

Although this C:2's depth (4) is higher than C:1's (3), Maven selects C:2, because C:1 is excluded.

# What if C:2 is unselected in some way?

When C:2 is unselected, there must be other reason than depth, such as D, F, G, or C:2 is a provided-scope or optional dependency. In such case, Maven does not select C:2 regardless of its depth.

In usual troubleshootings (#1551), I don't check why C:2 is unselected because C:2 is invisible to me.


@elharo Do you want to share a scenario you have in mind?

elharo:
Not quite. My question is this case

```
P -> A:1  (depends on B with exclusion C) -> B -> C:1 (excluded by A; depth: 3)
```


```
P -> D ->  A:1  (no exclusions) -> B -> C:1 
```

Is C:1 in the classpath because there's a non-excluded path to it even though the path to A:1 does exclude it?

suztomo:
(In your scenario, it's strange that the first "A:1" has exclusion and the second "A:1" does not.)

> Is C:1 in the classpath because there's a non-excluded path to it even though the path to A:1 does exclude it?

Yes.

suztomo:
> Is this a missing dependency at all? 

Yes, it's a missing dependency. In https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1551, the class `com.google.auto.service.AutoService` is not found because the dependency `com.google.auto.service:auto-service-annotations:1.0-rc6` is missing.

Here is the dependency path:

```
com.google.cloud:google-cloud-nio:0.121.2
\ com.google.auto.value:auto-value:1.7.3 (provided; auto-value:1.7.3 depends on auto-service with exclusion of auto-service-annotations)
  \  com.google.auto.service:auto-service:1.0-rc6 (provided)
    \  com.google.auto.service:auto-service-annotations:1.0-rc6 (compile; excluded because of auto-value)
```

The exclusion is not at the root (`google-cloud-nio:0.121.2`). It's at `auto-value:1.7.3`.

suztomo:
I'll create an integration test case to demonstrate this.

suztomo:
@elharo @netdpb I created an example repository with 3 modules to demonstrate a linkage error caused by excluded dependency. PTAL. https://github.com/suztomo/linkage-error-by-exclusion

netdpb:
> @elharo @netdpb I created an example repository with 3 modules to demonstrate a linkage error caused by excluded dependency. PTAL. https://github.com/suztomo/linkage-error-by-exclusion

OK. What does the linkage checker report today for that? What would you like it to report instead?

suztomo:
@netdpb  The Linkage Checker in the current master tells:

```
Class org.example.c.C is not found;
  referenced by 1 class file
    org.example.b.B (org.example:b:1.0-SNAPSHOT)
  Cause:
    The valid symbol is in org.example:c:jar:1.0-SNAPSHOT at org.example:a:jar:1.0-SNAPSHOT / 
    org.example:b:1.0-SNAPSHOT (compile) / org.example:c:1.0-SNAPSHOT (compile) but it was not selected
```
(This does not explain why the artifact `org.example:c:1.0-SNAPSHOT` is missing)

With this PR 1553, Linkage Checker tells:

```
Class org.example.c.C is not found;
  referenced by 1 class file
    org.example.b.B (org.example:b:1.0-SNAPSHOT)
  Cause:
    The valid symbol is in org.example:c:jar:1.0-SNAPSHOT at org.example:a:jar:1.0-SNAPSHOT / 
    org.example:b:1.0-SNAPSHOT (compile) / org.example:c:1.0-SNAPSHOT (compile) but it was not selected
    because org.example:a:1.0-SNAPSHOT excludes org.example:c.
```

The last part of the description tells the exclusion was there.

netdpb:
OK, that looks good.

What gets reported if, in addition to that exclusion, there's a different non-excluded path from a to c, at a different version of c that also doesn't have the `org.example.c.C` symbol?

suztomo:
@netdpb It's reported as a dependency conflict. (In this case exclusion does not matter because the different version of c is closer to the root)

With `org.example:c:2.0-SNAPSHOT`, it reports as below:

```
Class org.example.c.C is not found;
  referenced by 1 class file
    org.example.b.B (org.example:b:1.0-SNAPSHOT)
  Cause:
    Dependency conflict: org.example:c:2.0-SNAPSHOT does not define Class org.example.c.C but org.example:c:1.0-SNAPSHOT defines it.
      selected: org.example:a:jar:1.0-SNAPSHOT / org.example:c:2.0-SNAPSHOT (compile)
      unselected: org.example:a:jar:1.0-SNAPSHOT / org.example:b:1.0-SNAPSHOT (compile) / org.example:c:1.0-SNAPSHOT (compile)
```

suztomo:
@elharo @netdpb Ping.

PR #1558, opened 2020-08-03
The enforcer rule to follow Maven's way for optional for non-BOM project

suztomo:
For BOM projects, it has been handled by https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1530.

For non-BOM projects, this PR 1558 changes the dependency graph resolution so that the Linkage Checker Enforcer Rule follows Maven's way to handle optional (not to include transitive optional dependencies).


enforcer-rules/src/main/java/com/google/cloud/tools/dependencies/enforcer/LinkageCheckerRule.java
```diff
@@ -276,6 +277,7 @@ private static ClassPathResult findProjectClasspath(
           new AndDependencySelector(
               new NonTestDependencySelector(),
               new ExclusionDependencySelector(),
+              new OptionalDependencySelector(),
```
suztomo, line 280: This is Maven's default behavior of handling optional.

PR #1555, opened 2020-08-01
The Gradle plugin to show the cause of linkage problems

suztomo:
- Gradle plugin to construct `com.google.cloud.tools.opensource.dependencies.DependencyGraph`. LinkageProblemCauseAnnotator uses this graph to explain the cause of linkage problems.

PR #1559, opened 2020-08-04
Making LinkageProblem cause field final

suztomo:
Followup of  https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1555#discussion_r465093781 .

LinkageProblem's cause field has been mutable because we do not know the cause when we create LinkageProblem instances. This PR makes it final to make the class immutable.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
@@ -22,7 +22,23 @@ package com.google.cloud.tools.opensource.classpath;
  * as {@link AbstractMethodError}s at runtime.
  */
 final class AbstractMethodProblem extends IncompatibleLinkageProblem {
+  private final MethodSymbol methodSymbol;
+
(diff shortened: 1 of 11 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
```diff
@@ -22,7 +22,23 @@
...
   AbstractMethodProblem(ClassFile sourceClass, ClassFile targetClass, MethodSymbol methodSymbol) {
-    super("is not implemented in the class", sourceClass, targetClass, methodSymbol);
+    this(sourceClass, targetClass, methodSymbol, null);
+  }
+
+  private AbstractMethodProblem(
+      ClassFile sourceClass,
+      ClassFile targetClass,
+      MethodSymbol methodSymbol,
+      LinkageProblemCause cause) {
```
elharo, line 35: This shouldn't even be passed here if we don't know it when the object is constructed.

suztomo:
Memo: Immutability is nice. But to making the class immutable in this PR introduces these additional lines of code. This increases complexity, compared to just one mutable field that we know is modified by only LinkageProblemCauseAnnotator.

PR #1560, opened 2020-08-04
Cont. of The Gradle plugin to show the cause of linkage problems

suztomo:
Cont of https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1555, without touching LinkageProblem's immutability.

Gradle plugin to construct `com.google.cloud.tools.opensource.dependencies.DependencyGraph`. LinkageProblemCauseAnnotator uses this graph to explain the cause of linkage problems.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/LevelOrderQueueItem.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/LevelOrderQueueItem.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2020 Google LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/LevelOrderQueueItem.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.Objects;
+
+/**
+ * The item in the queue for the graph traversal when constructing {@link DependencyGraph}.
+ *
+ * @param <T> The type of the input graph node. It's {@code DependencyNode} for Maven's graph and
+ *     {@code ResolvedDependency} for Gradle's graph.
+ */
+public final class LevelOrderQueueItem<T> {
```
elharo, line 27: We should consider renaming this class

PR #1519, opened 2020-07-06
To use gradle.properties under $HOME/.gradle

suztomo:
`$HOME/.gradle/gradle.properties` file is a recommended way to use API key: https://guides.gradle.org/publishing-plugins-to-gradle-plugin-portal/

PR #1567, opened 2020-08-07
Release 1.0.0-gradle

suztomo:
Release 1.0.0-gradle


suztomo:
The release script failed. There was a missing piece when I implemented #1519 .

PR #1577, opened 2020-08-07
Gradle plugin release script fix

suztomo:
#1519 had two issues

- The gradle property file needs a new line character to split the two variable
- Gradle needs the `GRADLE_USER_HOME` environment variable. (The default value uses `USER_HOME`, which is usually undefined)

I confirmed that this change #1575 worked good in Rapid.

### DIFF
--- a/gradle-plugin/kokoro/release.sh
+++ b/gradle-plugin/kokoro/release.sh
@@ -5,14 +5,15 @@ set -o xtrace
 
 mkdir -p $HOME/.gradle
 readonly HOME_GRADLE_PROPERTY="$HOME/.gradle/gradle.properties"
-export GRADLE_USER_HOME=$HOME/.gradle
+export GRADLE_USER_HOME="$HOME/.gradle"
 
 # Recommended way to store API key in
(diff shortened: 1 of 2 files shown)


gradle-plugin/kokoro/release.sh
```diff
@@ -4,13 +4,15 @@ set -o errexit
 set -o xtrace
 
 mkdir -p $HOME/.gradle
-readonly HOME_GRADLE_PROPERTY="$HOME/.gradle/gradle.properties"
+export GRADLE_USER_HOME="$HOME/.gradle"
```
suztomo, line 7: > GRADLE_USER_HOME
> Specifies the Gradle user home directory (which defaults to $USER_HOME/.gradle if not set).

https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_environment_variables

I assumed the default value should work fine. However, the default value uses a strange variable `USER_HOME`, which is undefined. (This is not `HOME`.)

suztomo:
Ready for review now.

PR #1581, opened 2020-08-11
Clarify that a BOM shall only be for your own project's modules

stephenc:
I've seen far too many bad BOMs. Please ensure that people undertsand the BOM is only for the stuff you ship and not for the dependencies of the stuff you ship

### DIFF
--- a/docs/JLBP-0015.md
+++ b/docs/JLBP-0015.md
@@ -23,9 +23,13 @@ in their own `<dependencyManagement>` section and omit the
 versions from the specific modules they import. The versions will
 be set by the BOM. This provides consistent versions for artifacts released
 by a single project.
-Your project's BOM should include every non-test-scoped module in the project.
+Your project's BOM shall include only those non-test-scoped modules in *your* project.
 Otherwise, a consumer who depends on an unspecified module must specify and
 upgrade the version of that module, defeating the purpose of the BOM.
+
+Your project's BOM shall not include any of your dependencies on other libraries.
+If your BOM specifies versions of a third party library then that will lead to BOM dependency 
+conflicts which are exceedingly tricky to resolve.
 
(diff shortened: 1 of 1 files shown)


stephenc:
@googlebot I signed it!

docs/JLBP-0015.md
```diff
@@ -23,10 +23,14 @@ in their own `<dependencyManagement>` section and omit the
 versions from the specific modules they import. The versions will
 be set by the BOM. This provides consistent versions for artifacts released
 by a single project.
-Your project's BOM should include every non-test-scoped module in the project.
+Your project's BOM shall include only those non-test-scoped modules in *your* project.
```
elharo, line 26: We haven't been using standards language like SHALL and MUST here. I think this sentence already says this, but we can clarify as you suggest below.

docs/JLBP-0015.md
```diff
@@ -23,10 +23,14 @@ in their own `<dependencyManagement>` section and omit the
...
 Otherwise, a consumer who depends on an unspecified module must specify and
 upgrade the version of that module, defeating the purpose of the BOM.
 
+Your project's BOM shall not include any of your dependencies on other libraries.
```
elharo, line 30: shall --> should

docs/JLBP-0015.md
```diff
@@ -23,10 +23,14 @@ in their own `<dependencyManagement>` section and omit the
...
+If your BOM specifies versions of a third party library then that will lead to BOM dependency
```
elharo, line 31: How about "Specifying versions of libraries from other projects can lead to BOM dependency conflicts that are exceedingly tricky to resolve"?

stephenc:
Fine with any changes you make which result in less risk of bad-style BOMs

PR #1594, opened 2020-08-14
remove instanceof and cast in favor of null check

### LINKED ISSUE #1529: The if-statement for SuperClassSymbol be pushed to ClassSymbol's logic?
To track Elliotte's finding.

> if blocks testing subtypes suggests there's a missed opportunity to use polymorphism. Could the logic of each if block be pushed into a method in ClassSymbol and its subtypes instead?

https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1523/files/32c0314ba2dfc06b12d7b2968d0a78a5441ae24d#diff-e275d9cce5c577df64bfd6fe55a5cdfa

elharo:
@suztomo This might be a matter of taste, and I've simply learned to be very wary of instanceof, but this is how we can fix #1529


suztomo:
IncompatibleLinkageProblem's role is to distinguish whether a linkage error has a target ClassFile or not.

Do you think it's worth keeping IncompatibleLinkageProblem?

PR #1604, opened 2020-08-18
Proper error message for AbstractMethodProblem and correct source and target

### LINKED ISSUE #1603: Error for unimplemented method from interface is backwards?
I think the direction of the reference is backwards when complaining about an unimplemented method.  For example, I got the following error:

```
(cglib:cglib:2.2) net.sf.cglib.transform.ClassVisitorTee's method visitMethod(int, String, String, String[], org.objectweb.asm.Attribute) is not implemented in the class;
  referenced by 1 class file
    org.objectweb.asm.ClassVisitor (asm:asm:1.5.3)
  Cause:
    Unknown
```

However, ClassVisitor is the interface and ClassVisitorTee is the implementation. So I believe the error message should say:

```
(asm:asm:1.5.3) org.objectweb.asm.ClassVisitor's method visitMethod(int, String, String, String[], org.objectweb.asm.Attribute) is not implemented in the class;
  referenced by 1 class file
    net.sf.cglib.transform.ClassVisitorTee (cglib:cglib:2.2)
  Cause:
    Unknown
```

I believe you also need to put the exclusion backwards:

```
    <LinkageError>
        <Target>
            <Class name="net.sf.cglib.transform.ClassVisitorTee"/>
        </Target>
        <Source>
            <Class name="org.objectweb.asm.ClassVisitor"/>
        </Source>
    </LinkageError>
```

which is especially weird since you might need to reverse the Target/Source if ClassVisitorTee is calling a method that doesn't exist in this version of ClassVisitor.

suztomo:
Fixes #1603 

- Enhancement on wording on AbstractMethodProblem. This focus on the error against the implementation class aligns how IDE shows errors for them.
- The source and target are now properly set. The implementation class is the source of the references and the interfaces or abstract classes are the targets.

# Example Message
Using guava-jdk5 and guava:27.1-android's incompatibility:

Normally, we report linkage errros grouped by targets:

```
(Target Class X) having problem XXX. referenced by
   (Source Class A)
   (Source Class B)
```

Before this change:

```
(com.google.guava:guava-jdk5:13.0) com.google.common.util.concurrent.RateLimiter$Bursty's method doGetRate() is not implemented in the class;    # Implementing Class
  referenced by 1 class file
    com.google.common.util.concurrent.RateLimiter (com.google.guava:guava:27.1-android). # Interface
```

After this change, we do not group AbstractMethodProblem  

```
com.google.common.util.concurrent.RateLimiter$Bursty (in com.google.api:gax-grpc:1.38.0) does not implement doGetRate(),
required by com.google.common.util.concurrent.RateLimiter (in com.google.guava:guava:27.1-android)
```


dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerMainIntegrationTest.java
```diff
@@ -131,7 +131,7 @@ public void testBom()
       LinkageCheckerMain.main(new String[] {"-b", "com.google.cloud:libraries-bom:1.0.0"});
       fail("LinkageCheckerMain should throw LinkageCheckResultException upon errors");
     } catch (LinkageCheckResultException expected) {
-      assertEquals("Found 800 linkage errors", expected.getMessage());
+      assertEquals("Found 801 linkage errors", expected.getMessage());
```
suztomo, line 134: ???

suztomo: LinkageCheckerMain groups by the targets of the linkage errors. As we're flipping the source and target, the number of groups changes.

suztomo:
Test fails with dashboard assertion, as expected. To fix later if we decide to go for this.

Also I noticed the dashboard has duplicate "The following paths contain com.google.guava:guava-jdk5:13.0:" message somehow for libraries-bom:1.0.0.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageProblem.java
```diff
@@ -150,9 +150,21 @@ public String formatSymbolProblem() {
...
+    // "... referenced by ..." format.
+    ImmutableSet.Builder<AbstractMethodProblem> abstractMethodProblems = ImmutableSet.builder();
+    ImmutableSet.Builder<LinkageProblem> problemsToGroupBySymbols = ImmutableSet.builder();
+    for (LinkageProblem linkageProblem : linkageProblems) {
+      if (linkageProblem instanceof AbstractMethodProblem) {
+        abstractMethodProblems.add((AbstractMethodProblem) linkageProblem);
+      } else {
+        problemsToGroupBySymbols.add(linkageProblem);
+      }
+    }
```
suztomo, line 163: When printing linkage problems, we group them by symbols with "... referenced by ..." format. With this PR, AbstractMethodProblem are printed without the grouping.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSymbol.java
```diff
@@ -94,4 +94,11 @@ public String toString() {
         + "'s method "
         + signature;
   }
+
+  /** Returns the method name with its signature. */
+  public String getMethodNameWithSignature() {
```
suztomo, line 99: toString would print the class name too, which was unnecessary for showing AbstractMethodProblem's toString.

enforcer-rules/src/it/abstract-method-errors/verify.groovy
```diff
@@ -0,0 +1,10 @@
+def buildLog = new File(basedir, "build.log").text.replaceAll("\\r\\n", "\n")
+
+assert buildLog.contains('''\
+com.google.api.gax.grpc.InstantiatingGrpcChannelProvider (in com.google.api:gax-grpc:1.38.0) does not implement com.google.api.gax.rpc.TransportChannelProvider withCredentials(com.google.auth.Credentials), required by com.google.api.gax.rpc.TransportChannelProvider (in com.google.api:gax:1.48.0)
+  Cause:
+    Dependency conflict: com.google.api:gax:1.48.0 defines incompatible version of com.google.api.gax.rpc.TransportChannelProvider but com.google.api:gax:1.38.0 defines compatible one.
+      selected: com.google.cloud.tools.opensource:abstract-method-error-example:jar:1.0-SNAPSHOT / com.google.api:gax:1.48.0 (compile)
+      unselected: com.google.cloud.tools.opensource:abstract-method-error-example:jar:1.0-SNAPSHOT / com.google.api:gax-grpc:1.38.0 (compile) / com.google.api:gax:1.38.0 (compile)''')
+
```
suztomo, line 9: This is the enhancement.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
```diff
@@ -17,12 +17,29 @@
 package com.google.cloud.tools.opensource.classpath;
 
 /**
- * The referenced {@code methodSymbol} is not implemented in the {@code targetClass} but the class
- * is declared to implement the method by {@code sourceClass}. Such unimplemented methods manifest
+ * The referenced {@code methodSymbol} is not implemented in the {@code sourceClass} but the class
+ * is declared to implement the method by {@code targetClass}. Such unimplemented methods manifest
```
suztomo, line 21: Switching source and target. Now implementation class is source because it references the interface.

elharo (commented):
I'm having trouble following this, and I think it's because "source class" and "target class" are not the right words to use here. What else can we call these that might be more obvious?

suztomo:
The source class can be renamed to "implementationClass". The target can be an interface or abstract class.

"requirementClass"?

suztomo:
@elharo How about the followings?

```
AbstractMethodProblem(
      ClassFile implementationClass,
      ClassFile interfaceOrAbstractClass,
      MethodSymbol methodSymbol)
```

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
```diff
@@ -17,15 +17,22 @@
 package com.google.cloud.tools.opensource.classpath;
 
 /**
- * The referenced {@code methodSymbol} is not implemented in the {@code sourceClass} but the class
- * is declared to implement the method by {@code targetClass}. Such unimplemented methods manifest
- * as {@link AbstractMethodError}s at runtime.
+ * The referenced {@code methodSymbol} is not implemented in the {@code implementationClass} but the
```
elharo, line 20: active voice: `{@code implementationClass} does not implement the abstract method {@code methodSymbol} declared by {@code interfaceOrAbstractClass}`

marquiswang: Quick thought that I want to make sure you guys are thinking about. Presumably we still want to be able to add an exclusion to ignore an AbstractMethodProblem. The exclusions file uses the "source" and "target" terminology. Switching this to use implementationClass and interfaceOrAbstractClass might make it a bit confusion which one maps tot source/target when excluding. I personally thought calling the implementClass the "source" class was clear enough.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
```diff
@@ -17,12 +17,36 @@
 package com.google.cloud.tools.opensource.classpath;
 
 /**
- * The referenced {@code methodSymbol} is not implemented in the {@code targetClass} but the class
- * is declared to implement the method by {@code sourceClass}. Such unimplemented methods manifest
- * as {@link AbstractMethodError}s at runtime.
+ * {@code implementationClass} (the source of an invalid reference) does not implement the abstract
+ * method {@code methodSymbol} declared by {@code interfaceOrAbstractClass} (the target of the
+ * reference). Such unimplemented methods manifest as {@link AbstractMethodError}s at runtime.
```
suztomo, line 22: @marquiswang I updated the javadoc to clarify `implementationClass` is the source and `interfaceOrAbstractClass` is the target of the reference.

elharo: This still doesn't work. I don't see how implementationClass is the "source of an invalid reference." There simply is no invalid reference here. There is a problem but it is something else entirely. It is **not** a reference.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageProblem.java
```diff
@@ -150,9 +150,21 @@ public String formatSymbolProblem() {
   public static String formatLinkageProblems(Set<LinkageProblem> linkageProblems) {
     StringBuilder output = new StringBuilder();
 
+    // Not to group AbstractMethodProblems by symbols because they do not fit in the
```
elharo, line 153: Not to --> Don't

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSymbol.java
elharo: Fine, though it probably doesn't need to be public

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
suztomo: It is a reference. Implementation class references an interface via `implements` keyword or an abstract class via `extends` keyword. Implementation-wise, it's a reference embed in the constant pool section of a class file.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageProblem.java
suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/MethodSymbol.java
suztomo: Good catch. Removed public.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
elharo: Yes, but this reference is not invalid. It's a valid, found reference. The problem is that the subtype doesn't implement an abstract method. Sources, targets, and references are not the conventionally understood way to describe this sort of problem.

suztomo: I drew a diagram to explain how I recognize that invalid references describe the abstract method problems in the same manner as missing field problems.

Screen Shot 2020-08-20 at 16 33 02

I'll setup.a meeting to understand your viewpoint better.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
```diff
@@ -17,12 +17,39 @@
...
-    super("is not implemented in the class", sourceClass, methodSymbol, targetClass);
+  MethodSymbol methodSymbol;
+
+  AbstractMethodProblem(
+      ClassFile implementationClass,
+      ClassFile interfaceOrAbstractClass,
+      MethodSymbol methodSymbol) {
+
+    // implementationClass is the source of the invalid symbolic reference, and
+    // interfaceOrAbstractClass is the target class of the reference.
```
elharo, line 33: target class --> target
reference --> symbolic reference

or just delete the comment completely. It's not obviously helpful.

elharo, line 33: target class --> target
reference --> symbolic reference

or just delete the comment completely. It's not obviously helpful.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
```diff
@@ -17,12 +17,39 @@
...
  */
 final class AbstractMethodProblem extends LinkageProblem {
-  AbstractMethodProblem(ClassFile sourceClass, ClassFile targetClass, MethodSymbol methodSymbol) {
-    super("is not implemented in the class", sourceClass, methodSymbol, targetClass);
+  MethodSymbol methodSymbol;
+
+  AbstractMethodProblem(
+      ClassFile implementationClass,
+      ClassFile interfaceOrAbstractClass,
+      MethodSymbol methodSymbol) {
```
elharo, line 30: methodSymbol should be the second argument and interfaceOrAbstractClass the third to match the argument order in the superclass constructor.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
```diff
@@ -17,12 +17,39 @@
...
+        " does not exist in the implementing class",
+        implementationClass,
+        methodSymbol,
+        interfaceOrAbstractClass);
+    this.methodSymbol = methodSymbol;
+  }
+
+  @Override
+  public final String toString() {
+    ClassFile sourceClass = getSourceClass();
```
elharo, line 44: sourceClass -->  implementationClass

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
```diff
@@ -17,12 +17,39 @@
...
+    ClassPathEntry sourceClassPathEntry = sourceClass.getClassPathEntry();
+    ClassFile targetClass = getTargetClass();
```
elharo, line 46: targetClass --> interfaceOrAbstractClass, or perhaps supertype

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
```diff
@@ -17,12 +17,39 @@
...
+ * AbstractMethodError}s at runtime.
  */
 final class AbstractMethodProblem extends LinkageProblem {
-  AbstractMethodProblem(ClassFile sourceClass, ClassFile targetClass, MethodSymbol methodSymbol) {
-    super("is not implemented in the class", sourceClass, methodSymbol, targetClass);
+  MethodSymbol methodSymbol;
+
+  AbstractMethodProblem(
+      ClassFile implementationClass,
+      ClassFile interfaceOrAbstractClass,
```
elharo, line 29: interfaceOrAbstractClass could perhaps be named supertype instead since a class can have more than one supertype when interfaces are in play.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/DependencyConflict.java
```diff
@@ -62,18 +62,35 @@ public DependencyPath getPathToSelectedArtifact() {
...
-        + " but "
-        + Artifacts.toCoordinates(unselected)
-        + " defines it.\n"
-        + "  selected: "
-        + pathToSelectedArtifact
-        + "\n  unselected: "
-        + pathToArtifactThruSource;
+    String unselectedCoordinates = Artifacts.toCoordinates(unselected);
+
+    if (linkageProblem instanceof AbstractMethodProblem) {
```
elharo, line 69: Anytime I see instanceof I assume it can be replaced with polymorphism. This can be a single method in LinkageProblem.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -310,15 +310,18 @@ private boolean problemFilter(LinkageProblem linkageProblem) {
    * methods manifest as {@link AbstractMethodError}s at runtime.
    */
   private ImmutableList<LinkageProblem> findInterfaceProblems(
-      ClassFile classFile, InterfaceSymbol interfaceSymbol, ClassFile sourceClassFile) {
+      ClassFile interfaceClassFile,
```
elharo, line 313: This is clearer. Thanks.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageProblem.java
```diff
@@ -150,9 +150,21 @@ public String formatSymbolProblem() {
   public static String formatLinkageProblems(Set<LinkageProblem> linkageProblems) {
     StringBuilder output = new StringBuilder();
 
+    // Don't group AbstractMethodProblems by symbols because they do not fit in the
+    // "... referenced by ..." format.
+    ImmutableSet.Builder<AbstractMethodProblem> abstractMethodProblems = ImmutableSet.builder();
+    ImmutableSet.Builder<LinkageProblem> problemsToGroupBySymbols = ImmutableSet.builder();
+    for (LinkageProblem linkageProblem : linkageProblems) {
+      if (linkageProblem instanceof AbstractMethodProblem) {
```
elharo, line 158: There's probably a cleaner way to do this without instanceof.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
suztomo: Updated to supertype

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
suztomo: Nice idea. Moved methodSymbol.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
suztomo: Updated. I keep this comment as I'm likely to forget our discussion.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
suztomo: Updated this to supertype.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/DependencyConflict.java
suztomo: I'm feeling that's unnecessary abstraction to remove just one if-statement. Let me try what it looks like.

dashboard/src/main/resources/templates/macros.ftl
```diff
@@ -52,7 +52,9 @@
   <#list jarsInProblem?values as jarInProblem>
     <@showDependencyPath dependencyPathRootCauses classPathResult jarInProblem />
   </#list>
-  <@showDependencyPath dependencyPathRootCauses classPathResult classPathEntry />
```
suztomo, line 55: In the dashboard for libraries-bom 1.0.0, guava-jdk5 was appearing twice, because it involves both source and target.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
```diff
@@ -25,16 +29,12 @@
...
+      ClassFile implementationClass, MethodSymbol methodSymbol, ClassFile supertype) {
 
     // implementationClass is the source of the invalid symbolic reference, and supertype is the
     // target of the symbolic reference.
     super(
-        " does not exist in the implementing class",
-        implementationClass,
-        methodSymbol,
-        supertype);
+        " does not exist in the implementing class", implementationClass, methodSymbol, supertype);
```
elharo, line 37: It's consistent that we have a space at the beginning and not at the end of the first argument. Something to look at later, not in this PR.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageProblem.java
elharo: It might require some reorganizing of how and where we group things. This is OK for now. I'll take a look at this when I get a minute.

suztomo: Thank you. I created https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1618 as a followup.

PR #1602, opened 2020-08-17
Linkage Checker to skip references to JDK system classes

suztomo:
The JDK's tools.jar contains one linkage error.

```
[ERROR] Linkage Checker rule found 1 error. Linkage error report:
Class com.sun.xml.internal.ws.api.BindingID$SOAPHTTPImpl is not accessible;
  referenced by 1 class file
    com.sun.tools.internal.ws.wscompile.WsgenOptions (com.sun:tools:1.8)
  Cause:
    Unknown
```

# Diagnosis

`com.sun.xml.internal.ws.api.BindingID$SOAPHTTPImpl` (in rt.jar) is not a public class. Therefore `com.sun.tools.internal.ws.wscompile.WsgenOptions` (in tools.jar; different package) cannot access the class.

Linkage Checker users cannot take any action on this.

# Proposed Solution

As we cannot take any action on JDK's system classes, we don't need to report linkage errors if the target of the reference is a system classes.

PR #1606, opened 2020-08-19
Adding the invalid reference from tools.jar into the default exclusion rule

suztomo:
https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1602/files tried to check whether a class is from JDK's JAR files (tools.jar). However from the current way to check it via the ClassLoaders, it cannot tell `com.sun.tools.internal.ws.wscompile.WsgenOptions` is from JDK or not.


suztomo:
This was false positive. https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1608

PR #1614, opened 2020-08-21
reference --> dependency

elharo:
@suztomo

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -24,7 +24,7 @@ permalink: /glossary
 - **Linkage error**: an abnormal condition of a classpath in which a
   "class has some dependency on another class; however, the
   latter class has incompatibly changed after the compilation of the
-  former class."<sup>[1](#myfootnote1)</sup> The reference can be
+  former class."<sup>[1](#myfootnote1)</sup> The dependency can be
   through a class literal, a field access, or a method invocation.
   Linkage errors encountered at runtime manifest as a subclass of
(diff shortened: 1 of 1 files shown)


docs/glossary.md
```diff
@@ -24,7 +24,7 @@ permalink: /glossary
 - **Linkage error**: an abnormal condition of a classpath in which a
   "class has some dependency on another class; however, the
   latter class has incompatibly changed after the compilation of the
-  former class."<sup>[1](#myfootnote1)</sup> The reference can be
+  former class."<sup>[1](#myfootnote1)</sup> The dependency can be
   through a class literal, a field access, or a method invocation.
```
suztomo, line 28:
```suggestion
  symbolic reference to a class literal, a field access, or a method invocation.
```

elharo: That introduces a new term here that is not defined and not otherwise used on this page.

suztomo: @elharo I thought about the difference between "dependency" and "symbolic reference". The symbolic references in the constant pool of class files are the implementation detail of the "dependency" written in the [java.lang.LinkageError Javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/LinkageError.html).  Therefore the term "symbolic reference" does not belong to this doc.

Does this match your thought?

elharo: It's simpler than that. We're quoting here from the official JDK docs, with citation. We use the same term they use. We do not need to introduce new and unfamiliar terms to define a linkage error here.

suztomo: Okay.

PR #1615, opened 2020-08-21
clarify JavaDoc around references

elharo:
@suztomo less circular definitions


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/AbstractMethodProblem.java
```diff
@@ -17,7 +17,7 @@
 package com.google.cloud.tools.opensource.classpath;
 
 /**
- * The referenced {@code methodSymbol} is not implemented in the {@code targetClass} but the class
+ * A {@code methodSymbol} is not implemented in the {@code targetClass} but the class
```
suztomo, line 20:
```suggestion
 * A {@code methodSymbol} is not implemented in the {@code implementationClass} but the class
```

Issue #1622, opened 2020-08-27
Check kokoro scripts in google3 for JDK choice


elharo:
audit the various release scripts and kokoro configs in google3 to check whether they explicitly specify a JDK. java 8 or Java 11, this should be a deliberate choice, not an accident.

elharo:
In the first case we noticed this is set in a Python script:

cl/328599533

I don't know whether this is a pretty standard way of doing this or a one off for the protobuf team.

elharo:
cs doesn't find many examples of this. While there are some, most must be doing something else.

elharo:
Obvious question: how does this repo's kokoro+rapid builds select a JDK? I don't see it in any of our config files, here or on google3, so we're probably picking up a default somewhere.

suztomo:
Right. I don't see JDK selection https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/dependencies/kokoro/release_build.sh .

Finding non-existence of something sounds difficult.

PR #1623, opened 2020-08-27
mvn -V to show the version of Maven and JDK

suztomo:
One followup for https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1622#issuecomment-681970018 on knowing the JDK verison we are using when building our artifacts.


suztomo:
From Kokoro log, I see Kokoro MacOS and Kokoro Windows use Java 8.

The release build uses gcp_ubuntu, which does not run part of pull request check.

Issue #1621, opened 2020-08-27
Idea: batch job to detect Java8-incompatible class files


suztomo:
A daily Kokoro job to check all the Maven artifacts that appear in the BOM dashboard ("Recommended versions" section) for the problematic class references that are not compatible with Java 8.

2nd part of the job checks the latest versions of the artifacts, so that we can work with the owner to fix the issue **before** GCP Java client libraries start depending on it.

Note that since yesterday, Linkage Checker can detect such invalid references.

![image](https://user-images.githubusercontent.com/28604/91388338-156a7400-e805-11ea-8cf3-3f5bc448a018.png)

@elharo What do you think?

elharo:
Better than nothing, but even better would be a build time check on each PR, like CLIRR or animal sniffer.

suztomo:
Compared to this idea, Kokoro release script and animal sniffer installation can detect issues before a release. Good. I don't object to the Kokoro audit and animal sniffer installation. They are good. However this daily job is more effective.

Compared with Kokoro release script audit, this idea has the following advantages

- This checks libraries outside Google
- This checks libraries not using Kokoro

Compared with animal sniffer installation, this idea has the following advantages:

- This is easy to implement (one Java class + job setup) compared to installing PR checks in tens if repository.
- We can see the result in a central place.
- If a release script uses different JDK or setting than PR checks, animal sniffer cannot detect bad config as PR checks.

PR #1625, opened 2020-08-27
Java8IncompatibleReferenceCheck

suztomo:
Towards #1621 .

Output: https://gist.github.com/suztomo/60ca017b17b06ee9fbadb3c85628750b (It takes around 5 minutes to run)

```
# From the directory cloud-opensource-java/dependencies
$ mvn exec:java -Pbad-jdk-reference-check -Dexec.arguments="../boms/cloud-oss-bom/pom.xml"

...

Aug 27, 2020 6:28:44 PM com.google.cloud.tools.opensource.dependencies.BadJdkReferenceChecker main
SEVERE: The following artifacts contain bad references to classes in 'java' package, which does not work for Java 8
  com.google.protobuf:protobuf-java:jar:3.13.0
  com.google.protobuf:protobuf-java:jar:3.12.4
The following artifacts in the BOM contain the bad artifacts in their dependencies
  com.google.protobuf:protobuf-java:jar:3.13.0 due to [com.google.protobuf:protobuf-java:jar:3.13.0]
  com.google.protobuf:protobuf-java-util:jar:3.13.0 due to [com.google.protobuf:protobuf-java:jar:3.13.0]
  com.google.api.grpc:proto-google-iam-v1:jar:1.0.0 due to [com.google.protobuf:protobuf-java:jar:3.12.4]
  com.google.api.grpc:grpc-google-iam-v1:jar:1.0.0 due to [com.google.protobuf:protobuf-java:jar:3.12.4]
  com.google.cloud:google-cloud-bigquery:jar:1.116.10 due to [com.google.protobuf:protobuf-java:jar:3.12.4]
...
```


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/BadJdkReferenceChecker.java
```diff
@@ -0,0 +1,125 @@
...
+import com.google.common.collect.ImmutableSetMultimap;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.logging.Logger;
+import org.eclipse.aether.artifact.Artifact;
+
+public class BadJdkReferenceChecker {
```
elharo, line 38: we need a better name for this. What's a "bad JDK"?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/BadJdkReferenceChecker.java
```diff
@@ -0,0 +1,125 @@
...
+
+  private static final Logger logger = Logger.getLogger(BadJdkReferenceChecker.class.getName());
+
+  public static void main(String[] arguments)
+      throws MavenRepositoryException, IOException, URISyntaxException {
+
+    if (arguments.length != 1) {
+      System.err.println("Please specify a path to the BOM file");
```
elharo, line 46: per google style do not use "Please" in UI messages

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/BadJdkReferenceChecker.java
```diff
@@ -0,0 +1,125 @@
...
+
+    if (bomMemberToBadDependencies.isEmpty()) {
+      logger.info("No problematic artifacts");
+      return;
+    }
+
+    StringBuilder message = new StringBuilder();
+    message.append(
+        "The following artifacts contain bad references to classes in 'java' package,"
+            + " which does not work for Java 8\n");
```
elharo, line 112: does --> do

suztomo:
For future enhancement, let's support different Java versions than Java 8.

suztomo:
Let's not setup daily job for now.

# Any way to prevent this before a library is released?

Maven builds JAR file(s). Enforcer rule can verify the JAR file, before publishing to Maven Central. Animal sniffer?

But setting up enforcer rule / animal sniffer is less simple than "--release 8" in pom.xml.

suztomo:
Problem document and solution proposal document for check.

Type1 the JAR itself . Type 2 in transitive deps.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
```diff
@@ -35,20 +35,20 @@
 import java.util.logging.Logger;
 import org.eclipse.aether.artifact.Artifact;
 
-public class BadJdkReferenceChecker {
+public class Java8IncompatibleReferenceCheck {
```
elharo, line 38: JavaDoc would help clarify this.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
```diff
@@ -109,7 +109,7 @@ public static void main(String[] arguments)
     StringBuilder message = new StringBuilder();
     message.append(
         "The following artifacts contain bad references to classes in 'java' package,"
```
elharo, line 111: delete "bad"
perhaps "classes in 'java' package" --> classes in the core library
which do not work for --> which are not present in

suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
suztomo: Added Javadoc. I'll prepare another document (wiki) to explain how to use this tool.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.logging.Logger;
+import org.eclipse.aether.artifact.Artifact;
+
+/**
+ * A tool to find Java 8-incompatible references in the class files in a BOM. It checks the class
+ * files in the following artifacts:
+ *
+ * <ul>
+ *   <li>The artifacts in the {@code dependencyManagement} section of the BOM
+ *   <li>The artifacts in the dependency graph for each of the artifact above
```
elharo, line 44: dependency graph of each of the artifacts above

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
```diff
@@ -0,0 +1,145 @@
...
+ *   <li>The latest version of the artifacts in the dependency graphs
+ * </ul>
+ *
+ * <p>If it finds references to classes in the core library, which are not present in Java 8, it
```
elharo, line 48: no comma after library

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
```diff
@@ -0,0 +1,145 @@
...
+      System.err.println("Specify a path to the BOM file");
+      System.exit(1);
+    }
+
+    URI exclusionFileUri =
+        Java8IncompatibleReferenceCheck.class
+            .getClassLoader()
+            .getResource("java8-incompatible-reference-check-exclusion.xml")
+            .toURI();
+    Path exclusionFile = Paths.get(exclusionFileUri).toAbsolutePath();
```
elharo, line 73: I'm surprised this works since it's a resource in a jar, not a file. Could we skip both URIs and paths here and get a stream from the resource instead?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
```diff
@@ -0,0 +1,145 @@
...
+
+      ImmutableSet<LinkageProblem> linkageProblems = linkageChecker.findLinkageProblems();
+
+      ImmutableSet<LinkageProblem> invalidJdkReferences =
+          linkageProblems.stream()
+              .filter(problem -> problem.getSymbol().getClassBinaryName().startsWith("java."))
+              .collect(toImmutableSet());
+
+      if (!invalidJdkReferences.isEmpty()) {
+        invalidJdkReferences.stream()
```
elharo, line 111: This is what would contain the return type change?

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
suztomo: Fixed.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
suztomo: Fixed.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
suztomo: This works because it reads the XML file from `target` directory. With maven-exec-plugin (used by java8-incompatible-reference-check.sh), the value (toString) of `exclusionFile` is `Users/suztomo/cloud-opensource-java/dependencies/target/classes/java8-incompatible-reference-check-exclusion.xml`, not from a JAR file.

`LinkageChecker.create()` takes `Path` for an exclusion file.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
suztomo: Yes. Renamed this to `invalidReferencesToJavaCoreLibrary`.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
elharo: I'll trust you that it works, but this still looks very brittle to me. Can we add a method to LinkageChecker that takes an InputStream to avoid this.

suztomo: I tried adding another factory method for LinkageChecker to take URL (not InputStream, because ExcludedErrors already takes URL) but the change is unexpectedly large https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1625/commits/01b6dc4023dec3417224af53fdc530d96c0f63e7

I will just use Path for the exclusion xml as second parameter, without URL or InputStream.

elharo (approved):

dependencies/pom.xml
```diff
@@ -120,4 +120,26 @@
...
+        <plugins>
+          <plugin>
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>exec-maven-plugin</artifactId>
+            <configuration>
+              <skip>false</skip>
+              <mainClass>com.google.cloud.tools.opensource.dependencies.Java8IncompatibleReferenceCheck</mainClass>
+              <arguments>
+                <argument>../boms/cloud-oss-bom/pom.xml</argument>
+                <argument>src/main/resources/java8-incompatible-reference-check-exclusion.xml</argument>
```
suztomo, line 137: Now the exclusion file is passed to the program as the second argument.

dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/Java8IncompatibleReferenceCheck.java
suztomo: Now the exclusion file is passed as the second parameter. No need to deal with URL.

PR #1647, opened 2020-09-11
clean up code

elharo:


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerMain.java
```diff
@@ -148,10 +150,13 @@ public static void main(String[] arguments)
               .collect(toImmutableSet());
     }
 
+    // TODO this should be a separate method; not part of findLinkageProblems
     Path outputExclusionFile = linkageCheckerArguments.getOutputExclusionFile();
     if (outputExclusionFile != null) {
       ExclusionFiles.write(outputExclusionFile, linkageProblems);
       System.out.println("Wrote the linkage errors as exclusion file: " + outputExclusionFile);
+      
+      // TODO why do we return an empty set here?
```
suztomo, line 159: Because when writing the linkage problems into the file, we don't want to print the problems and the operation should not throw LinkageCheckResultException

suztomo (approved):

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerMain.java
elharo: OK, that makes sense. Splitting out logically distinct operations (printing linkage errors, writing exclusion file, etc.) into separate methods will probably help clarify this.

PR #1649, opened 2020-09-14
The Gradle plugin to use linkage checker 1.5.2

suztomo:


gradle-plugin/src/main/java/com/google/cloud/tools/dependencies/gradle/LinkageCheckTask.java
```diff
@@ -109,7 +110,7 @@ private boolean findLinkageErrors(Configuration configuration) throws IOExceptio
 
     ClassPathResult classPathResult =
         createClassPathResult(configuration.getResolvedConfiguration());
-    ImmutableList.Builder<ClassPathEntry> classPathBuilder = ImmutableList.builder();
+    ImmutableList.Builder<ClassPathEntry> classPathEntryBuilder = ImmutableList.builder();
```
elharo, line 113: classPahBuilder was correct. ClassPathEntry is the element type, not the list type.

suztomo: Good point. Updated to `classPathEntriesBuilder`.

gradle-plugin/src/main/java/com/google/cloud/tools/dependencies/gradle/LinkageCheckTask.java
```diff
@@ -142,7 +143,8 @@ private boolean findLinkageErrors(Configuration configuration) throws IOExceptio
 
       ImmutableSet<LinkageProblem> linkageProblems = linkageChecker.findLinkageProblems();
 
-      LinkageProblemCauseAnnotator.annotate(classPathResult, linkageProblems);
+      ClassPathBuilder classPathBuilder = new ClassPathBuilder();
```
suztomo, line 146: Because this is classPathBuilder, I renamed the existing `classPathBuilder` to something else.

elharo (approved):

PR #1653, opened 2020-09-15
The Gradle plugin to use dependencies module through Maven local

### LINKED ISSUE #1651: Gradle plugin in master to depend on the Linkage Checker master
As of now, a change to the LinkageChecker class is not immediately used in the Gradle plugin, because the Gradle plugin uses a fixed version of the "com.google.cloud.tools:dependencies" artifact. This disconnection causes pain in:
- the increased release procedure for the Gradle plugin
- the late validation of a LinkageChecker's change against the Gradle plugin

## Current Release Procedure
Everytime we change LinkageChecker in the dependencies module, we need to do the following

- release the dependencies module as part of https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/RELEASING.md
- wait until the latest version of the dependencies module appear in Maven Central
- update Gradle plugin to use the latest version ([example pull request](https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1649))
  - This may involve code change in the Gradle plugin. This change could happen before the dependencies module if the Gradle plugin in master branch relies on the LinkageChecker in master.
  - **If this upgrade reveals a LinkageChecker's bug for Gradle plugin, we have to go back to the change that caused the bug.**
- release the Gradle plugin https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/gradle-plugin/RELEASING.md ([example pull request](https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1652]))
  This uses `vX.Y.Z-gradle` release tag. This version is independent from the Maven enforcer rule's version. 

# Proposal

## The Enforcer Rule in Maven and the Gradle plugin in Gradle

This is an enhancement of current setup of mixing Gradle and Maven and this proposes to enhance the release script "scripts/prepare_release.sh" so that:

- the Gradle plugin in master branch relies on the LinkageChecker in master. (version `latest.integration`)
- When the release script creates a release tag, it sets the dependencies module version in `build.gradle` to the releasing version. Suppose we use `v1.6.0-dependencies`.
- We release the Linkage Checker enforcer rule and the Gradle plugin using the tag `v1.6.0-dependencies`. No need to wait for Maven Central. As the result, the enforcer rule and the Gradle plugin will have the same version for one release. 
  (We no longer use the `vX.Y.Z-gradle` release tag suffix.)

## The Gradle plugin to have java.sourceSet to include ../dependencies/src/main/java

This would have duplicate 

### Everything in Maven

https://stackoverflow.com/questions/27555560/how-can-i-build-a-gradle-plugin-with-maven . Spring Boot's Gradle plugin was developed in Maven. But now they develop it in Gradle.

The stackoverflow explains https://repo.gradle.org/gradle/libs-releases-local but the latest of https://repo.gradle.org/gradle/libs-releases-local/org/gradle/gradle-core-api/ is 6.1.1. (The latest Gradle version is 6.6). Yet, if some artifacts are missing in Maven Central but available somewhere (as part of Gradle distribution), we can create a Maven local ("libs" directory) repository that holds the artifact in this cloud-opensource-java repository just to build our Gradle plugin. Alternative to the lib directory, we can use  the maven-install-plugin to move them to he local repository.

If we don't use Gradle's "com.gradle.plugin-publish" plugin, we'll need to figure out how to upload artifacts to Gradle's plugin repository. (If the Gradle repository changes how it works, we're responsible to adjust it.)

### Everything in Gradle

Jib project is all Gradle. They develop [Jib Maven plugin](https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugine) in Gradle.

We have settings done in Maven:
- The Maven invoker plugin for the enforcer rule integration tests.
- The release script uses `mvn versions:set` to apply changes to pom.xml at once.

It's feasible but this effort does not sound good just for this purpose to align the Gradle enforcer rule and Maven enfrocer rule.

suztomo:
Starts on #1651


elharo (commented):
Checked out the branch and tried to build. This is what I got:

Starting a Gradle Daemon (subsequent builds will be faster)
> Task :compileJava FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileJava'.
> Could not resolve all files for configuration ':compileClasspath'.
   > Could not find com.google.cloud.tools:dependencies:1.5.3-SNAPSHOT.
     Searched in the following locations:
       - file:/home/elharo/.m2/repository/com/google/cloud/tools/dependencies/1.5.3-SNAPSHOT/maven-metadata.xml
       - file:/home/elharo/.m2/repository/com/google/cloud/tools/dependencies/1.5.3-SNAPSHOT/dependencies-1.5.3-SNAPSHOT.pom
       - https://jcenter.bintray.com/com/google/cloud/tools/dependencies/1.5.3-SNAPSHOT/maven-metadata.xml
       - https://jcenter.bintray.com/com/google/cloud/tools/dependencies/1.5.3-SNAPSHOT/dependencies-1.5.3-SNAPSHOT.pom
     Required by:
         project :

suztomo:
"mvn install" needs to run before the Gradle project. I'll check why Circle CI failed.

elharo (commented):
so this is reading from the local repo instead of Maven central which is a step forward, but it's still not fully building from the source tree itself.

suztomo:
> but it's still not fully building from the source tree itself.

That's a good point. Let me find a way to specify Java source directory as part of gradle-plugin module rather than declaring dependency.

suztomo:
I'll add documentation on the Gradle module.

gradle-plugin/README.md
```diff
@@ -4,7 +4,14 @@ For usage of this plugin, see the documentation.
 
 # Build Instruction
 
+This Gradle project depends on the `dependencies` module through the Maven local repository.
+
 ```
+# At the root of the cloud-opensource-java project, first install the Maven projects to Maven local
```
elharo, line 10: the local Maven repository

gradle-plugin/README.md
```diff
@@ -4,7 +4,14 @@ For usage of this plugin, see the documentation.
...
+# repository.
+$ mvn install
```
elharo, line 12: indent four spaces for code

suztomo: Done with 3-space indentation (as in boms/cloud-oss-bom/RELEASING.md; 4-space didn't work in GitHub's markdown viewer.)

gradle-plugin/README.md
suztomo: Updated

suztomo (commented):
@elharo PTAL.

I created an example pull request using the updated script https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1654. It's working good.

gradle-plugin/README.md
```diff
@@ -4,24 +4,35 @@ For usage of this plugin, see the documentation.
...
-suztomo-macbookpro44% ls ~/.m2/repository/com/google/cloud/tools/linkage-checker-gradle-plugin/0.1.0-SNAPSHOT/
-linkage-checker-gradle-plugin-0.1.0-SNAPSHOT.jar
-linkage-checker-gradle-plugin-0.1.0-SNAPSHOT.pom
-```
+   ```
+   $ cd gradle-plugin
+   $ ./gradlew build publishToMavenLocal
+   ```
+
+This command installs the Linkage Checker Gradle plugin in Maven local repository.
```
elharo, line 23: the local Maven

elharo (approved):

gradle-plugin/README.md
suztomo: Updated.

PR #1659, opened 2020-09-17
don't throw exception for user input error

elharo:
@suztomo


dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -115,9 +115,11 @@ private static void generateAllVersions(String versionlessCoordinates)
       throws IOException, TemplateException, RepositoryException, URISyntaxException,
       MavenRepositoryException {
     List<String> elements = Splitter.on(':').splitToList(versionlessCoordinates);
-    checkArgument(
-        elements.size() == 2,
-        "The versionless coordinates should have one colon character: " + versionlessCoordinates);
+    if ( elements.size() != 2 ) {
+      System.err.println(
+          "Versionless coordinates should have one colon: " + versionlessCoordinates);
+      return;
```
suztomo, line 121: This unexpectedly returns success status to the caller of this tool. For example maven-exec-plugin would show "success" if we pass wrong argument to DashboardMain.

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -115,9 +115,11 @@ private static void generateAllVersions(String versionlessCoordinates)
       throws IOException, TemplateException, RepositoryException, URISyntaxException,
       MavenRepositoryException {
     List<String> elements = Splitter.on(':').splitToList(versionlessCoordinates);
-    checkArgument(
-        elements.size() == 2,
-        "The versionless coordinates should have one colon character: " + versionlessCoordinates);
+    if ( elements.size() != 2 ) {
```
suztomo, line 118:
```suggestion
    if (elements.size() != 2) {
```

suztomo:
@elharo Can you update the style part and the return code part? Then let's merge this PR.

PR #1663, opened 2020-09-22
RepositoryUtility.findHighestVersion to work with non-semantic version

suztomo:
RepositoryUtility.findHighestVersion to work with non-semantic version such as "v2-rev20200818-1.30.10".

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
@@ -300,7 +300,7 @@ public final class RepositoryUtility {
       String artifactId)
       throws MavenRepositoryException {
 
-    Artifact artifactWithVersionRange = new DefaultArtifact(groupId, artifactId, null, "(0,]");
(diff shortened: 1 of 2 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/dependencies/RepositoryUtility.java
```diff
@@ -300,7 +300,7 @@ private static VersionRangeResult findVersionRange(
       String artifactId)
       throws MavenRepositoryException {
 
-    Artifact artifactWithVersionRange = new DefaultArtifact(groupId, artifactId, null, "(0,]");
+    Artifact artifactWithVersionRange = new DefaultArtifact(groupId, artifactId, null, "(,]");
```
suztomo, line 303: As per https://cwiki.apache.org/confluence/display/MAVENOLD/Dependency+Mediation+and+Conflict+Resolution#DependencyMediationandConflictResolution-DependencyVersionRanges, both min and max are optional. We didn't have to specify the minimum limit.

elharo (approved):

PR #1676, opened 2020-09-23
Suppress InaccessibleClassProblem if the class reference is unused in the code attribute

### LINKED ISSUE #1608: Private subclasses of public classes should not be reported as linkage errors
Followup of https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1599#issuecomment-675056070. It was a false positive.

`com.sun.tools.internal.ws.wscompile.WsgenOptions` (source) references `com.sun.xml.internal.ws.api.BindingID$SOAPHTTPImpl` (target). The latter is a private class. Linkage Checker detected it as a linkage error. However, the reference turned out to be not harmful. This is because the source class is not instantiating the private class or declaring a variable with the type.

One of the referencing method is the code below:

```
// in WsgenOptions.java
  BindingID getBindingID(String protocol) {
    if (protocol.equals("soap1.1")) {
      return BindingID.SOAP11_HTTP;
    } else if (protocol.equals("Xsoap1.2")) {
      return BindingID.SOAP12_HTTP;
    } else {
      String lexical = (String)this.nonstdProtocols.get(protocol);
      return lexical != null ? BindingID.parse(lexical) : null;
    }
  }
```

Here, `BindingID.SOAP11_HTTP` has type `com.sun.xml.internal.ws.api.BindingID$SOAPHTTPImpl` (private class), which is a subclass of `BindingID` class. As WsgenOptions is not touching the private class, calling the `getBindingID` method does not cause a `LinkageError` at runtime.

# Simplified Example

Suppose we have the following class `A`, `B`, and `C`:

```
package foo;

public class A {
  public static final B f = new B();

  private static class B extends A {
  }
}
```

```
package foo;

public class C {
  public static void main(String[] arguments) {
    A a = A.f;
    System.out.println(a);
  }

  void f(A.B b) { // This should fail. How?
    System.out.println(b);
  }
}
```

When compiled, `C`'s class file contains a reference to (private) `A$B` class. Linkage Checker would report this reference as a linkage error, simply because `A$B` is not accessible from `C`. However, this is a false positive as the code compiles and run without a problem. 

```
  Compiled from "C.java"
Constant pool:
...
   #3 = Fieldref           #23.#24        // foo/A.f:Lfoo/A$B;
...
  #23 = Class              #32            // foo/A
  #24 = NameAndType        #33:#37        // f:Lfoo/A$B;
...
  #32 = Utf8               foo/A
  #33 = Utf8               f
  #34 = Class              #41            // foo/A$B
  #37 = Utf8               Lfoo/A$B;
...
  #41 = Utf8               foo/A$B
```

Interestingly the field `#34` is not referenced by other constant pool entries or byte code. How about the case of `com.sun.tools.internal.ws.wscompile.WsgenOptions`?

`com.sun.tools.internal.ws.wscompile.WsgenOptions` has the class reference  ([javap output](https://gist.github.com/suztomo/482d33f3f7c56f99ad19cd91cbc4470d)) used in InnerClasses attribute:

```
#392 = Class              #408          // com/sun/xml/internal/ws/api/BindingID$SOAPHTTPImpl
...
InnerClasses:
     private static final #393= #392 of #300; //SOAPHTTPImpl=class com/sun/xml/internal/ws/api/BindingID$SOAPHTTPImpl of class com/sun/xml/internal/ws/api/BindingID
```

https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.7.6

> If the constant pool of a class or interface C contains at least one CONSTANT_Class_info entry (§4.4.1) which represents a class or interface that is not a member of a package, then there must be exactly one InnerClasses attribute in the attributes table of the ClassFile structure for C.

How about "If a class reference is only used by the InnerClasses annotation, we skip it"?

# Java Language Specification: 12.3.3. Resolution of Symbolic References

https://docs.oracle.com/javase/specs/jls/se7/html/jls-12.html#jls-12.3.3

For constructor, we can check`new`. But it cannot detect wrong access of `A.B`

How about "If a class reference is only used by NameAndType, we ignore it"?

suztomo:
Fixes #1608.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -561,6 +561,10 @@ class ClassDumper {
     return null;
   }
 
+  boolean isClassSymbolReferenceUsed(String sourceClassName, ClassSymbol classSymbol) {
+    return !isUnusedClassSymbolReference(sourceClassName, classSymbol);
+  }
+
   /**
    * Returns true if the class symbol reference is unused in the source class file. It checks
    * following places for the usage in the source class:
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -656,7 +656,7 @@ class ClassDumper {
             Instruction instruction = instructionHandle.getInstruction();
             if (instruction instanceof CPInstruction) {
               // Checking JVM instructions that take a symbolic reference to a class in
-              // JVM Instruction Set: anewarray, checkcast, instanceof, ldc, ldc2_w, multianewarray,
+              // JVM Instruction Set: anewarray, checkcast, instanceof, ldc, ldc_w, multianewarray,
               // and new.
               // https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-6.html#jvms-6.5
               int classIndex = ((CPInstruction) instruction).getIndex();
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -656,7 +656,8 @@ class ClassDumper {
             Instruction instruction = instructionHandle.getInstruction();
             if (instruction instanceof CPInstruction) {
               // Checking JVM instructions that take a symbolic reference to a class in
-              // JVM Instruction Set
+              // JVM Instruction Set: anewarray, checkcast, instanceof, ldc, ldc2_w, multianewarray,
+              // and new.
               // https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-6.html#jvms-6.5
               int classIndex = ((CPInstruction) instruction).getIndex();
               if (targetConstantPoolIndices.contains(classIndex)) {
@@ -699,7 +700,8 @@ class ClassDumper {
           "The source class in the reference is no longer available in the class path", ex);
     }
 
-    // The target class is unused
+    // The target class is unused in the source class. For example a class reference in the constant
+    // pool is used only in its InnerClasses attribute.
     return true;
   }
 
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
@@ -561,12 +561,8 @@ class ClassDumper {
     return null;
   }
 
-  boolean isClassSymbolReferenceUsed(String sourceClassName, ClassSymbol classSymbol) {
-    return !isUnusedClassSymbolReference(sourceClassName, classSymbol);
-  }
-
   /**
-   * Returns true if the class symbol reference is unused in the source class file. It checks
+   * Returns true if the class symbol reference is used in the source class file. It checks
    * following places for the usage in the source class:
    *
    * <ul>
@@ -585,10 +581,10 @@ class ClassDumper {
    * @see <a href="https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-2.html#jvms-2.10">Java
    *     Virtual Machine Specification: Exceptions</a>
    */
-  boolean isUnusedClassSymbolReference(String sourceClassName, ClassSymbol classSymbol) {
+  boolean isClassSymbolReferenceUsed(String sourceClassName, ClassSymbol classSymbol) {
     if (classSymbol instanceof SuperClassSymbol) {
       // The target class is used in class inheritance
-      return false;
+      return true;
     }
 
     String targetClassName = classSymbol.getClassBinaryName();
@@ -599,7 +595,7 @@ class ClassDumper {
       for (String interfaceName: sourceJavaClass.getInterfaceNames()) {
         if (interfaceName.equals(targetClassName)) {
           // The target class is used in interfaces
-          return false;
+          return true;
         }
       }
 
@@ -625,7 +621,7 @@ class ClassDumper {
               int classIndex = constantCp.getClassIndex();
               if (targetConstantPoolIndices.contains(classIndex)) {
                 // The class reference is used in another constant pool
-                return false;
+                return true;
               }
               break;
           }
@@ -636,7 +632,7 @@ class ClassDumper {
         // Type.toString returns binary name (for example, io.grpc.MethodDescriptor)
         String fieldTypeSignature = field.getType().toString();
         if (targetClassName.equals(fieldTypeSignature)) {
-          return false;
+          return true;
         }
       }
 
@@ -644,12 +640,12 @@ class ClassDumper {
       for (Method method : sourceJavaClass.getMethods()) {
 
         if (targetClassName.equals(method.getReturnType().toString())) {
-          return false;
+          return true;
         }
         for (Type argumentType : method.getArgumentTypes()) {
           String argumentTypeSignature = argumentType.toString();
           if (targetClassName.equals(argumentTypeSignature)) {
-            return false;
+            return true;
           }
         }
 
@@ -666,7 +662,7 @@ class ClassDumper {
               int classIndex = ((CPInstruction) instruction).getIndex();
               if (targetConstantPoolIndices.contains(classIndex)) {
                 // The target class is used in a JVM instruction (including `new`).
-                return false;
+                return true;
               }
             }
           }
@@ -679,7 +675,7 @@ class ClassDumper {
           for (int exceptionIndexTableEntry : exceptionIndexTable) {
             if (targetConstantPoolIndices.contains(exceptionIndexTableEntry)) {
               // The target class is used in throws clause
-              return false;
+              return true;
             }
           }
         }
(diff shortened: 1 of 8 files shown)


elharo (approved):

elharo (changes requested):
tests seem to fail

suztomo:
```
2 tests failed
  out of 260
testFindLinkageProblems_shouldStripSourceInnerClasses - com.google.cloud.tools.opensource.classpath.LinkageCheckerTest
org.apache.bcel.classfile.ClassFormatException: The source class in the reference is no longer available in the class path
	at com.google.cloud.tools.opensource.classpath.LinkageCheckerTest.testFindLinkageProblems_shouldStripSourceInnerClasses(LinkageCheckerTest.java:599)
Caused by: java.lang.ClassNotFoundException: Exception while looking for class com.google.foo.Bar$Baz: java.io.IOException: Couldn't find: com/google/foo/Bar$Baz.class
	at com.google.cloud.tools.opensource.classpath.LinkageCheckerTest.testFindLinkageProblems_shouldStripSourceInnerClasses(LinkageCheckerTest.java:599)
Caused by: java.io.IOException: Couldn't find: com/google/foo/Bar$Baz.class
	at com.google.cloud.tools.opensource.classpath.LinkageCheckerTest.testFindLinkageProblems_shouldStripSourceInnerClasses(LinkageCheckerTest.java:599)
testFindClassReferences_privateClass - com.google.cloud.tools.opensource.classpath.LinkageCheckerTest
org.apache.bcel.classfile.ClassFormatException: The source class in the reference is no longer available in the class path
	at com.google.cloud.tools.opensource.classpath.LinkageCheckerTest.testFindClassReferences_privateClass(LinkageCheckerTest.java:569)
Caused by: java.lang.ClassNotFoundException: Exception while looking for class com.google.cloud.tools.opensource.classpath.LinkageCheckerTest: java.io.IOException: Couldn't find: com/google/cloud/tools/opensource/classpath/LinkageCheckerTest.class
	at com.google.cloud.tools.opensource.classpath.LinkageCheckerTest.testFindClassReferences_privateClass(LinkageCheckerTest.java:569)
Caused by: java.io.IOException: Couldn't find: com/google/cloud/tools/opensource/classpath/LinkageCheckerTest.class
	at com.google.cloud.tools.opensource.classpath.LinkageCheckerTest.testFindClassReferences_privateClass(LinkageCheckerTest.java:569)
```

I'll need to find a good test case of InaccessibleClassProblem

suztomo:
I tried to find an artifact that contains a class that became private from public, and another artifact containing a class that referenced the class. No luck.

Maybe I'll commit a JAR file that contains such bad class files and add instruction to generate one.

suztomo:
@elharo PTAL. I updated the tests.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -656,7 +656,8 @@ boolean isUnusedClassSymbolReference(String sourceClassName, ClassSymbol classSy
             Instruction instruction = instructionHandle.getInstruction();
             if (instruction instanceof CPInstruction) {
```
suztomo, line 657: @netdpb FYI, BCEL has an abstract class "CPInstruction" that represents an instruction that takes a constant pool index, and we already had this function to determine whether a constant pool index is unused by JVM instructions. (Therefore this function does not list "safe" operations)

elharo (approved):

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
netdpb: That's fine, as long as it doesn't list "dangerous" operations either.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -656,7 +656,8 @@ boolean isUnusedClassSymbolReference(String sourceClassName, ClassSymbol classSy
...
               // Checking JVM instructions that take a symbolic reference to a class in
-              // JVM Instruction Set
+              // JVM Instruction Set: anewarray, checkcast, instanceof, ldc, ldc2_w, multianewarray,
+              // and new.
```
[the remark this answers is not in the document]
netdpb: I think you mean ldc_w, not ldc2_w. ldc2_w is for longs or doubles.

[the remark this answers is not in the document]
suztomo: Thanks. Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -512,7 +512,13 @@ private boolean isMemberAccessibleFrom(
       }
 
       if (!isClassAccessibleFrom(targetClass, sourceClassName)) {
-        return Optional.of(new InaccessibleClassProblem(sourceClassFile, targetClassFile, symbol));
+        if (classDumper.isUnusedClassSymbolReference(sourceClassName, symbol)) {
```
netdpb, line 515: What about:

```java
if (!isClassAccessibleFrom(targetClass, sourceClassName)
    && !classDumper.isUnusedClassSymbolReference(sourceClassName, symbol)) {
  return Optional.of(new InaccessibleClassProblem(sourceClassFile, targetClassFile, symbol));
}
return Optional.empty();
```

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerTest.java
```diff
@@ -548,53 +551,56 @@ public void testFindClassReferences_innerClass() throws IOException {
...
-    // The superclass of AbstractApiService$InnerService (Guava's ApiService) is not in the paths
-    ClassPathEntry dummySource = firestoreJar;
-    List<ClassPathEntry> paths =
-        ImmutableList.of(classPathEntryOfResource("testdata/api-common-1.7.0.jar"));
-    LinkageChecker linkageChecker = LinkageChecker.create(paths);
+  public void testFindClassReferences_inaccessibleClass() throws IOException, URISyntaxException {
+    // io.grpc.grpclb.GrpclbLoadBalancer in grpc-grpclb 0.12.0 had a reference to
+    // io.grpc.internal.SingleTransportChannel. The SingleTransportChannel became non-public in
+    // grpc-core 0.15.0.
+    ClassPathBuilder classPathBuilder = new ClassPathBuilder();
```
netdpb, line 559: You only use this once. Inline?

netdpb (approved):

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
suztomo: I tried that and I was confused. Maybe this is more clear:

```
if (isClassAccessibleFrom(targetClass, sourceClassName)
    || classDumper.isUnusedClassSymbolReference(sourceClassName, symbol)) {
  return Optional.empty();
} else {
  return Optional.of(new InaccessibleClassProblem(sourceClassFile, targetClassFile, symbol));
}
```

netdpb: Maybe, but it does seem like an inversion of the test.

If the class is inaccessible and it's not unused... What if you inverted `isUnusedClassSymbolReference`?

```java
if (!isClassAccessibleFrom(targetClass, sourceClassname)
    && classDumpter.isClassSymbolReferenceUsed(sourceClassName, symbol)) {
  return Optional.of(…);
}
return Optional.empty();
```

suztomo: Added isClassSymbolReferenceUsed to ClassDumper.

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerTest.java
suztomo: Updated.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -561,6 +561,10 @@ private static String outerClassName(JavaClass sourceJavaClass) {
     return null;
   }
 
+  boolean isClassSymbolReferenceUsed(String sourceClassName, ClassSymbol classSymbol) {
+    return !isUnusedClassSymbolReference(sourceClassName, classSymbol);
```
netdpb, line 565: Oh. Do we need both? I think positive method names are usually better than negative ones. Where else is `isUnusedClassSymbolReference` used, and can it be switched?

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java
```diff
@@ -561,12 +561,8 @@ private static String outerClassName(JavaClass sourceJavaClass) {
     return null;
   }
 
-  boolean isClassSymbolReferenceUsed(String sourceClassName, ClassSymbol classSymbol) {
-    return !isUnusedClassSymbolReference(sourceClassName, classSymbol);
-  }
-
   /**
-   * Returns true if the class symbol reference is unused in the source class file. It checks
+   * Returns true if the class symbol reference is used in the source class file. It checks
```
elharo, line 565: checks the following

suztomo: Fixed.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ClassDumper.java

>  positive method names are usually better than negative ones

Yes, especially when we negate the result.

PR #1689, opened 2020-09-30
Method handle

### LINKED ISSUE #1684: MethodHandle's invoke method reference is always marked as a linkage error
Linkage Checker generates false positive errors for `java.lang.invoke.MethodHandle`'s invoke method.

```java
package org.example;

import java.lang.invoke.MethodHandle;
import java.lang.invoke.MethodHandles;
import java.lang.invoke.MethodType;

public class App 
{
    static MethodType mt = MethodType.methodType(String.class, String.class);

    public static void main( String[] args ) throws Throwable {

        MethodHandles.Lookup lookup = MethodHandles.lookup();
        MethodHandle mh = lookup.findStatic(App.class, "foo", mt);

        mh.invoke("hello");
    }

    public static String foo(String param) {
        System.out.println(param + " foo");
        return "foo";
    }
}
```

The program runs without a runtime error. However, Linkage Checker reports false positive error:

```
java.lang.invoke.MethodHandle's method invoke(String) is not found;
  referenced by 1 class file
    org.example.App (org.example:method-handle-invoke:1.0-SNAPSHOT)
  Cause:
    Unknown
Problematic artifacts in the dependency tree:
org.example:method-handle-invoke:1.0-SNAPSHOT is at:
  org.example:method-handle-invoke:jar:1.0-SNAPSHOT
```

Indeed, the `invoke` method's argument is an array of `Object`s ([javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/MethodHandle.html#invoke-java.lang.Object...-
)). The class file of the program above contains the reference to 

```
$ javap -verbose -cp target/classes org.example.App
...
Constant pool:
...
   #8 = Methodref          #55.#56        // java/lang/invoke/MethodHandle.invoke:(Ljava/lang/String;)V
...
```

So Linkage Checker is working as designed in that it verified the expected argument in the constant pool (`String`) is not the one available in the class (`Object[]`).

It seems that JVM has special handling for the `invoke` method look up in that it does not throw NoSuchMethodError even when the argument type is different.

# JVM Specification

The Java 8's JVM specification says the following special handling of `MethodHandle.invoke` and related methods.

> The Java Virtual Machine gives special treatment to signature polymorphic methods in the invokevirtual instruction (§invokevirtual), in order to effect invocation of a method handle. A method handle is a strongly typed, directly executable reference to an underlying method, constructor, field, or similar low-level operation (§5.4.3.5), with optional transformations of arguments or return values. These transformations are quite general, and include such patterns as conversion, insertion, deletion, and substitution. See the java.lang.invoke package in the Java SE platform API for more information.

[2.9. Special Methods](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-2.html#jvms-2.9)

> First, a reference to an instance of java.lang.invoke.MethodType is obtained as if by resolution of a symbolic reference to a method type (§5.4.3.5) with the same parameter and return types as the descriptor of the method referenced by the invokevirtual instruction.
> 
>-  If the named method is invokeExact, the instance of java.lang.invoke.MethodType must be semantically equal to the type descriptor of the receiving method handle objectref. The method handle to be invoked is objectref.
>- If the named method is invoke, and the instance of java.lang.invoke.MethodType is semantically equal to the type descriptor of the receiving method handle objectref, then the method handle to be invoked is objectref.
>-  If the named method is invoke, and the instance of java.lang.invoke.MethodType is not semantically equal to the type descriptor of the receiving method handle objectref, then the Java Virtual Machine attempts to adjust the type descriptor of the receiving method handle, as if by a call to java.lang.invoke.MethodHandle.asType, to obtain an exactly invokable method handle m. The method handle to be invoked is m.

[Chapter 6. The Java Virtual Machine Instruction Set: invokevirtual](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-6.html#jvms-6.5.invokevirtual)

# Exclusion Rules with MethodHandle

- Appengine-api-sdk uses MethodHandle :https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/dependencies/src/main/resources/java8-incompatible-reference-check-exclusion.xml
- The default exclusion rule had the MethodHandle used in GraalVM: https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/dependencies/src/main/resources/linkage-checker-exclusion-default.xml#L46

suztomo:
Fixes #1684 .


elharo (commented):
tests are failing

suztomo:
@elharo PTAL.

dependencies/pom.xml
```diff
@@ -134,7 +134,6 @@
               <mainClass>com.google.cloud.tools.opensource.dependencies.Java8IncompatibleReferenceCheck</mainClass>
               <arguments>
                 <argument>../boms/cloud-oss-bom/pom.xml</argument>
-                <argument>src/main/resources/java8-incompatible-reference-check-exclusion.xml</argument>
```
suztomo, line 137: This file contained only the linkage errors on MethodHandle.

dependencies/pom.xml
```diff
@@ -134,7 +134,6 @@
               <mainClass>com.google.cloud.tools.opensource.dependencies.Java8IncompatibleReferenceCheck</mainClass>
```
suztomo, line 134: I'll remove this class once protobuf-java fixes the issue.
https://github.com/protocolbuffers/protobuf/issues/7827

enforcer-rules/src/it/bom-project-using-spring-repository/verify.groovy
```diff
@@ -12,4 +12,4 @@ assert !buildLog.text.contains(
 
 assert !buildLog.text.contains("NullPointerException")
 
-assert buildLog.text.contains("Linkage Checker rule found 854 errors:")
\ No newline at end of file
+assert buildLog.text.contains("Linkage Checker rule found 850 errors:")
```
suztomo, line 15: MethodHandle linkage errors are not reported.

elharo (approved):

PR #1693, opened 2020-10-02
Google-cloud-bom 0.139.0

suztomo:


suztomo:
Compared to the last released version, it's strange that gax's element has `<classifier>testlib<classifier>`:

```
       <dependency>
         <groupId>com.google.api</groupId>
         <artifactId>gax</artifactId>
-        <version>1.58.2</version>
+        <version>1.59.0</version>
         <classifier>testlib</classifier>
       </dependency>
```

https://gist.github.com/suztomo/c7b1562255a9fabb6772bd2c5b585a98

suztomo:
I see that gax-bom has included the variable of `<classifier>testlib<classifier>` for these artifacts since the beginning ([commit](https://github.com/googleapis/gax-java/commit/c0c847598eebac400e5510a43c6ae6ef3f2f5e4e#diff-dd687a733c81dae5b25ca72b9a7db379R42)).

```
       <dependency>
         <groupId>com.google.api</groupId>
         <artifactId>gax</artifactId>
-        <version>1.58.2</version>
+        <version>1.59.0</version>
       </dependency>
       <dependency>
         <groupId>com.google.api</groupId>
         <artifactId>gax</artifactId>
-        <version>1.58.2</version>
+        <version>1.59.0</version>
         <classifier>testlib</classifier>
       </dependency>
```

This should not be a problem for google-cloud-bom 0.139.0.

suztomo:
As per the diff https://gist.github.com/suztomo/c7b1562255a9fabb6772bd2c5b585a98, the only differences are the versions of these libraries and the order of the dependencies elements. This is expected.

elharo (commented):
I want to understand the GAX classifier issue a little better before we move forward with this.

elharo:
You're right that we've had this for a while, but it's always been weird. See:

https://github.com/googleapis/gax-java/blob/master/gax/src/test/java/com/google/api/gax/rpc/testing/FakePagedApi.java

Notice that this comes from src/test, not src/main.

I'm not sure anyone should depend on this, or that we should ship it.

boms/cloud-oss-bom/pom.xml
```diff
@@ -45,7 +45,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <guava.version>29.0-android</guava.version>
-    <google.cloud.java.version>0.137.0</google.cloud.java.version>
+    <google.cloud.java.version>0.138.0</google.cloud.java.version>
```
elharo, line 48: why is this 138 instead of 139?

suztomo: Good catch. Fixed.

suztomo:
Updated the diff of the effective pom https://gist.github.com/suztomo/c7b1562255a9fabb6772bd2c5b585a98 The difference is the version number and the order of dependency element. No removal or addition of artifacts.

suztomo:
Updated the diff. https://gist.github.com/suztomo/c7b1562255a9fabb6772bd2c5b585a98 . I see gax and gax-grpc testlib have been removed.

elharo (approved):

elharo (commented):
gax shouldn't have been removed. Are you sure?

suztomo:
I meant gax testlib-classifier and gax-grpc testlib-classifier have been removed.

PR #1698, opened 2020-10-05
Release 12.0.0-bom

suztomo:
Release 12.0.0-bom


elharo (changes requested):
not yet

elharo (approved):
OK, we can move ahead with this

PR #1704, opened 2020-10-13
update to google-cloud-bom 0.140.0

elharo:
@suztomo


suztomo:
The error contains the old version "12.0.1-SNAPSHOT".

I guess this PR needs to update `./upper-bounds-check/pom.xml` to have the new version "12.1.0-SNAPSHOT".

PR #1736, opened 2020-10-19
prefer release_bom.sh

elharo:
@suztomo There are still some improvements to be made, but I've now run the new process end-to-end multiple times. It is much quicker and easier than clicking through multiple Rapid screens.

### DIFF
--- a/boms/cloud-oss-bom/RELEASING.md
+++ b/boms/cloud-oss-bom/RELEASING.md
@@ -14,10 +14,15 @@ tool if you not previously done so.
 
 ## Steps
 
-Run `prepare_release.sh` with `bom` argument in `cloud-opensource-java` directory:
+All on your corp desktop: 
+
+1. Run gcert if you have not done so in the last twelve hours or so.
+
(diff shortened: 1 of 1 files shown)


boms/cloud-oss-bom/RELEASING.md
```diff
@@ -14,10 +14,15 @@ tool if you not previously done so.
 
 ## Steps
 
-Run `prepare_release.sh` with `bom` argument in `cloud-opensource-java` directory:
+All on your corp desktop: 
+
+1. Run gcert if you have not done so in the last twelve hours or so.
```
suztomo, line 19:
```suggestion
1. Run gcert.
```

Shorter is clear.

suztomo (approved):

boms/cloud-oss-bom/RELEASING.md
elharo: Yes, but even better is not doing it at all, which is often the case. :-)

PR #1763, opened 2020-10-29
Add reason to JLBP-0005.md

### LINKED ISSUE #1695: Add reason to JLBP 5
https://jlbp.dev/JLBP-5 gives examples of the problem with overlapping classes, and a good reason to not use them in Java 9 and later. However it doesn't come out and say why they shouldn't be used in Java 8 and earlier. Add this reason. E.g.

It is unpredictable which version of a class will be found at runtime. Indeed this can change depending on how the POMs are set up. This leads to confusing, hard to diagnose bugs.

elharo:
@suztomo fixes #1695

### DIFF
--- a/docs/JLBP-0005.md
+++ b/docs/JLBP-0005.md
@@ -12,7 +12,7 @@ Definition: When the same fully qualified class name is provided by
 When a class appears in more than one location in the classpath, it is unpredictable which version
 is loaded at runtime. Indeed this can change depending on how the POMs are set up.
 If the overlapping classes are not identical—for example, one version has methods the other does not,
-or one has fixed bugs the other hasn't—this leads to confusing, hard to diagnose bugs.
+or one has fixed bugs the other hasn't—this leads to confusing, hard-to-diagnose bugs.
 
 This can happen in several ways:
 
--- a/docs/JLBP-0005.md
+++ b/docs/JLBP-0005.md
@@ -8,6 +8,11 @@ permalink: /JLBP-5
 Definition: When the same fully qualified class name is provided by
   two or more distinct artifacts (different group IDs, different
   artifact IDs, or both) the classes are said to *overlap*.
+  
+When a class appears in more than one location in the classpath, it is unpredictable which version
+is loaded at runtime. Indeed this can change depending on how the POMs are set up.
+If the overlapping classes are not identical—for example, one version has methods the other does not,
+or one has fixed bugs the other hasn't—this leads to confusing, hard to diagnose bugs.
 
 This can happen in several ways:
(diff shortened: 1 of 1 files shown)


docs/JLBP-0005.md
```diff
@@ -8,6 +8,11 @@ permalink: /JLBP-5
 Definition: When the same fully qualified class name is provided by
   two or more distinct artifacts (different group IDs, different
   artifact IDs, or both) the classes are said to *overlap*.
+  
+When a class appears in more than one location in the classpath, it is unpredictable which version
```
suztomo, line 12:
```suggestion
When a class appears in more than one location in the class path, it is unpredictable which version
```
https://docs.oracle.com/javase/8/docs/technotes/tools/windows/classpath.html

docs/JLBP-0005.md
```diff
@@ -8,6 +8,11 @@ permalink: /JLBP-5
...
+is loaded at runtime. Indeed this can change depending on how the POMs are set up.
+If the overlapping classes are not identical—for example, one version has methods the other does not,
+or one has fixed bugs the other hasn't—this leads to confusing, hard to diagnose bugs.
```
suztomo, line 15:
```suggestion
or one has fixed bugs the other hasn't—this leads to confusing, hard-to-diagnose bugs.
```

suztomo (approved):

docs/JLBP-0005.md
elharo: done

docs/JLBP-0005.md
suztomo: 1. Java is Oracle's technology. Not O’Reilly's.
1. It's more up-to-date to call it "class path". Java 8 referenced the option as "-classpath". Newer distribution Java 11 references the option "-class-path". 

(old)
```
-classpath classpath1:classpath2...
```
Java 8 https://docs.oracle.com/javase/8/docs/technotes/tools/unix/classpath.html#CBHHCGFB

(new)
```
--class-path classpath, -classpath classpath, or -cp classpath
A semicolon (;) separated list of directories, JAR archives, and ZIP archives to search for class files.

Specifying classpath overrides any setting of the CLASSPATH environment variable. If the class path option isn’t used and classpath isn’t set, then the user class path consists of the current directory (.).
```
Java 11 https://docs.oracle.com/en/java/javase/11/tools/java.html#GUID-3B1CE181-CD30-4178-9602-230B800D4FAE

elharo: done

suztomo (approved):

PR #1765, opened 2020-10-29
JLBP 21: Upload artifacts to Maven Central

### LINKED ISSUE #1264: JLBP Upload to Maven Central
Many developers, especially ones who don't use Maven itself, get confused by artifacts that live in repos other than Central. Suggest all artifacts should be in central.

JCenter et al also may introduce additional security issues not present in Central.

elharo:
@suztomo @saturnism @lesv Initial draft. 

Fixes #1264


docs/JLBP-0021.md
```diff
@@ -0,0 +1,28 @@
...
+While Maven supports additional remote repositories,
+this requires special configuration
+in the pom.xml file. Tools other than Maven itself that 
+use the repository system do not always read this information from the 
+pom.xml files and often cannot load artifacts that are not available in Maven Central.
+
+Developers who manually search for particular artifacts usually start at
+Maven Central and sometimes don't look elsewhere.
+
+Firewalls sometimes block access to third party repositories.
```
suztomo, line 24: I didn't find this sentence useful. Should this paragraph guide readers to take some action upon firewall problem?

docs/JLBP-0021.md
```diff
@@ -0,0 +1,28 @@
+---
+jlbp:
+  id: JLBP-21
+permalink: /JLBP-21
+---
+# Upload artifacts to Maven Central
```
suztomo, line 6: Did you have specific problems in your mind when you wrote this? I want to see it in this document.

docs/JLBP-0021.md
elharo: The action to take it to put things in Maven Central, not in the repositories the firewall blocks.

saturnism:
great point. i take it for granted, but it's possible that some of our public libraries are not published to maven central in the past.

saturnism:
wdyt abt making this more generic?

1. Artifacts should be made available in a Maven repository
1. If it's for public consumption, use Maven Central
1. Otherwise, a private repo for internal consumption

On additional point to is that consuming a JAR file manually also requires the developer to figure out all the dependency libraries too.

elharo:
I've mostly encountered problems here with non-Google dependencies, Eclipse and JBoss in particular.

docs/JLBP-0021.md
elharo: Typically it's happened with things like Gradle and SBT. I'd have to spend some time digging up specific cases.

docs/JLBP-0021.md
suztomo: I misread this. Now I know this sentence is here to support "Upload artifacts to Maven Central".

My confusion came from the fact that my previous workplace blocked Maven Central and copied only approved artifacts. So Maven Central is not immune to firewalls.

docs/JLBP-0021.md
suztomo: Looking forward.

elharo: one example: https://github.com/grpc/grpc-java/issues/7276

suztomo: Thanks. If I read that correctly grpc-java have always published the library to Maven Central, but not always with JCenter. So this problem cannot be fixed by uploading some artifacts to Maven Central.

elharo: I don't think that's true. Regardless, we've hit this repeatedly, every few months, in one context or another. I don't think we should list every example here, which might be resolved in the future.

suztomo: (You definitely have troubleshooted more cases than I have.) No need to list every occurrence. I want to learn just one case where this JLBP "Upload artifacts to Maven Central" would have prevented it.

suztomo: Ktor is a Kotlin library. https://ktor.io/docs/quickstart-index.html#set-up-a-ktor-project I guess jcenter is Kotlin's standard.

elharo: That Ktor involves Kotlin is not relevant. Someone tried to use this library from Maven and it failed because it wasn't in Maven Central. This is a very common occurrence. I've encountered it multiple times myself. There are many different causes for it, but one solution. Deploy to Maven Central.

docs/JLBP-0021.md
suztomo: For companies that maintains their own repositories inside their network, Maven Central is a third party repository. They sometimes block Maven Central by firewalls or HTTP proxy.

docs/JLBP-0021.md
suztomo: Yes, one solution is for the Kotlin library maintainers to deploy the library to Maven central. Another solution, which I believe most common, is to include repositories in pom.xml.

From Java developers' perspective, having everything in Maven Central is simpler. I just feel that it's strange to recommend Kotlin library maintainers to publish their libraries to Maven Central when their standard is to use jcenter.

elharo: It does not seem to be the Kotlin standard to use JCenter. Regardless, I have no compunctions about recommending they use Maven Central instead. There are many, many cases where using other repos causes needless pain for developers. None of these are unfixable, but there's no reason to make developers worry about this when there's such a simple solution. I do not want to litigate each and every instance of this problem.

lesv: Are you aware of: https://maven.google.com/web/index.html ??

elharo: Yes. I think we've had problems with android dependencies in the past for the reasons addressed in this JLBP. Tools start to break when they don't find artifacts in Maven Central. Yes there's usually some way to fix it, but the JLBPs don't stop at what works if you toggle the right knobs and flip the right switches. The JLBPs recommend practices that make development as easy as possible for clients. It's easier if the artifacts are in maven Central, the default where everything looks, than if they're in some other repo.

elharo (commented):
Another case where we had ourselves had problems and had to do extra work because someone didn't publish their artifacts to maven central:

https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/338

docs/JLBP-0021.md
elharo: And still another case of this that bit me today in one of our own repos:

```
[ERROR] Failed to execute goal on project google-api-client-android: Could not resolve dependencies for project com.google.api-client:google-api-client-android:jar:1.31.2-SNAPSHOT: Could not find artifact com.google.android.google-play-services:google-play-services:jar:1 in central (https://repo.maven.apache.org/maven2) -> [Help 1]
```

I stand behind this one and I think we should publish it. Not uploading jars to Maven Central can be made to work, but that's not enough. Everyone's job is easier if libraries put their jars in Maven Central.

suztomo: One part of Google (this PR JLBP 21) is going to say everyone should put artifacts to Maven Central. Another part of Google may continue to use https://maven.google.com/web/index.html.

Do you think it's worth getting input from Android team?

docs/JLBP-0021.md
suztomo: Do you still need this line of Maven Central's immunity against firewalls? Companies may block Maven Central by firewalls or HTTP proxy. Having said that, it's easier for them to host a repository mirror inside their network when every library is hosted in Maven Central.

docs/JLBP-0021.md
elharo: Feel free to ask them to comment on this PR. However in general we should not assume that just because Google does something, it is ideal. Teams often optimize for local convenience and development velocity, even when it costs external developers more time and effort. The JLBPs are about how to publish artifacts in a way that is easiest for customers to consume, even at the cost of extra effort for library developers.

elharo:
And yet another case where I personally encountered a problem building because android doesn't publish to Maven central:

Downloading from central: https://repo.maven.apache.org/maven2/com/google/android/google-play-services/google-play-services/1/google-play-services-1.pom
[WARNING] The POM for com.google.android.google-play-services:google-play-services:jar:1 is missing, no dependency information available
Downloading from central: https://repo.maven.apache.org/maven2/com/google/android/google-play-services/google-play-services/1/google-play-services-1.jar
...
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  3.014 s
[INFO] Finished at: 2021-01-06T15:38:49-05:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal on project google-api-client-android: Could not resolve dependencies for project com.google.api-client:google-api-client-android:jar:1.31.2-SNAPSHOT: Could not find artifact com.google.android.google-play-services:google-play-services:jar:1 in central (https://repo.maven.apache.org/maven2) -> [Help 1]



This was in google-api-java-client

docs/JLBP-0021.md
suztomo: > Firewalls sometimes block access to third party repositories. 

I still think this paragraph is unnecessary, because sometimes firewalls block access to Maven Central.

elharo: I suspect it happens, but rarely. When it does, there's nothing a library developer can do about it. If the end user developer's firewall is blocking maven central, they have to fix that before they can build anything.

suztomo: Right. The existence of the restricted networks also supports having everything in Maven Central. How about saying this?

"If a library is hosted in certain repositories other than Maven Central, then it would a chore for network administrators maintaining a restricted network to setup additional mirrors or firewall rules of the repositories"

docs/JLBP-0021.md
```diff
@@ -0,0 +1,28 @@
...
+
+Developers expect to find artifacts in [Maven Central](https://repo1.maven.org/maven2/). All artifacts
+intended for public consumption
+should be published in the [Maven Central repository maintained by Sonatype](https://search.maven.org). Even if a project is not built with Maven,
+its artifacts can and should be published to Maven Central.
+
+From here the artifacts are mirrored to many different local and public repositories. 
+
+While Maven supports additional remote repositories,
```
suztomo, line 15: Clarify "additional remote repositories" and "third party repositories".

elharo, line 15: per @netdpb need to clarify third party repositories

elharo (commented):
FYI, Sonatyhpe agrees: "In addition we discourage the usage of `<repositories>` and `<pluginRepositories>` and instead publish any required components to the Central Repository. This applies for your own components as well as for 3rd party artifacts."


https://central.sonatype.org/pages/requirements.html

docs/JLBP-0021.md
elharo: done

elharo: done

docs/JLBP-0021.md
elharo: done

suztomo (approved):

PR #1766, opened 2020-10-29
JLBP: jar -> JAR file

suztomo:
"JAR" is the format of a file.

"jar" is the name of the command.

https://docs.oracle.com/javase/tutorial/deployment/jar/basicsindex.html

### DIFF
--- a/docs/JLBP-0001.md
+++ b/docs/JLBP-0001.md
@@ -89,7 +89,7 @@ In Gradle these libraries should have `compileOnly` scope:
 ```
 
 The `provided` scope should be used when the dependency is needed at runtime
-and compile time. However, the specific JAR will be supplied by the environment
+and compile time. However, the specific dependency will be supplied by the environment
 in which the code runs. For example, Java servlet containers such as Tomcat,
 Jetty, and App Engine supply `javax.servlet:javax.servlet-api`.
 
--- a/docs/JLBP-0001.md
+++ b/docs/JLBP-0001.md
@@ -89,8 +89,8 @@ In Gradle these libraries should have `compileOnly` scope:
 ```
 
 The `provided` scope should be used when the dependency is needed at runtime
-and compile time. However, the specific dependency will be supplied by the environment
-in which the code runs. For example, Java servlet containers such as Tomcat,
+and compile time. However, the JAR file of the specific dependency will be supplied
+by the environment in which the code runs. For example, Java servlet containers such as Tomcat,
 Jetty, and App Engine supply `javax.servlet:javax.servlet-api`.
 
 ## Separate the tool classpath from the product classpath
--- a/docs/JLBP-0001.md
+++ b/docs/JLBP-0001.md
@@ -89,7 +89,7 @@ In Gradle these libraries should have `compileOnly` scope:
 ```
 
 The `provided` scope should be used when the dependency is needed at runtime
-and compile time. However, the JAR file of the specific dependency will be supplied
+and compile time. However, the specific JAR file will be supplied
 by the environment in which the code runs. For example, Java servlet containers such as Tomcat,
 Jetty, and App Engine supply `javax.servlet:javax.servlet-api`.
 
--- a/docs/JLBP-0001.md
+++ b/docs/JLBP-0001.md
@@ -89,7 +89,7 @@ In Gradle these libraries should have `compileOnly` scope:
 ```
 
 The `provided` scope should be used when the dependency is needed at runtime
-and compile time. However, the specific jar will be supplied by the environment
+and compile time. However, the specific JAR will be supplied by the environment
 in which the code runs. For example, Java servlet containers such as Tomcat,
 Jetty, and App Engine supply `javax.servlet:javax.servlet-api`.
 
--- a/docs/JLBP-0006.md
+++ b/docs/JLBP-0006.md
@@ -93,7 +93,7 @@ Given this scenario, here are the possible combinations of renamings:
 - **Rename Java package**:
   - **Case 3: Keep Maven ID**. The classes from `g1:a1:1.0.0` and
     `g1:a1:2.0.0` could technically be used together, but since they share the
-    Maven group ID and artifact ID, only one artifact can be pulled in.
+    Maven group ID and artifact ID, only the classes in one of the artifacts can be pulled in.
     Users must change their code to add references to the new version.
     It is strictly better to either also rename the Maven ID (case 4)
     or to keep the Maven ID and also bundle the old package (case 5), so that
--- a/docs/JLBP-0006.md
+++ b/docs/JLBP-0006.md
@@ -82,7 +82,7 @@ Given this scenario, here are the possible combinations of renamings:
     will only choose one of them to use. Users are forced to change their code
     between versions, but only for library surface that was changed.
 
-  - **Case 2: Rename Maven ID**. Users can accidentally pull in the JAR files of both the
+  - **Case 2: Rename Maven ID**. Build systems can unexpectedly pull in the JAR files of both the
     old artifact (`g1:a1:1.0.0`) and the new artifact (`g2:a2:2.0.0`) into their class path
     through transitive dependencies, because Maven artifact
     resolution treats the two artifacts as distinct.  This results in
@@ -92,7 +92,7 @@ Given this scenario, here are the possible combinations of renamings:
 
 - **Rename Java package**:
   - **Case 3: Keep Maven ID**. The classes from `g1:a1:1.0.0` and
-    `g1:a1:2.0.0` could technically be used together, but since they share the
+    `g1:a1:2.0.0` could be used together, but since the two artifacts share the
     Maven group ID and artifact ID, the build system selects only one of them.
     Users must change their code to add references to the new version.
     It is strictly better to either also rename the Maven ID (case 4)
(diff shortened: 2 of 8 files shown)


docs/JLBP-0006.md
```diff
@@ -82,8 +82,8 @@ Given this scenario, here are the possible combinations of renamings:
     will only choose one of them to use. Users are forced to change their code
     between versions, but only for library surface that was changed.
 
-  - **Case 2: Rename Maven ID**. Users can easily pull in both the old jar
-    (`g1:a1:1.0.0`) and the new jar (`g2:a2:2.0.0`) into their classpath
+  - **Case 2: Rename Maven ID**. Users can easily pull in both the old artifact
+    (`g1:a1:1.0.0`) and the new artifact (`g2:a2:2.0.0`) into their classpath
```
suztomo, line 86: A JAR file does not have groupID or artifactID. It's a Maven artifact that has groupID and artifactID. A Maven artifact usually has a JAR file (except for pom-type artifacts).

docs/JLBP-0001.md
```diff
@@ -89,7 +89,7 @@ In Gradle these libraries should have `compileOnly` scope:
 ```
 
 The `provided` scope should be used when the dependency is needed at runtime
-and compile time. However, the specific jar will be supplied by the environment
+and compile time. However, the specific dependency will be supplied by the environment
```
elharo, line 92: JAR is correct here

docs/JLBP-0006.md
elharo: but only JARs are added to classpaths

docs/JLBP-0019.md
```diff
@@ -41,12 +41,12 @@ For example, `com.google.i18n.Localization` can be in the artifact
 It is also acceptable to bundle classes from both `com.google.i18n` and
 `com.google.i18n.charactersets` in a single Maven module.
 Similarly, it is fine to combine completely different packages such as
-`com.google.foo` and `org.example.bar` in the same JAR with the same group ID and artifact ID.
+`com.google.foo` and `org.example.bar` in the same artifact with the same group ID and artifact ID.
```
elharo, line 44: JAR file

docs/JLBP-0001.md
suztomo: Updated.

docs/JLBP-0006.md
suztomo: That's true. Updated.

docs/JLBP-0019.md
suztomo: Updated to "a single module", because this page discusses modules.

I didn't get the meaning of 2 occurrences of "the same" in this line. We cannot publish two Maven artifacts with "the same" group ID and artifact ID. Do they mean "one" or "a"?

docs/JLBP-0006.md
```diff
@@ -93,7 +93,7 @@ Given this scenario, here are the possible combinations of renamings:
 - **Rename Java package**:
   - **Case 3: Keep Maven ID**. The classes from `g1:a1:1.0.0` and
     `g1:a1:2.0.0` could technically be used together, but since they share the
-    Maven group ID and artifact ID, only one jar can be pulled in.
+    Maven group ID and artifact ID, only one artifact can be pulled in.
```
elharo, line 96: I think JAR file is right here. "pulled in" really means (and probably should be rewritten as) "added to the class path". JAR files are added to the class path. Artifacts aren't.  What do you think?

docs/glossary.md
```diff
@@ -75,7 +75,7 @@ permalink: /glossary
 
 - A particular conflict cannot be both a linkage error and behavior conflict.
   They are mutually exclusive.
-- A combination of jars at runtime can have any number of linkage errors and
+- A combination of JAR files at runtime can have any number of linkage errors and
```
elharo, line 78: combination is an unclear word here. Not in this PR, but we can clarify this.

docs/JLBP-0019.md
```diff
@@ -41,12 +41,12 @@ For example, `com.google.i18n.Localization` can be in the artifact
 It is also acceptable to bundle classes from both `com.google.i18n` and
 `com.google.i18n.charactersets` in a single Maven module.
 Similarly, it is fine to combine completely different packages such as
-`com.google.foo` and `org.example.bar` in the same JAR with the same group ID and artifact ID.
+`com.google.foo` and `org.example.bar` in a single module with a group ID and an artifact ID.
```
elharo, line 44: I'm not sure about this change.

suztomo:
@elharo PTAL. I need your help to clarify the meaning of "the same"s in the following sentence:

> Similarly, it is fine to combine completely different packages such as `com.google.foo` and `org.example.bar` in the same JAR with the same group ID and artifact ID.

docs/JLBP-0019.md
elharo: Probably change "in a single module with a group ID and an artifact ID" to "in a single JAR file".

suztomo: Updated it to "in a single JAR file".

docs/JLBP-0006.md
suztomo: How about "_only the classes in one of the artifacts can be pulled in_"? This sentence starts with "The classes" and classes are part of class path (through JAR files).

suztomo (commented):
@elharo PTAL.

docs/JLBP-0006.md
suztomo: Updated to "the build system selects only one of them" (active voice; without using "pull in")

docs/JLBP-0001.md
```diff
@@ -89,8 +89,8 @@ In Gradle these libraries should have `compileOnly` scope:
 ```
 
 The `provided` scope should be used when the dependency is needed at runtime
-and compile time. However, the specific jar will be supplied by the environment
-in which the code runs. For example, Java servlet containers such as Tomcat,
+and compile time. However, the JAR file of the specific dependency will be supplied
```
elharo, line 92: No, specific jar was correct here.

docs/JLBP-0006.md
```diff
@@ -82,9 +82,9 @@ Given this scenario, here are the possible combinations of renamings:
     will only choose one of them to use. Users are forced to change their code
     between versions, but only for library surface that was changed.
 
-  - **Case 2: Rename Maven ID**. Users can easily pull in both the old jar
-    (`g1:a1:1.0.0`) and the new jar (`g2:a2:2.0.0`) into their classpath
-    accidentally through transitive dependencies because Maven artifact
+  - **Case 2: Rename Maven ID**. Users can accidentally pull in the JAR files of both the
```
elharo, line 85: Users --> Build system?
Also, I'm not sure it's accidental.

docs/JLBP-0006.md
```diff
@@ -93,7 +93,7 @@ Given this scenario, here are the possible combinations of renamings:
 - **Rename Java package**:
   - **Case 3: Keep Maven ID**. The classes from `g1:a1:1.0.0` and
     `g1:a1:2.0.0` could technically be used together, but since they share the
```
elharo, line 95: delete "technically"
they --> the two artifacts

docs/JLBP-0006.md
suztomo: Updated to "build systems".

Updated to "unexpectedly".

docs/JLBP-0006.md
suztomo: Updated.

suztomo (commented):
@elharo PTAL.

docs/JLBP-0006.md
```diff
@@ -11,7 +11,7 @@ repository system, library B needs two identifiers to find classes in library A:
 1. The Maven coordinates of library A, following the form
    `group_ID:artifact_ID:version`; for example
    `com.google.guava:guava:26.0-jre`. The Maven coordinates are used to locate the
-   library's jar in a Maven repository. For each pair of group ID and artifact ID
+   library's JAR file in a Maven repository. For each pair of group ID and artifact ID
```
elharo, line 14: "library's JAR file" --> library's files 

(this is used for POMs, signatures, and other non-JAR files too)

docs/JLBP-0006.md
```diff
@@ -82,18 +82,18 @@ Given this scenario, here are the possible combinations of renamings:
...
     "overlapping classes" since they have classes with the same fully-qualified
     path. This can lead to runtime exceptions such as `NoSuchMethodDefError`
     and can be a compile-time error in Java 9 and later. **NEVER DO THIS**.
 
 - **Rename Java package**:
   - **Case 3: Keep Maven ID**. The classes from `g1:a1:1.0.0` and
-    `g1:a1:2.0.0` could technically be used together, but since they share the
-    Maven group ID and artifact ID, only one jar can be pulled in.
+    `g1:a1:2.0.0` could be used together, but since the two artifacts share the
+    Maven group ID and artifact ID, the build system selects only one of them.
```
elharo, line 96: "selects only one of them" --> adds only one of them to the class path.

docs/JLBP-0006.md
suztomo: Yes, updated.

docs/JLBP-0006.md
suztomo: Updated.

suztomo (commented):
@elharo PTAL.

elharo (approved):

PR #1768, opened 2020-10-29
deps: update google-http-client to 1.37.0

chingor13:
The next version of google-cloud-storage declares google-http-client 1.37.0

https://github.com/googleapis/java-cloud-bom/pull/1074

### DIFF
--- a/boms/cloud-oss-bom/pom.xml
+++ b/boms/cloud-oss-bom/pom.xml
@@ -48,7 +48,7 @@
     <google.cloud.java.version>0.141.0</google.cloud.java.version>
     <io.grpc.version>1.33.0</io.grpc.version>
     <protobuf.version>3.13.0</protobuf.version>
-    <http.version>1.36.0</http.version>
+    <http.version>1.37.0</http.version>
     <!-- We don't use gax-bom because it includes the artifacts with 'testlib' classifier.
         When updating gax.version, update gax.httpjson.version too. -->
     <gax.version>1.60.0</gax.version>


elharo:
don't think you can upgrade this one in isolation:

Failed while enforcing RequireUpperBoundDeps. The error(s) are [
Require upper bound dependencies error for com.fasterxml.jackson.core:jackson-core:2.11.1 paths to dependency are:
+-com.google.cloud.tools.opensource:upper-bounds-check:2.0.0-SNAPSHOT
  +-com.google.cloud:google-cloud-storage:1.113.1
    +-com.fasterxml.jackson.core:jackson-core:2.11.1
and
+-com.google.cloud.tools.opensource:upper-bounds-check:2.0.0-SNAPSHOT
  +-com.google.cloud:google-cloud-core:1.93.9
    +-com.google.http-client:google-http-client-jackson2:1.37.0
      +-com.fasterxml.jackson.core:jackson-core:2.11.3

chingor13:
So we can't update google-cloud-storage which is causing that in google-cloud-bom without google-http-client 1.37.0 in libraries-bom

elharo (commented):
I think you'll need to add an explicit java-storage dependency here to override the one from google-cloud-bom.

suztomo:
> an explicit java-storage dependency here

@elharo Which file are you suggesting updating?

elharo:
Try adding an explicit dependency on google-cloud-storage to this pom.xml file, instead of just importing google-cloud-bom. I think that should override the version  we import from google-cloud-bom.

elharo:
seems GAX is way behind and causing problems. We should update all its dependencies and release it.

elharo:
we also need a new google-cloud-core release

elharo (commented):
Here's what I've figured out so far:

1. GAX needs many updates and a release.
1.5 Possible shared dependencies needs the the new GAX in which case it needs a release. 
2. google-cloud-core needs a release
3. There needs to be a new google-cloud-bom that includes all of these.
4. libraries-bom needs to update GAX, http-client, auth-client, and google-cloud-bom at the same time.

These are necessary conditions. I'm not 100% sure they're sufficient. I suggest blocking out some time for the two of us to pair to send PRs, review, and release so we don't have to wait for day+ reviews.

elharo:
The root cause is that several libraries have not been keeping up to date with dependencies and gotten behind. I'd really like to take a stab at fixing these.

elharo:
I think flattening is confusing matters, at least for me, because what's published on Maven central is not what I see in the repos. It's giving some dependencies more weight than they should have.

elharo:
closing in favor of #1771 but we are going to need to make some of these updates in GAX et al

PR #1771, opened 2020-10-30
Update auth and http client

elharo:
@suztomo @chingor13 I had to swap the order of dependencies we declare, though this really shouldn't matter.

### DIFF
--- a/boms/cloud-oss-bom/pom.xml
+++ b/boms/cloud-oss-bom/pom.xml
@@ -230,7 +230,7 @@
        <dependency>
          <groupId>com.google.auth</groupId>
          <artifactId>google-auth-library-bom</artifactId>
-         <version>0.21.1</version>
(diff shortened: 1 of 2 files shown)


boms/cloud-oss-bom/pom.xml
```diff
@@ -67,6 +67,7 @@
 
   <dependencyManagement>
     <dependencies>
+
```
suztomo, line 70: Unnecessary empty line.

suztomo (approved):
Nice solution.

boms/cloud-oss-bom/pom.xml
elharo: removed

elharo (commented):
I'm still not happy with this. It unblocks us, but ideally the upper bounds conditions should be satisfied in any order.

PR #1797, opened 2020-11-09
tighten language

elharo:
@suztomo

### DIFF
--- a/docs/JLBP-0016.md
+++ b/docs/JLBP-0016.md
@@ -5,8 +5,7 @@ permalink: /JLBP-16
 ---
 # Ensure upper version alignment of dependencies for consumers
 
-Upper version alignment means that the version of each dependency 
-added to the classpath is the highest version in the dependency tree.
+The version of each dependency added to the classpath should the highest version in the dependency tree.
 Upper version alignment ensures that upgrading packages to compatible higher versions doesn't introduce new linkage conflicts.
 
 ## Achieving upper version alignment


docs/JLBP-0016.md
```diff
@@ -5,8 +5,7 @@ permalink: /JLBP-16
 ---
 # Ensure upper version alignment of dependencies for consumers
 
-Upper version alignment means that the version of each dependency 
-added to the classpath is the highest version in the dependency tree.
```
suztomo, line 9: This sentence was defining the term "upper version alignment" and the next sentence was referring to it. Removing the definition breaks the sequence.

elharo: Definitions belong in the glossary. I think this is clear enough in context, and now it affirmatively recommends the practice rather than simply defining it.

docs/JLBP-0016.md
```diff
@@ -5,8 +5,7 @@ permalink: /JLBP-16
...
+The version of each dependency added to the classpath should the highest version in the dependency tree.
 Upper version alignment ensures that upgrading packages to compatible higher versions doesn't introduce new linkage conflicts.
```
suztomo, line 9:
```suggestion
This upper version alignment ensures that upgrading packages to compatible higher versions doesn't introduce new linkage conflicts.
```

How about this?

elharo: I don't know. Reads a little awkward to me. maybe it impies there's some other upper version alignment in play?

docs/JLBP-0016.md
```diff
@@ -5,8 +5,7 @@ permalink: /JLBP-16
 ---
 # Ensure upper version alignment of dependencies for consumers
 
-Upper version alignment means that the version of each dependency 
-added to the classpath is the highest version in the dependency tree.
+The version of each dependency added to the classpath should the highest version in the dependency tree.
```
suztomo, line 8:
```suggestion
The version of each dependency added to the classpath should the highest version in the dependency tree. This state is called upper version alignment.
```

How about this?

suztomo (approved):
Added a suggestion about the term.

PR #1813, opened 2020-11-17
Google cloud bom 0.144.0

suztomo:


boms/cloud-oss-bom/pom.xml
```diff
@@ -246,11 +246,6 @@
          <type>pom</type>
          <scope>import</scope>
        </dependency>
-      <dependency>
-        <groupId>com.google.cloud</groupId>
-        <artifactId>google-cloud-storage</artifactId>
-        <version>1.113.3</version>
-      </dependency>
```
suztomo, line 253: We no longer declare this explicitly.

elharo (approved):

PR #1836, opened 2020-11-24
LinkageCheckerTest.testFindLinkageProblems_referenceToJava11Method fails in Java 11 to use Assume Java 8 runtime

### LINKED ISSUE #1822: LinkageCheckerTest.testFindLinkageProblems_referenceToJava11Method fails in Java 11
Towards #1821 .

```
suztomo-macbookpro44% which java
/Library/Java/JavaVirtualMachines/jdk-11-latest/Contents/Home/bin/java
suztomo-macbookpro44% pwd        
/Users/suztomo/cloud-opensource-java/dependencies
suztomo-macbookpro44% mvn test -Dtest=LinkageCheckerTest#testFindLinkageProblems_referenceToJava11Method 
...
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.154 s <<< FAILURE! - in com.google.cloud.tools.opensource.classpath.LinkageCheckerTest
[ERROR] testFindLinkageProblems_referenceToJava11Method(com.google.cloud.tools.opensource.classpath.LinkageCheckerTest)  Time elapsed: 1.081 s  <<< FAILURE!
com.google.common.truth.AssertionErrorWithFacts: 
value of           : findLinkageProblems()
expected to contain: java.nio.CharBuffer's method flip() is expected to return java.nio.CharBuffer but instead returns java.nio.Buffer referenced by ClassFile{location=com.google.protobuf:protobuf-java:3.12.4, className=com.google.protobuf.TextFormat}
but was            : []
	at com.google.cloud.tools.opensource.classpath.LinkageCheckerTest.testFindLinkageProblems_referenceToJava11Method(LinkageCheckerTest.java:1150)

```

The problem of `java.nio.CharBuffer`'s invalid return type is only applicable to Java 8 runtime.

Should Linkage Checker have a mode to switch Java runtime version?

What's user's experience? In a certain scenario, should a user only care about Java 8? Or Just Java 11?

What about a user who currently uses Java 11 in their builds but wants to ensure their library runs well on Java 8?

suztomo:
Fixes #1822 . If we run the test on Java 11, the test says

```
org.junit.AssumptionViolatedException: got: "11.0.4", expected: a string starting with "1.8.0"
```

and JUnit ignores the test.

### DIFF
--- a/dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerTest.java
+++ b/dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerTest.java
@@ -19,7 +19,6 @@ package com.google.cloud.tools.opensource.classpath;
 import static com.google.cloud.tools.opensource.classpath.TestHelper.COORDINATES;
 import static com.google.cloud.tools.opensource.classpath.TestHelper.classPathEntryOfResource;
 import static com.google.common.collect.ImmutableList.toImmutableList;
-import static org.hamcrest.core.StringStartsWith.startsWith;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
@@ -47,6 +46,7 @@ import org.eclipse.aether.RepositoryException;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.artifact.DefaultArtifact;
 import org.eclipse.aether.graph.Dependency;
+import org.hamcrest.core.StringStartsWith;
 import org.junit.Assume;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerTest.java
```diff
@@ -1130,6 +1132,8 @@ public void testFindLinkageProblems_grpcAndGuava() throws IOException {
   public void testFindLinkageProblems_referenceToJava11Method() throws IOException {
     // protobuf-java 3.12.4 references a Java 11 method that does not exist in Java 8
     // https://github.com/protocolbuffers/protobuf/issues/7827
+    Assume.assumeThat(System.getProperty("java.version"), startsWith("1.8.0"));
```
suztomo, line 1135: Static import reads well, compared with `StringStartsWith.startsWith`.

elharo: why not use java.lang.String.startsWith()? In not, because that method is much more familiar, this really can't be a static import.

suztomo: `String.startsWith` would give a message `got false expected true`.

This `StringStartsWith.startsWith` gives more clear message: `got: "11.0.4", expected: a string starting with "1.8.0"`.

elharo: Fair enough. Static import is still not a good idea here given the confusion with the better known String.startsWith method.

elharo (approved):

Issue #1843, opened 2020-11-30
Implement a more deliberate release strategy for linkage monitor


elharo:

suztomo:
Ideas

# Checks

- Smoke test in front of deployments (the nightly builds). Better to test a JAR file after shading.
- CI in this repository to ensure Linkage Monitor works fine in
  - dummy projects
  - some of the downstream repositories. (the latest status of the repo is unknown)

# Release Process

suztomo:
Created "Linkage Monitor and GitHub Actions Versioning" http://go/jdd-github-actions . If we maintain the version through Github Actions, we need a new Github repository.

suztomo:
This has been implemented.
https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/RELEASING.md#release

PR #1851, opened 2020-12-02
smoke test for main method

elharo:
@suztomo step #1 for #1843


suztomo (approved):
Is this the smoke test you mentioned yesterday?

elharo:
It's one of them. We should still look at a test that runs on the jar before deployment.

Issue #1779, opened 2020-11-03
Investigate new linkage errors in gRPC 1.33.2


elharo:

suztomo:
Errors in https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1774#issuecomment-720601846

elharo:
done

PR #1865, opened 2020-12-07
Libraries BOM: gRPC and google-auth-library version upgrade

suztomo:
update dependency com.google.auth:google-auth-library-bom to v0.22.1
update io.grpc.version to v1.34.0

The build succeeded in my local without showing the error in https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1779. Does this work?

### DIFF
--- a/boms/cloud-oss-bom/pom.xml
+++ b/boms/cloud-oss-bom/pom.xml
@@ -46,7 +46,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <guava.version>30.0-android</guava.version>
     <google.cloud.java.version>0.145.0</google.cloud.java.version>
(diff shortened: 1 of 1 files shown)


suztomo:
Somehow kokoro-macos does not trigger the build. Posted a note to kokoro-users@.

elharo (approved):

elharo:
Per @ejona "It may be best to revert this change, or hold off on releasing until you can bump to a 1.34.1. I'm currently investigating grpc/grpc-java#7643, but it appears 1.34.0 has bad credential handling that would impact direct path for some Google properties."

I don't think we need to revert but let's not release until this is resolved.

suztomo:
Good information. Thank you.

PR #1877, opened 2020-12-11
gRPC 1.34.0 is broken

elharo:
@suztomo


suztomo (approved):
I thought we would wait. But if you know it takes longer for gRPC team to release a new version, then let's downgrade the version.

PR #1879, opened 2020-12-13
google-cloud-bom 0.146.0

elharo:
@suztomo


elharo (commented):
As I feared, 0.146.0 pulls in the broken gRPC. We'll have to wait for a new gRPC version (or a google-cloud-bom rollback) before we can do this.

PR #1888, opened 2020-12-17
try JRE

elharo:
I'm curious what exactly breaks and why if we use the JRE version. Java 7 of course. What else?

PR #1889, opened 2020-12-21
Trying Guava JRE flavor without using Guava BOM

suztomo:
https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1888#issuecomment-749044402

- Guava Android BOM has "guava" and "guava-testlib" artifacts.
- Guava JRE BOM has "guava", "guava-gwt", and "guava-testlib".

This PR uses "guava" and "guava-testlib" so that the Libraries BOM does not include the "guava-gwt".

### DIFF
--- a/boms/cloud-oss-bom/pom.xml
+++ b/boms/cloud-oss-bom/pom.xml
@@ -69,10 +69,13 @@
     <dependencies>
       <dependency>
         <groupId>com.google.guava</groupId>
-        <artifactId>guava-bom</artifactId>
+        <artifactId>guava</artifactId>
         <version>${guava.version}</version>
-        <type>pom</type>
-        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava-testlib</artifactId>
+        <version>${guava.version}</version>
       </dependency>
 
       <!-- protobufs from https://github.com/protocolbuffers/protobuf/tree/master/java -->
--- a/boms/cloud-oss-bom/pom.xml
+++ b/boms/cloud-oss-bom/pom.xml
@@ -44,7 +44,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <guava.version>30.1-android</guava.version>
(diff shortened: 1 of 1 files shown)


boms/cloud-oss-bom/pom.xml
```diff
@@ -69,10 +69,13 @@
     <dependencies>
       <dependency>
         <groupId>com.google.guava</groupId>
-        <artifactId>guava-bom</artifactId>
+        <artifactId>guava</artifactId>
+        <version>${guava.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava-testlib</artifactId>
```
elharo, line 77: We really shouldn't include this one. guava-gwt is intended for clients, though we don't have to include it if we aren't already. guava-testlib isn't.

suztomo:
5eeb408 passes the build. Trying removing guava-testlib.

Note for myself: the discussion on guava-testlib 2 years ago: https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/225#discussion_r232422368. Garrett says guava-testlib is used by google-cloud-clients. The shared dependencies BOM imports guava-bom (and therefore guava-testlib). https://search.maven.org/artifact/com.google.cloud/google-cloud-shared-dependencies/0.17.0/pom . Where is it used?

elharo:
It does have some nice classes. They're all beta though.

suztomo:
Yes. I see it's at least used in google-cloud-spanner: https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/pom.xml#L295 . As the scope is `test`, the dependency does not leak into users. [The BOM Dashboard's dependency tree](https://storage.googleapis.com/cloud-opensource-java-dashboard/com.google.cloud/libraries-bom/snapshot/dependency_trees.html) also shows guava-testlib is **not** part of the published pom.xml in the artifacts.

suztomo:
Discussed with Elliotte and David. When we switch to jre flavor, we will continue to include guava-testlib in the BOM for the sake of compatibility. There should be no harm to do that.

Note that we'd like to defer to the jre switching to the mid of 2021, unless we have to do so early.

PR #1903, opened 2021-01-12
remove unused warning suppression

elharo:
@suztomo


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/ExclusionFiles.java
```diff
@@ -82,7 +82,6 @@
  *   <li>A FieldMatcher matches field symbol specified by className and name attribute.
  * </ul>
  */
-@SuppressWarnings("deprecation")
```
suztomo, line 85: I see it is for XMLReaderFactory.

Screen Shot 2021-01-12 at 10 12 03

How about moving this to the `createXmlReader()` method (a smaller scope than the entire class)?

PR #1904, opened 2021-01-12
test possible LTS version scheme

elharo:
@chingor13

### DIFF
--- a/dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/VerTest.java
+++ b/dependencies/src/test/java/com/google/cloud/tools/opensource/dependencies/VerTest.java
@@ -31,4 +31,25 @@ public class VerTest {
     Assert.assertTrue(new ComparableVersion("1.38.0.1").compareTo(new ComparableVersion("1.39.0")) < 0);
   }
 
+  @Test
+  public void testTwoDigits() {
+    ComparableVersion lesser = new ComparableVersion("1.38.9.1");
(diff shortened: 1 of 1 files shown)


suztomo:
This observation defies this (maybe outdated?) Oracle document https://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#MAVEN401. Interesting.


> Maven arranges the version list in the following manner:
> 1.0.1.0
> 1.0.10.1
> 1.0.10.2
> 1.0.9.3
> Version 1.0.9.3 should come before 1.0.10.1 and 1.0.10.2, but the unexpected fourth field (.3) forced Maven to evaluate the version as a string.

elharo:
The Oracle document you cite is neither normative nor correct. https://maven.apache.org/pom.html#Version_Order_Specification is the normative documentation and it does not change the sorting of numerics when there are more than three pieces.

suztomo:
Right. Maven versioning scheme is more human-friendly than I imagined:

This test passes to my surprise:

```
  @Test
  public void testVersionWithLabels2() {
    ComparableVersion sts = new ComparableVersion("1.38.0");
    ComparableVersion lts1 = new ComparableVersion("1.38.0-LTS2");
    ComparableVersion lts2 = new ComparableVersion("1.38.0-LTS10");
    Assert.assertTrue(sts.compareTo(lts1) < 0);
    Assert.assertTrue(lts1.compareTo(lts2) < 0);
  }
```

PR #1907, opened 2021-01-14
latest versions

elharo:
@suztomo anything else we need to update?


suztomo:
As per https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1674, there are other new versions:

- com.google.cloud:google-cloud-bom to v0.147.0
- iam.protos.version to v1.0.7 (com.google.api.grpc:grpc-google-iam-v1, com.google.api.grpc:proto-google-iam-v1)

elharo (commented):
OK, so let's commit this and then update those before we release.

suztomo (approved):

PR #1913, opened 2021-01-15
Fixing sed parameter for the release script in gLinux

suztomo:
The `sed` command's options have slightly different between Mac and gLinux:

In Mac, in-place update without a backup is `-i ""`, while Linux it tries to open a filename with an empty string. In gLinux we do not need to specify the empty string for `-i` option:

```
suztomo@suztomo:~/cloud-opensource-java$ sed -i "" -e "s/version = .*/version = ${VERSION}/" gradle-plugin/gradle.properties
sed: can't read : No such file or directory
```

In gLinux, `-i "s/version = .*/version = ${NEXT_SNAPSHOT}/"` means opening file `"s/version = .*/version = ${NEXT_SNAPSHOT}/"` as a backup. We need to specify `-e` explicitly for the sed expression.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -83,7 +83,7 @@ fi
 mvn org.codehaus.mojo:versions-maven-plugin:2.7:set -DnewVersion=${VERSION} -DgenerateBackupPoms=false
 
 if [[ "${SUFFIX}" = "dependencies" ]]; then
-  sed -i "" "s/version = .*/version = ${VERSION}/" gradle-plugin/gradle.properties
+  sed -i -e "s/version = .*/version = ${VERSION}/" gradle-plugin/gradle.properties
 fi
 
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -83,7 +83,7 @@ fi
 mvn org.codehaus.mojo:versions-maven-plugin:2.7:set -DnewVersion=${VERSION} -DgenerateBackupPoms=false
 
 if [[ "${SUFFIX}" = "dependencies" ]]; then
-  sed -i "" "s/version = .*/version = ${VERSION}/" gradle-plugin/gradle.properties
```
suztomo, line 86: It failed with `sed: can't read s/version = .*/version = 1.5.4/: No such file or directory`

elharo (approved):

PR #1927, opened 2021-01-27
Upgrading dependencies

suztomo:
No major version change.


elharo (commented):
not sure about mockito.
We should separate BOM updates from our own dependencies

suztomo:
Right. That will create a clear PR.

PR #1951, opened 2021-02-23
don't throw exception for user input error

elharo:
@suztomo

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
@@ -115,9 +115,11 @@ public class DashboardMain {
       throws IOException, TemplateException, RepositoryException, URISyntaxException,
       MavenRepositoryException {
     List<String> elements = Splitter.on(':').splitToList(versionlessCoordinates);
-    checkArgument(
-        elements.size() == 2,
-        "The versionless coordinates should have one colon character: " + versionlessCoordinates);
+    if ( elements.size() != 2 ) {
+      System.err.println(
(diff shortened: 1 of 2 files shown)


dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
```diff
@@ -115,9 +115,11 @@ private static void generateAllVersions(String versionlessCoordinates)
       throws IOException, TemplateException, RepositoryException, URISyntaxException,
       MavenRepositoryException {
     List<String> elements = Splitter.on(':').splitToList(versionlessCoordinates);
-    checkArgument(
-        elements.size() == 2,
-        "The versionless coordinates should have one colon character: " + versionlessCoordinates);
+    if (elements.size() != 2) {
+      System.err.println(
+          "Versionless coordinates should have one colon: " + versionlessCoordinates);
+      return;
```
suztomo, line 121: This unexpectedly returns success for invalid input. Maven-exec-plugin would show SUCCESS for bad input. Can you return none-zero status by “System.exit(1)”?

suztomo: Now I’m not sure whether System.exit is a good way to tell Maven. It might abruptly finish the exec plugin.

suztomo (approved):

dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/DashboardMain.java
elharo: That depends on whether the goal is `exec:exec` or `exec:java`.

PR #1920, opened 2021-01-21
Linkage Monitor to read content of a BOM

suztomo:
Fixes #1919

PR #1960, opened 2021-02-25
Linkage Monitor not to read dependencyManagement section of parent poms

### LINKED ISSUE #1958: Linkage Monitor unexpectedly recognizes the root parent pom as a BOM
(This is a regression of #1920)

The problem in java-asset repository https://github.com/googleapis/java-asset/issues/570 was caused by Linkage Monitor unexpectedly recognizing the root pom.xml as a BOM and thus reading the dependencyManagement section to create a snapshot Libraries BOM.

We want to read the dependencyManagement section of BOMs to populate `LinkageMonitor.localArtifacts`, because gax-java repository has gax-bom to tell which artifacts are generated by the repository.

Note that google-cloud-bom is a BOM but `LinkageMonitor.localArtifacts` does not need to have its dependencyManagement section.

suztomo:
Fixes #1958.

@elharo Do you think of a better way than checking "-bom" to distinguish a BOM from a non-BOM?

- On one hand, I want Linkage Monitor to read the dependencyManagement section of the gax-bom.
  The gax-java repository is a Gradle project and the repository does not have pom.xml files for its artifacts (gax, gax-httpjson, gax-grpc). This is the background of https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/1920.
- On the other hand, I don't want Linkage Monitor to read the dependencyManagement section of the [google-cloud-asset-parent pom.xml](https://github.com/googleapis/java-asset/blob/master/pom.xml#L68). This behavior caused  #1958.

### DIFF
--- a/linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
+++ b/linkage-monitor/src/main/java/com/google/cloud/tools/dependencies/linkagemonitor/LinkageMonitor.java
@@ -194,8 +194,17 @@ public class LinkageMonitor {
         artifactToVersion.put(versionlessCoordinates, model.getVersion());
         logger.fine("Found local artifact: " + model);
         DependencyManagement dependencyManagement = model.getDependencyManagement();
-        if ("pom".equals(model.getPackaging()) && dependencyManagement != null) {
-          // Read the content of a BOM.
+
+        // For gax-java repository (Gradle), the dependencyManagement section of the gax-bom/pom.xml
+        // tells what artifacts the repository produces. However, we have to distinguish BOMs from
+        // normal parent poms
+        // (https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1958).
+        // Unfortunately the difference between a BOM and a parent POM is how we name it. Therefore
+        // this logic checks whether the artifactId of the pom.xml files has "-bom" or not.
+        if (model.getArtifactId().endsWith("-bom")
+            && "pom".equals(model.getPackaging())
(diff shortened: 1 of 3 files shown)


elharo (commented):
Would a BOM have a dependenciesManagement section and not a dependencies section?

suztomo:
BOM has a dependencyManagement section. BOM would not have dependencies section.

The same for a parent POM (such as google-cloud-asset-parent pom.xml). A parent pom may have a dependencyManagement section and would not have dependencies section.

elharo:
Is there a difference between a BOM and a parent POM then? Or could one theoretically be used in place of the other? That is,

* BOMs: have a dependencyManagement section and do not have a dependencies section
* non-BOMs: have a dependencies section. Might have a dependencyManagement section.

Thinking about it, I think the difference between a BOM and a parent pom is how they're imported, not any content in the pom.xml itself. Is that a useful way to think about this?

suztomo:
Yes, and I agree with your analysis. The difference is only how they’re
imported.

Do you also think this PR checking the name is best available way to fix
the problem?

(Your comment brought me an idea of checking “parent” section of all pom
files in the repository to identify parent poms. But it may be overkill if
checking its name suffices)

elharo:
I just saw Stephanie's issue. I'm not sure what the right fix is here. I might need to sleep on this one.

Generically what's the problem with checking a parent pom a we would a BOM? Do we think linkage errors are OK there? I suppose it's possible that the entire set of artifacts is never used together, but that's true for any BOM.

suztomo:
To implement it as reading a file that lists artifacts.

Issue #1977, opened 2021-03-10
Validate netty's linkage errors. If they are false positive, then add them to the default exclusion rule


suztomo:
https://storage.googleapis.com/cloud-opensource-java-dashboard/com.google.cloud/libraries-bom/snapshot/io.grpc_grpc-netty_1.36.0.html

# org.jboss.marshalling.Marshaller

If the netty user is not using MarshallingEncoder, the linkage errors will not cause an error.

```
io.netty:netty-codec:4.1.52.Final
25 target classes causing linkage errors referenced from 42 source classes.

Class org.jboss.marshalling.Marshaller is not found, referenced from 3 classes ▼

io.netty.handler.codec.marshalling.ThreadLocalMarshallerProvider
io.netty.handler.codec.marshalling.MarshallingEncoder
io.netty.handler.codec.marshalling.CompatibleMarshallingEncoder
Class org.jboss.marshalling.MarshallerFactory is not found, referenced from 4 classes ▼

io.netty.handler.codec.marshalling.ThreadLocalMarshallerProvider
io.netty.handler.codec.marshalling.ThreadLocalUnmarshallerProvider
io.netty.handler.codec.marshalling.DefaultMarshallerProvider
io.netty.handler.codec.marshalling.DefaultUnmarshallerProvider
Class org.jboss.marshalling.Unmarshaller is not found, referenced from 4 classes ▼

io.netty.handler.codec.marshalling.ThreadLocalUnmarshallerProvider
io.netty.handler.codec.marshalling.ContextBoundUnmarshallerProvider
io.netty.handler.codec.marshalling.MarshallingDecoder
io.netty.handler.codec.marshalling.CompatibleMarshallingDecoder

Class org.jboss.marshalling.ByteInput is not found, referenced from 2 classes ▼

io.netty.handler.codec.marshalling.LimitingByteInput
io.netty.handler.codec.marshalling.ChannelBufferByteInput
```

netty-codec declares jboss-marshalling as optional.

```
    <dependency>
      <groupId>org.jboss.marshalling</groupId>
      <artifactId>jboss-marshalling</artifactId>
      <optional>true</optional>
    </dependency>

```

This indicates that only those who need this serialization algorithm declare the dependencies.

# protobuf-nano

Netty supports protobuf serialization.

```
Class com.google.protobuf.nano.MessageNano is not found, referenced from 2 classes ▼

io.netty.handler.codec.protobuf.ProtobufDecoderNano
[...]

PR #1979, opened 2021-03-12
Excluding false positive linkage errors from Netty's Blockhound integration

suztomo:
For #1977

### DIFF
--- a/dependencies/src/main/resources/linkage-checker-exclusion-default.xml
+++ b/dependencies/src/main/resources/linkage-checker-exclusion-default.xml
@@ -141,9 +141,8 @@
     </Source>
     <Reason>
       If BlockHound library is present in the class path, it loads the Hidden class to enable
-      Netty-BlockHound integration that detects blocking calls. In this case, the missing classes
-      are on the class path. If BlockHound is not on the class path, nothing loads this source
-      class.
+      Netty-BlockHound integration. In this case, the target classes are on the class path.
+      If BlockHound is not on the class path, nothing loads this source class.
       https://github.com/reactor/BlockHound/blob/master/docs/custom_integrations.md
     </Reason>
   </LinkageError>
--- a/dependencies/src/main/resources/linkage-checker-exclusion-default.xml
+++ b/dependencies/src/main/resources/linkage-checker-exclusion-default.xml
@@ -140,9 +140,11 @@
       <Class name="io.netty.util.internal.Hidden" />
     </Source>
     <Reason>
-      Netty users can optionally add BlockHound dependency to detect blocking calls in their usage
-      of netty. The io.netty.util.internal.Hidden class implements the integration and this class
-      has references to missing classes by default.
+      If BlockHound library is present in the class path, it loads the Hidden class to enable
+      Netty-BlockHound integration that detects blocking calls. In this case, the missing classes
+      are on the class path. If BlockHound is not on the class path, nothing loads this source
+      class.
+      https://github.com/reactor/BlockHound/blob/master/docs/custom_integrations.md
     </Reason>
   </LinkageError>
   <LinkageError>
--- a/dependencies/src/main/resources/linkage-checker-exclusion-default.xml
+++ b/dependencies/src/main/resources/linkage-checker-exclusion-default.xml
@@ -133,6 +133,19 @@
     </Reason>
   </LinkageError>
   <LinkageError>
+    <Target>
+      <Package name="reactor.blockhound" />
+    </Target>
+    <Source>
+      <Class name="io.netty.util.internal.Hidden" />
+    </Source>
+    <Reason>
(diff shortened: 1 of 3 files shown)


dependencies/src/main/resources/linkage-checker-exclusion-default.xml
```diff
@@ -132,6 +132,19 @@
...
+  <LinkageError>
+    <Target>
+      <Package name="reactor.blockhound" />
+    </Target>
+    <Source>
+      <Class name="io.netty.util.internal.Hidden" />
+    </Source>
+    <Reason>
+      Netty users can optionally add BlockHound dependency to detect blocking calls in their usage
+      of netty. The io.netty.util.internal.Hidden class implements the integration and this class
```
elharo, line 144: So if they don't add this class, then there are linkage errors?

suztomo: Right. If Netty users don't add the optional BlockHound dependency, then Linkage Checker detects linkage errors in io.netty.util.internal.Hidden class has linkage errors. This won't manifest as runtime errors.

suztomo:
I need to update test criteria

```
[INFO] 
Error:  Failures: 
Error:    LinkageCheckerMainIntegrationTest.testBom_java11:173 expected:<Found 7[9]3 linkage errors> but was:<Found 7[1]3 linkage errors>
[INFO] 
```

dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerMainIntegrationTest.java
```diff
@@ -136,7 +136,7 @@ public void testBom_java8()
       LinkageCheckerMain.main(new String[] {"-b", "com.google.cloud:libraries-bom:1.0.0"});
       fail("LinkageCheckerMain should throw LinkageCheckResultException upon errors");
     } catch (LinkageCheckResultException expected) {
-      assertEquals("Found 771 linkage errors", expected.getMessage());
+      assertEquals("Found 691 linkage errors", expected.getMessage());
```
suztomo, line 139: Adding default exclusion rules reduces the number of linkage errors reported in a BOM.

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
elharo: I know we've seen this before. Do you recall why we didn't do this then? 

We should probably VC and screenshare to figure out how (or whether) this code is guarded.

elharo (commented):
We saw this in https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/1296

Did we fix this then?

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
elharo: Netty itself doesn't seem to call this Hidden class outside of tests. It's not obviou to me who does. It's only accessed via a service loader.

elharo: Perhaps Blockhound itself is the thing that loads this? Netty just makes it available to be loaded?

elharo: https://github.com/reactor/BlockHound/blob/master/docs/custom_integrations.md

elharo: I think this is OK, but the reason needs to be clearer and less generic. E.g. "This class is only loaded by Blockhound itself, in which case the missing classes are on the classpath. If Blockhound is not on the classpath, nothing loads this class."

suztomo: Thanks. Updated the explanation with the URL.

suztomo:
The references in #1296 were the references _from_ the BlockHound classes.

The exclusion rule in this PR is the reference _to_ the BlockHound classes from Netty.

dependencies/src/main/resources/linkage-checker-exclusion-default.xml
```diff
@@ -132,6 +132,21 @@
...
+  <LinkageError>
+    <Target>
+      <Package name="reactor.blockhound" />
+    </Target>
+    <Source>
+      <Class name="io.netty.util.internal.Hidden" />
+    </Source>
+    <Reason>
+      If BlockHound library is present in the class path, it loads the Hidden class to enable
+      Netty-BlockHound integration that detects blocking calls. In this case, the missing classes
```
elharo, line 144: delete "that detects blocking calls" since it's not relevant to the linkage error

suztomo: Removed

elharo (approved):

PR #1987, opened 2021-03-17
Upgrade google-auth-library-bom to 0.25.0

suztomo:
For https://github.com/googleapis/gax-java/pull/1302, which fails Linkage Monitor.

CC: @arithmetic1728

### DIFF
--- a/boms/cloud-oss-bom/pom.xml
+++ b/boms/cloud-oss-bom/pom.xml
@@ -235,7 +235,7 @@
        <dependency>
          <groupId>com.google.auth</groupId>
          <artifactId>google-auth-library-bom</artifactId>
-         <version>0.24.1</version>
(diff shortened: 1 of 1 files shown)


elharo (approved):

suztomo:
The BOM in the master branch does not affect Linkage Monitor check (It's different from the upper bound check).

The check still fails

```
SEVERE: Newly introduced problem:
(com.google.auth:google-auth-library-oauth2-http:0.24.1) com.google.auth.oauth2.GoogleCredentials's method createScoped(java.util.Collection, java.util.Collection) is not found
  referenced from com.google.api.gax.core.GoogleCredentialsProvider (com.google.api:gax:1.62.1-SNAPSHOT)
```

I think I'll need another release for this change to make the change pass.

PR #2019, opened 2021-04-09
deps: update googlecloud- libraries

elharo:
@suztomo probably don't have time to finish this today given how slow Github CI is running, but once this is in I can start on the next libraries-bom release


suztomo:
About the failures, it seems the BOM is not in the GCS copy of Maven Central yet.

suztomo (approved):

PR #2028, opened 2021-04-15
deps: remove maps

elharo:
@suztomo Based on discussion with the client, they think we can leave this one out for now. If it ends up causing problems for them, we may have to put it back in but for now I hope the maps API is separable from all the cloud APIs.


suztomo (approved):

suztomo:
kokoro-macos didn't trigger. Kicked it off via fusion.

PR #2033, opened 2021-04-23
Test Kokoro

suztomo:
Did Kokoro stop? Probably because the access token?


suztomo:
Kokoro is working.

Screen Shot 2021-04-23 at 14 50 49

suztomo:
The access token page shows the key is "Last used within the last week". Good.

PR #2044, opened 2021-05-05
Ltstest

elharo:
@suztomo experimenting with grabbing dependency lists

### DIFF
--- a/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/LtsBomIntegrationTest.java
+++ b/dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/LtsBomIntegrationTest.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
+import java.io.BufferedReader;
 import java.io.IOException;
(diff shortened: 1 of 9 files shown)


dashboard/src/main/java/com/google/cloud/tools/opensource/dashboard/LtsBomIntegrationTest.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  private static void testArtifact(Artifact artifact) throws IOException {
+    // download the pom.xml from Maven Central
+    URL remote = buildPomUrl(artifact);
+    Path temp = Files.createTempDirectory("pom");
+    Path local = temp.resolve("pom.xml");
+    Files.copy(remote.openStream(), local);
+    
+    // todo how to locate maven?
```
suztomo, line 69: Let the PATH environment variable take care of this. It should work in GitHub Actions and local computer.

elharo (commented):
Flattening removes the test dependencies from the published pom.xml. That's a problem. I need to think about what to do about that.

suztomo:
(Year-end cleanup)

PR #2058, opened 2021-05-12
organize the LTS BOM versions

elharo:
@suztomo I think you first suggested making all the versions properties. The more I work with this file the more sense that makes to me. What do you think?


boms/cloud-lts-bom/pom.xml
```diff
@@ -47,28 +47,37 @@
...
-    <io.grpc.version>1.36.1</io.grpc.version>
     <protobuf.version>3.16.0</protobuf.version>
-
     <http.version>1.39.1</http.version>
     <!-- We don't use gax-bom because it includes the artifacts with 'testlib' classifier.
         When updating gax.version, update gax.httpjson.version too. -->
     <gax.version>1.62.0</gax.version>
     <gax.httpjson.version>0.79.0</gax.httpjson.version>
+    
+    <!-- finalized versions -->
```
suztomo, line 73: Nice.

suztomo (approved):

PR #2063, opened 2021-05-13
Libraries-BOM: protobuf 3.17.0

suztomo:


elharo (commented):
Again the Maven mirror fails us. I thought we had stopped using this?

 Error:  Failed to execute goal on project linkage-monitor: Could not resolve dependencies for project com.google.cloud.tools:linkage-monitor:jar:1.5.10-SNAPSHOT: Could not transfer artifact org.apache.maven.resolver:maven-resolver-spi:jar:1.7.0 from/to google-maven-central (https://maven-central.storage-download.googleapis.com/maven2/): Transfer failed for https://maven-central.storage-download.googleapis.com/maven2/org/apache/maven/resolver/maven-resolver-spi/1.7.0/maven-resolver-spi-1.7.0.jar: Connection reset -> [Help 1]

suztomo:
When the mirror fixed its stale problem, we switched back to the mirror.

suztomo:
The "connection reset" problem is not caused by the staleness of the mirror. It could be due to GitHub Actions machines or the mirror (Google Cloud Storage).

elharo:
Either way, I'm still skeptical of depending on mirrors. They don't seem any more reliable than Maven Central itself in practice, and add an additional single point of failure. Maybe if we could configure it to fallback to Maven Central if the mirror fails?

elharo (approved):

PR #2114, opened 2021-06-15
gcp-lts-bom: all versions are finalized

suztomo:
I forgot to move the beam version from interim to the finalized.

### DIFF
--- a/boms/cloud-lts-bom/pom.xml
+++ b/boms/cloud-lts-bom/pom.xml
@@ -47,12 +47,9 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    
-    <!-- interrim placeholder versions -->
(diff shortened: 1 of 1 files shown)


boms/cloud-lts-bom/pom.xml
```diff
@@ -47,12 +47,9 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    
-    <!-- interrim placeholder versions -->
-    <!-- TODO: For each library, update the versions to the LTS version -->
+
+    <!-- finalized versions -->
```
elharo, line 51: I think you can drop this comment completely.

suztomo: Removed this line.

elharo (approved):

PR #2121, opened 2021-06-22
Kokoro: Removing the use of Google Maven Central mirror

suztomo:
We decided to stop using Google Maven Central mirror but there are some configuration that still references it.


kokoro/continuous.bat
```diff
@@ -5,9 +5,6 @@ set PATH=%JAVA_HOME%\bin;%PATH%
 
 cd github/cloud-opensource-java
 
-mkdir %USERPROFILE%\.m2
-copy settings.xml %USERPROFILE%\.m2
```
elharo, line 9: Should we remove settings.xml from this repo? It's not used, I think.

suztomo: It's being removed in this PR (the 3rd file)

suztomo:
@elharo All checks passed. PTAL.

kokoro/continuous.bat
elharo: show missed it because github doesn't deletes :-)

elharo (approved):

PR #2124, opened 2021-06-24
Update app engine SDK version to 1.9.89

AlanGasperini:
Update the version of the App Engine SDK to 1.9.89, the latest released version.

PR #2125, opened 2021-06-24
lts-bom: appengine-api-sdk 1.9.89

suztomo:
(Taking over https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/2124)

As per the chat room discussion, choosing the latest appengine API SDK.

CC: @AlanGasperini


elharo:
1.9.86 is a deliberate choice for consistency with Beam 2.30, and what we've tested the other components against.

elharo (approved):

PR #2126, opened 2021-06-25
Add gax testlib dependencies

gilles:


suztomo:
GAX's testlib-classifier artifacts are not useful outside Google. Do you want to share your intent of the PR?

gilles:
Sure, I inherited some code that makes use of `MockGrpcService` with the intent of mocking the memcache memorystore API but it turns out the code is not using any of that. Not sure what was the original intent so I deleted the dead code on my side and thus do not depend on `gax-grpc:testing` anymore.

suztomo:
Thank you for update.

PR #2168, opened 2021-08-05
Fixing BOM release script to work for a patch release

suztomo:
Before this PR, the release script always worked on the master branch.

However, for making a patch release on top of an old BOM release, we need to commit changes to a release branch so that we can release patch versions of old releases. The release script needs to adjust to follow this practice. Updating RELEASING.md with this explanation.

### DIFF
--- a/boms/cloud-lts-bom/RELEASING.md
+++ b/boms/cloud-lts-bom/RELEASING.md
@@ -38,13 +38,16 @@ We maintain multiple branches for the LTS BOM releases.
 Let's call the branches _LTS release branches_.
 An LTS release branch has a name "N.0.0-lts", where N is the major release version number,
 such as `1.0.0-lts` and `2.0.0-lts`.
-In this GitHub repository, the branches are configured as protected,
-so that we do not accidentally push changes to them.
+In this GitHub repository, the branches are [configured as protected](
+https://github.com/GoogleCloudPlatform/cloud-opensource-java/settings/branches),
+so that we do not accidentally push changes into them.
 
-Any change for a patch release to the LTS BOM should be merged to one of the LTS release branches.
-This practice enables us to release multiple versions of the BOM at the same time.
+Any changes for a patch release to the LTS BOM should be merged to one of the LTS release branches.
+This practice enables us to release a patch version of one of the old versions of the BOM,
+without using the master branch (or _HEAD_).
+
 Note that even when you prepare a patch release with higher patch number than 1,
-you commit changes to `N.0.0-lts` branch (the patch part is `0`).
+you commit changes to `N.0.0-lts` branch (the patch part is "0").
 For example, when you make changes for LTS BOM release 5.0.3, your pull request would merge
 the changes into `5.0.0-lts` branch.
 The branch includes all the changes between 5.0.0 and 5.0.2 releases.
@@ -69,3 +72,5 @@ has non-zero patch part.
 For example, when you run the release script with argument `./scripts/release.sh lts 5.0.3`,
 it creates a release branch `5.0.3-lts` based on the LTS release branch `5.0.0-lts`.
 
+For LTS patch releases, the release script creates a pull request that bumps the patch version of the LTS release branch
+(not the master branch).
--- a/boms/cloud-lts-bom/RELEASING.md
+++ b/boms/cloud-lts-bom/RELEASING.md
@@ -51,3 +51,4 @@ For example, when you run the release script with argument `./scripts/release.sh
 it creates a release branch `2.0.4-lts` from the LTS release branch `2.0.0-lts`,
 (not from `2.0.3-lts`).
 
+
--- a/boms/cloud-lts-bom/RELEASING.md
+++ b/boms/cloud-lts-bom/RELEASING.md
@@ -29,3 +29,25 @@ The Rapid workflow uploads the artifact to OSSRH staging repository.
 [Instructions for releasing from OSSRH are on the internal team 
 site](https://g3doc.corp.google.com/company/teams/cloud-java/tools/developers/releasing.md#verify-and-release).
 
+
+## Making changes to a released LTS BOM
+
+We maintain the branch of each release, such as `1.0.0-lts` and `2.0.0-lts`. Let's call
+them _LTS release branches_. An LTS release branch has a name "N.0.0-lts", where N is
+the major release version number.
+They are configured as protected branches so that we do not accidentally make changes.
+
+Any change for a patch release should be merged to the LTS release branches , rather than the master branch;
+otherwise we cannot release multiple versions of the BOM at the same time.
+Note that even when you prepare a patch release with higher patch number than 1,
+you commit changes to `N.0.0-lts` branch.
+For example, when you make changes for 2.0.5 LTS BOM release, your pull request would merge
+the changes into `2.0.0-lts` branch. The branch includes all the changes between
+2.0.0 and 2.0.4 releases.
+
+The release script creates a release branch for a patch release by checking out the LTS release
+branches (`N.0.0-lts`) if the releasing version has non-zero patch version.
+For example, when you run the release script with argument `./scripts/release.sh lts 2.0.4`,
+it creates a release branch `2.0.4-lts` from the LTS release branch `2.0.0-lts`,
+(not from `2.0.3-lts`).
+
--- a/boms/cloud-lts-bom/RELEASING.md
+++ b/boms/cloud-lts-bom/RELEASING.md
@@ -15,7 +15,7 @@ All on your corp desktop:
 the `cloud-opensource-java` directory:
 
 ```
-$ ./scripts/release.sh lts <release version> <post-release-version>
+$ ./scripts/release.sh lts <release version>
 ```
 
 This script creates a pull request and initiates the Rapid release workflow.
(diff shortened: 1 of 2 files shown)


scripts/release.sh
```diff
@@ -60,8 +60,18 @@ if [[ $(git status -uno --porcelain) ]]; then
   Die 'There are uncommitted changes.'
 fi
 
-# Make sure client is up to date with the latest changes.
-git checkout master
+if [[ "${SUFFIX}" = "lts" && "${VERSION}" != *.*.0 ]]; then
+  # LTS patch release is based on N.0.0-lts branch, where N is the major release number.
+  BASE_VERSION=$(echo $VERSION | sed 's/\([0-9][0-9]*\.[0-9][0-9]*\)\.[0-9][0-9]*/\1.0/')
```
chanseokoh, line 65:
```suggestion
  BASE_VERSION=$(echo $VERSION | sed 's/\([0-9]\+\.[0-9]\+\)\.[0-9]\+/\1.0/')
```

scripts/release.sh
```diff
@@ -60,8 +60,18 @@ if [[ $(git status -uno --porcelain) ]]; then
   Die 'There are uncommitted changes.'
 fi
 
-# Make sure client is up to date with the latest changes.
-git checkout master
+if [[ "${SUFFIX}" = "lts" && "${VERSION}" != *.*.0 ]]; then
```
chanseokoh, line 63: So if you run `./release.sh lts 5.1.3`, you will take the `else` branch, and `BASE_BRANCH` will be `master`. Is this correct?

scripts/release.sh
```diff
@@ -60,8 +60,18 @@ if [[ $(git status -uno --porcelain) ]]; then
...
+  # LTS patch release is based on N.0.0-lts branch, where N is the major release number.
+  BASE_VERSION=$(echo $VERSION | sed 's/\([0-9][0-9]*\.[0-9][0-9]*\)\.[0-9][0-9]*/\1.0/')
+  # For example, LTS BOM patch release 5.0.3 would create "5.0.3-lts" branch based on the base
+  # branch "5.0.0-lts". For the details of a patch release, see boms/cloud-lts-bom/RELEASING.md.
```
chanseokoh, line 67: According to the code, for more clarity, I'd say
```suggestion
  # For example, LTS BOM patch release 5.1.3 would create "5.1.3-lts" branch based on the base
  # branch "5.1.0-lts". For the details of a patch release, see boms/cloud-lts-bom/RELEASING.md.
```
Or, is this really the intention?

scripts/release.sh
chanseokoh: Also update the usage on line 2.
```
#!/bin/bash -
# Usage: ./release.sh <dependencies|bom> <release version>
```

suztomo: > if you run ./release.sh lts 5.1.3, you will take the else branch, and BASE_BRANCH will be master. Is this correct?

No, that command uses "lts-5.1.0" as BASE_BRANCH.

The second part of the if-condition is 'not matching'.

scripts/release.sh
suztomo: Nice. updated.

scripts/release.sh
suztomo: This BOM doesn't touch minor version. The minor version is always zero. I added clarification in the comment.

boms/cloud-lts-bom/RELEASING.md
```diff
@@ -29,3 +29,48 @@ The Rapid workflow uploads the artifact to OSSRH staging repository.
...
+
+## Performing a patch release
+
+The command for patch releases is the same as the normal releases: 
+
+```
+$ ./scripts/release.sh lts <patch release version>
+```
+
+Behind the scene, for a patch release the release script creates a release branch based on the
```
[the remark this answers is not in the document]
dzou:
```suggestion
Behind the scene, for a patch release, the release script creates a release branch based on the
```

scripts/release.sh
suztomo: The line 2 is updated.

suztomo (commented):
@chanseokoh PTAL.

boms/cloud-lts-bom/RELEASING.md
```diff
@@ -29,3 +29,48 @@ The Rapid workflow uploads the artifact to OSSRH staging repository.
...
+In this GitHub repository, the branches are [configured as protected](
+https://github.com/GoogleCloudPlatform/cloud-opensource-java/settings/branches),
+so that we do not accidentally push changes into them.
+
+Any changes for a patch release to the LTS BOM should be merged to one of the LTS release branches.
+This practice enables us to release a patch version of one of the old versions of the BOM,
+without using the master branch (or _HEAD_).
+
+Note that even when you prepare a patch release with higher patch number than 1,
+you commit changes to `N.0.0-lts` branch (the patch part is "0").
```
dzou, line 50: Yeah as Dan G. mentioned in chat, it might be more clear to name these branches like `N.0.x-lts` instead of `N.0.0-lts` to indicate that the branch includes all the changes in the minor version range.

We used a similar naming practice in spring-cloud-gcp: https://github.com/spring-cloud/spring-cloud-gcp/branches
i.e. branches named 1.0.x and 1.1.x

dzou (approved):

scripts/release.sh
chanseokoh: I see. But when you do `./release.sh lts 5.1.3`, it will take this "if" branch, and `BASE_VERSION` will be `5.1.0` (according to the `sed` expression`, and then `BASE_BRANCH=5.1.0-lts`.

Or, is it that the user should never give an input like `./release.sh lts 5.1.3`? Should it be guarded?

suztomo:
Thanks. I will revise the script to work with N.0.x branches, where N is a major version number.

boms/cloud-lts-bom/RELEASING.md
```diff
@@ -11,16 +11,39 @@ All on your corp desktop:
...
+```
+
+If this is a patch version release, the LTS release branch already exists.
+No action required for this step #2.
+
+3. Run `release.sh` with `lts` argument in 
 the `cloud-opensource-java` directory:
 
 ```
-$ ./scripts/release.sh lts <release version> <post-release-version>
```
suztomo, line 18: Removing the post-release-version, it bumps the patch version automatically.

boms/cloud-lts-bom/RELEASING.md
```diff
@@ -11,16 +11,39 @@ All on your corp desktop:
 
 1. Run gcert if you have not done so in the last twelve hours or so.
 
-2. Run `release.sh` with `lts` argument in 
+2. If this is a major version release, create a LTS release branch `N.0.x-lts`, where
```
suztomo, line 14: Now this operation includes creating an LTS release branch before performing a release.

suztomo (commented):
@chanseokoh @dzou PTAL. Now this operation uses "N.0.x-lts" branches, where N is the major release number.

dzou (approved):

scripts/release.sh
suztomo: Now the strategy is updated. In this example, the BASE_BRANCH will be `5.1.x-lts`.

scripts/release.sh
```diff
@@ -60,8 +60,18 @@ if [[ $(git status -uno --porcelain) ]]; then
   Die 'There are uncommitted changes.'
 fi
 
-# Make sure client is up to date with the latest changes.
-git checkout master
+if [[ "${SUFFIX}" = "lts" ]]; then
+  # LTS releases are based on N.0.x-lts branch, where N is the major release number.
+  # (Note that the minor version part of this BOM is always zero)
+  BASE_BRANCH=$(echo $VERSION | sed 's/\([0-9]\+\.[0-9]\+\)\.[0-9]\+/\1.x-lts/')
```
chanseokoh, line 66: So my understanding is that when it's LTS, the minor version is always 0, right?. Then I guess there's no need for the complex regex.
```suggestion
  BASE_BRANCH=$(echo $VERSION | sed 's/\([0-9]\+\)\.0\.[0-9]\+/\1.0.x-lts/')
```

If that's not the case, I think it's better to update the example to use `5.1.3` instead of `5.0.3`, which will give more clarity.

suztomo: Updated the regular expression.

scripts/release.sh
```diff
@@ -60,8 +60,21 @@ if [[ $(git status -uno --porcelain) ]]; then
   Die 'There are uncommitted changes.'
 fi
 
-# Make sure client is up to date with the latest changes.
-git checkout master
+if [[ "${SUFFIX}" = "lts" ]]; then
+  # LTS releases are based on N.0.x-lts branch, where N is the major release number.
+  # (Note that the minor version part of this BOM is always zero)
+  BASE_BRANCH=$(echo $VERSION | sed 's/\([0-9]\+\).0\.[0-9]\+/\1.0.x-lts/')
+  if [[ "${BASE_BRANCH}" != *0.x-lts ]]; then
```
chanseokoh, line 67: Probably
```suggestion
  BASE_BRANCH=$(echo $VERSION | sed 's/\([0-9]\+\)\.0\.[0-9]\+/\1.0.x-lts/')
  if [[ "${BASE_BRANCH}" != *.0.x-lts ]]; then
```

suztomo: Fixed.

chanseokoh (approved):

PR #2181, opened 2021-08-11
Gradle plugin: suppress duplicate circular dependency output

suztomo:
When dealing with an artifact with large dependency tree, the circular dependency message is printed many times (picture below). This PR fixes that.

Screen Shot 2021-08-11 at 19 52 22

### DIFF
--- a/gradle-plugin/src/functionalTest/groovy/com/google/cloud/tools/dependencies/gradle/BuildStatusFunctionalTest.groovy
+++ b/gradle-plugin/src/functionalTest/groovy/com/google/cloud/tools/dependencies/gradle/BuildStatusFunctionalTest.groovy
@@ -184,4 +184,33 @@ class BuildStatusFunctionalTest extends Specification {
     !result.output.contains("StackOverflowError")
     result.task(":linkageCheck").outcome == TaskOutcome.FAILED
   }
+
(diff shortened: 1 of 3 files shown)


gradle-plugin/src/functionalTest/groovy/com/google/cloud/tools/dependencies/gradle/BuildStatusFunctionalTest.groovy
```diff
@@ -184,4 +184,33 @@ class BuildStatusFunctionalTest extends Specification {
...
+    def result = GradleRunner.create()
+        .withProjectDir(testProjectDir.root)
+        .withArguments('linkageCheck', '--stacktrace')
+        .withPluginClasspath()
+        .buildAndFail()
+
+    then:
+    result.output.contains("Task :linkageCheck")
+    result.output.count("Circular dependency for: com.fasterxml.jackson.core:jackson-core:2.12.1") == 1
+    result.output.count("Circular dependency for: com.fasterxml.jackson:jackson-bom:2.12.1") == 1
```
suztomo, line 213: The description of the PR shows there are many such outputs before this change. Now it only shows them once.

dzou (approved):

Issue #2179, opened 2021-08-11
Linkage Error should show return type of method signature mismatch


suztomo:
Linkage Error should show return type of method signature mismatch.

```
io.grpc:grpc-stub:1.24.1 does not define io.grpc.stub.ClientCalls's method futureUnaryCall(io.grpc.ClientCall, Object) but io.grpc:grpc-stub:1.2.0 defines it.
```

suztomo:
```
[ERROR] The following builds failed:
[ERROR] *  bom-project-error/pom.xml
[ERROR] *  fail-build-for-linkage-errors/pom.xml
[ERROR] *  return-type-mismatch/pom.xml
[ERROR] *  report-only-reachable/pom.xml
```

PR #2182, opened 2021-08-12
Symbols to include type signatures in string representation

suztomo:
Fields to include their types and methods to include return types.

Fixes https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/2179


elefeint (approved):
That's helpful!

PR #2183, opened 2021-08-12
Allow protected member access of java.lang.Object

### LINKED ISSUE #2177: Empty Groovy file generating linkage errors on  Object's finalize() method
A user reported that an empty Groovy file generates linkage errors.

suztomo:
Allow protected member access of java.lang.Object, because all objects are subclasses of java.lang.Object.

Fixes #2177. Before this change Linkage Checker would throw errors on the methods of java.lang.Object:

```
java.lang.Object's method finalize() is not accessible;
  referenced by 6 class files
    groovy.lang.GroovyLogTestCase (org.codehaus.groovy:groovy-all:2.3.6)
    groovy.util.GroovyShellTestCase (org.codehaus.groovy:groovy-all:2.3.6)
    groovy.util.JavadocAssertionTestSuite (org.codehaus.groovy:groovy-all:2.3.6)
    org.codehaus.groovy.tools.shell.CommandsMultiCompleter (org.codehaus.groovy:groovy-all:2.3.6)
```

However, the member is "protected" (https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#finalize()) and java.lang.Object is the super class of all Java classes. These errors are false positives. This PR fixes this problem.

### DIFF
--- a/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
+++ b/dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
@@ -509,6 +509,11 @@ public class LinkageChecker {
       if (ClassDumper.classesInSamePackage(targetClass.getClassName(), sourceClassName)) {
         return true;
       }
+      if (Object.class.getName().equals(targetClass.getClassName())) {
+        // All objects are subclasses of java.lang.Object. Therefore the protected members are
+        // accessible to all objects.
+        return true;
+      }
       try {
         JavaClass sourceClass = classDumper.loadJavaClass(sourceClassName);
         if (ClassDumper.isClassSubClassOf(sourceClass, targetClass)) {
--- a/dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerTest.java
+++ b/dependencies/src/test/java/com/google/cloud/tools/opensource/classpath/LinkageCheckerTest.java
@@ -1321,11 +1321,9 @@ public class LinkageCheckerTest {
 
   @Test
   public void testProtectedMethodsOfObject() throws Exception {
-    Artifact groovy =
-        new DefaultArtifact("org.codehaus.groovy:groovy-all:2.3.6");
+    Artifact groovy = new DefaultArtifact("org.codehaus.groovy:groovy-all:2.3.6");
     ClassPathResult classPathResult =
-        new ClassPathBuilder()
-            .resolve(ImmutableList.of(groovy), false, DependencyMediation.MAVEN);
+        new ClassPathBuilder().resolve(ImmutableList.of(groovy), false, DependencyMediation.MAVEN);
 
     LinkageChecker linkageChecker = LinkageChecker.create(classPathResult.getClassPath());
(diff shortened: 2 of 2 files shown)


dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
```diff
@@ -509,6 +509,11 @@ private boolean isMemberAccessibleFrom(
       if (ClassDumper.classesInSamePackage(targetClass.getClassName(), sourceClassName)) {
         return true;
       }
+      if (Object.class.getName().equals(targetClass.getClassName())) {
```
chanseokoh, line 512: Does it also need `... && member.notPrviate()`?

suztomo: Line 508 checks whether this member is protected or not. I don't think a method can have protected and private.

Line 519 checks whether it's subclass or not. Now I think this condition should be in `ClassDumper.isClassSubClassOf`. Updating this.

suztomo: I changed my mind. Line 519 requires to load java.lang.Object. But the current code (c883b018fe42a26aa5165706d175dc8987f8f6f0) doesn't require it and it's obvious that all objects are subclass of java.lang.Object. So I keep the code as it is.

suztomo (commented):
@chanseokoh PTAL.

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
chanseokoh: Line 508 checks whether this member is protected or not.

Ah, I missed it.


> Also, should we ensure `targetClass` is `sourceClass` or a subclass of it?

Maybe I misunderstood the context, but what I mean above is this:
```java
  Object obj = new Object();
  Runnable runnable = new Runnable() {
    @Override
    public void run() {
      // sourceClass = runnable
      // targetClass = obj
      // runnable cannot call obj.clone() as it's protected
      obj.clone();  // doesn't compile
    }
  };
```
Or, is it that I'm out of the context?

suztomo: Thank you for the example. That's a new perspective to me. Unfortunately Linkage Checker cannot determine such cases. In the example below, they are both the same 
Screen Shot 2021-08-12 at 15 30 16



As Linkage Checker deal with class files in JAR files, we can assume the code already passed the compilation.

chanseokoh: Cool, that's fair!

chanseokoh (approved):

dependencies/src/main/java/com/google/cloud/tools/opensource/classpath/LinkageChecker.java
suztomo: Created an issue for this potential false negatives.

https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/2184

PR #2228, opened 2021-10-21
Setting java.net.preferIPv6Stack=true

suztomo:


suztomo:
It works without the option.

PR #2237, opened 2021-11-15
Added a check for no-downgrade rule for BOMs

suztomo:
Adding a check to enforce the no-downgrade rule.

@dzou Would you review this?

### DIFF
--- a/boms/cloud-lts-bom/pom.xml
+++ b/boms/cloud-lts-bom/pom.xml
@@ -62,7 +62,7 @@
     <!-- We don't use gax-bom because it includes the artifacts with 'testlib' classifier.
         When updating gax.version, update gax.httpjson.version too. -->
     <gax.version>2.6.1</gax.version>
-    <gax-httpjson.version>0.92.0</gax-httpjson.version>
+    <gax-httpjson.version>0.91.1</gax-httpjson.version>
     <google-cloud-core.version>2.2.0</google-cloud-core.version>
 
     <!-- Layer 2: Cloud -->
@@ -80,7 +80,7 @@
     <bigtable-hbase-beam.version>1.25.2-sp.1</bigtable-hbase-beam.version>
     <google-cloud-bigtable.version>2.2.0</google-cloud-bigtable.version>
     <google-cloud-pubsub.version>1.114.7</google-cloud-pubsub.version>
-    <proto-google-cloud-pubsub-v1.version>1.93.0-sp.1</proto-google-cloud-pubsub-v1.version>
+    <proto-google-cloud-pubsub-v1.version>1.96.7</proto-google-cloud-pubsub-v1.version>
     <datastore-v1-proto-client.version>2.1.3</datastore-v1-proto-client.version>
     <proto-google-cloud-datastore-v1.version>0.92.3</proto-google-cloud-datastore-v1.version>
     <google-cloud-service-usage.version>2.1.3</google-cloud-service-usage.version>
--- a/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/BomTest.java
+++ b/dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/BomTest.java
@@ -16,40 +16,51 @@
 
 package com.google.cloud.tools.opensource.dashboard;
 
+import com.google.cloud.tools.opensource.classpath.ClassPathBuilder;
+import com.google.cloud.tools.opensource.classpath.ClassPathEntry;
+import com.google.cloud.tools.opensource.classpath.ClassPathResult;
+import com.google.cloud.tools.opensource.classpath.DependencyMediation;
+import com.google.cloud.tools.opensource.dependencies.Artifacts;
+import com.google.cloud.tools.opensource.dependencies.Bom;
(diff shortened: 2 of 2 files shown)


suztomo:
I believe the check fails because the gcp-lts-bom is incomplete.

suztomo:
This is the failure message:

```
Error:  Failures: 
Error:    BomTest.testLtsBom:47->checkBom:63->assertNoDowngradeRule:141 com.google.api:gax-httpjson:jar:0.92.0 has a transitive dependency com.google.api:gax:jar:2.7.0. This is higher version than com.google.api:gax:jar:2.6.1 in the BOM
com.google.cloud:google-cloud-pubsub:jar:1.114.7 has a transitive dependency com.google.api.grpc:proto-google-cloud-pubsub-v1:jar:1.96.7. This is higher version than com.google.api.grpc:proto-google-cloud-pubsub-v1:jar:1.93.0-sp.1 in the BOM
```

boms/cloud-lts-bom/pom.xml
```diff
@@ -80,7 +80,7 @@
     <bigtable-hbase-beam.version>1.25.2-sp.1</bigtable-hbase-beam.version>
     <google-cloud-bigtable.version>2.2.0</google-cloud-bigtable.version>
     <google-cloud-pubsub.version>1.114.7</google-cloud-pubsub.version>
-    <proto-google-cloud-pubsub-v1.version>1.93.0-sp.1</proto-google-cloud-pubsub-v1.version>
+    <proto-google-cloud-pubsub-v1.version>1.96.7</proto-google-cloud-pubsub-v1.version>
```
suztomo, line 83: https://search.maven.org/artifact/com.google.cloud/google-cloud-pubsub-bom/1.114.7/pom

```
        <artifactId>proto-google-cloud-pubsub-v1</artifactId>
        <version>1.96.7</version><!-- {x-version-update:proto-google-cloud-pubsub-v1:current} -->
```

boms/cloud-lts-bom/pom.xml
```diff
@@ -62,7 +62,7 @@
     <!-- We don't use gax-bom because it includes the artifacts with 'testlib' classifier.
         When updating gax.version, update gax.httpjson.version too. -->
     <gax.version>2.6.1</gax.version>
-    <gax-httpjson.version>0.92.0</gax-httpjson.version>
+    <gax-httpjson.version>0.91.1</gax-httpjson.version>
```
suztomo, line 65: https://search.maven.org/artifact/com.google.api/gax-bom/2.6.1/pom

```
        <artifactId>gax-httpjson</artifactId>
        <version>0.91.1</version><!-- {x-version-update:gax-httpjson:current} -->
```

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/BomTest.java
```diff
@@ -72,4 +83,64 @@ private static void assertReachable(String url) throws IOException {
...
+            artifact
+                + " has a transitive dependency "
+                + transitiveDependency
+                + ". This is higher version than "
+                + bomArtifact
+                + " in the BOM\n");
+      }
+    }
+
+    Assert.assertFalse(errorMessages.toString(), violationFound);
```
dzou, line 144: Instead of having a `violationFound` variable, you can just assertFail that `!errorMessages.isEmpty()`.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/BomTest.java
```diff
@@ -72,4 +83,64 @@ private static void assertReachable(String url) throws IOException {
...
+   * <p>For example, if google-http-client 1.40.1 is in the BOM, then no other libraries in the BOM
+   * must not depend on the higher version of the google-http-client.
+   *
+   * @param bom the BOM to assert with this no-downgrade rule.
+   */
+  private static void assertNoDowngradeRule(Bom bom) throws InvalidVersionSpecificationException {
+
+    boolean violationFound = false;
+    StringBuilder errorMessages = new StringBuilder();
+    Map<String, Artifact> versionlessCoordinatesToArtifact = new HashMap<>();
```
dzou, line 102: I think `versionlessCoordinates` is too verbose; it will be clear from the code that `Artifact.makeKey(..)` does not include the version.

I think it would be more clear to refer to this map as something like `bomArtifacts` to indicate these are the artifacts declared directly by the BOM. Then you will check that transitive dependencies do not include an artifact newer than what is directly declared by the BOM.

dzou (approved):
Nice test!

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/BomTest.java
suztomo: Nice idea. Updated.

dashboard/src/test/java/com/google/cloud/tools/opensource/dashboard/BomTest.java
suztomo: Updated to bomArtifacts and simplified the logic.

PR #2246, opened 2021-11-23
gcp-lts-bom: adding google-cloud-iamcredentials

suztomo:
@chingor13 @Neenu1995 As discussed in the chat, adding the google-cloud-iamcredentials artifact in the BOM.

### DIFF
--- a/boms/cloud-lts-bom/pom.xml
+++ b/boms/cloud-lts-bom/pom.xml
@@ -68,6 +68,7 @@
     <!-- Layer 2: Cloud -->
     <appengine-api-1.0-sdk.version>1.9.91</appengine-api-1.0-sdk.version>
     <google-iam-admin.version>1.0.0</google-iam-admin.version>
+    <google-cloud-iamcredentials.version>2.0.6</google-cloud-iamcredentials.version>
     <google-api-services-androidpublisher.version>v3-rev20211021-1.32.1</google-api-services-androidpublisher.version>
     <google-cloud-bigquery.version>2.3.3</google-cloud-bigquery.version>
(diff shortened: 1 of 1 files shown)


boms/cloud-lts-bom/pom.xml
```diff
@@ -267,8 +268,17 @@
       </dependency>
       <dependency>
         <groupId>com.google.cloud</groupId>
-        <artifactId>google-iam-admin</artifactId>
+        <artifactId>google-iam-admin-bom</artifactId>
```
suztomo, line 271: Using the BOM to cover API surface.

boms/cloud-lts-bom/pom.xml
```diff
@@ -267,8 +268,17 @@
...
         <version>${google-iam-admin.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-iamcredentials-bom</artifactId>
```
suztomo, line 278: Using the BOM to cover API surface.

boms/cloud-lts-bom/pom.xml
dzou: Sorry, just to clarify, what's the reason for putting bom here instead of standard dependencies?

dzou (approved):

boms/cloud-lts-bom/pom.xml
suztomo: It's to cover API surface.

PR #2310, opened 2022-07-08
BomContentTest disableMavenCentralCheck property

suztomo:


boms/integration-tests/src/test/java/com/google/cloud/BomContentTest.java
```diff
@@ -66,9 +70,17 @@ public void testLibrariesBom() throws Exception {
 
   private void checkBom(Path bomPath) throws Exception {
     Bom bom = Bom.readBom(bomPath);
+
+    // Sometimes the artifacts are not yet available in Maven Central and only available in local
+    // Maven repository. Use
+    boolean enableMavenCentralAvailabilityCheck =
```
blakeli0, line 76: The variable name is the opposite of the property name, can we keep them consistent? Either `disableMavenCentralCheck` or `enableMavenCentralAvailabilityCheck`?

suztomo: Updated.

boms/integration-tests/src/test/java/com/google/cloud/BomContentTest.java
```diff
@@ -66,9 +70,17 @@ public void testLibrariesBom() throws Exception {
 
   private void checkBom(Path bomPath) throws Exception {
     Bom bom = Bom.readBom(bomPath);
+
+    // Sometimes the artifacts are not yet available in Maven Central and only available in local
+    // Maven repository. Use this property in that case.
+    boolean disableMavenCentralCheck =
+        ! "true".equals(System.getProperty("disableMavenCentralCheck"));
```
blakeli0, line 77: I think we should also remove `!` here as well.

blakeli0 (approved):

PR #2326, opened 2023-02-16
Exclude appengine artifacts from duplicate class check

suztomo:
Appengine API SDK is known to have duplicate classes from the same source in Google-internal source code repository. Because they are from the same source and having the same content, it will not cause dependency conflicts.

@blakeli0 Can you confirm the error you observed was at `testLtsBom`:

```
  @Test
  public void testLtsBom() throws Exception {
    Path bomPath = Paths.get("..", "cloud-lts-bom", "pom.xml").toAbsolutePath();
```

This in turn calls `checkBom` and then `assertUniqueClasses`, where this fix is applied.


suztomo:
@blakeli0 If this looks good, then merge this after approval.

blakeli0 (approved):

PR #2400, opened 2025-03-21
ci: Fixed inefficiency in the findRootCause method

shawabhijit:
The fix improves efficiency by:

Adding a null/empty check to handle edge cases gracefully
Still using the first path to create the initial set of coordinates
Using Guava's Sets.intersection() method instead of repeated retainAll() calls
This creates new intersection sets rather than modifying the original set in-place
The result is a more efficient implementation that scales better with larger numbers of dependency paths and larger sets


shawabhijit:
Fixed inefficiency using Guava's sets utility for more efficient intersection

PR #2448, opened 2025-10-29
ci: dependencies as a module in Release Please in master branch

suztomo:
This change splits the dependencies module as an independent module rather than using the root. When we used the root "." as the Release Please's module, then it mixes up LTS changes with dependencies module. See https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/2447 having "LTS 9" as a minor version bump to the dependencies module release.

This change also updates the Gradle plugin's version declaration from the gradle.properties to build.gradle with Release Please annotation. Release Please annotation is easier in build.gradle as it supports `//` with source code comment.


mpeddada1 (approved):

suztomo:
I'll fix the Gradle build failure in subsequent pull request.

